PR #106, opened 2026-01-27
Extract x-trace-id header to constant

### LINKED ISSUE #99: Extract hardcoded header name 'x-trace-id' into a constant
The string `"x-trace-id"` is hardcoded in multiple places — `middleware/trace_id.rs`, `context.rs`, etc. If we ever rename this header, we'd have to hunt down every occurrence.

Should be a constant like:

```rust
pub const TRACE_ID_HEADER: &str = "x-trace-id";
```

Then use it everywhere instead of the raw string.

juv:
Fixes #99

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,12 +65,19 @@ async fn main() -> std::io::Result<()> {
 
 ## Why Rapina?
 
-| Principle | Description |
-|-----------|-------------|
-| **Opinionated** | Convention over configuration. Clear defaults, escape hatches when needed. |
-| **Type-safe** | Typed extractors, typed errors, everything checked at compile time. |
-| **AI-friendly** | Predictable patterns that humans and LLMs understand equally well. |
-| **Production-ready** | Standardized errors with `trace_id`, JWT auth, observability built-in. |
+| Principle              | Description |
+|------------------------|-------------|
+| **Opinionated**        | Convention over configuration. Clear defaults, escape hatches when needed. |
+| **Type-safe**          | Typed extractors, typed errors, everything checked at compile time. |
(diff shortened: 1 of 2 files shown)


rapina/src/middleware/trace_id.rs
```diff
@@ -7,6 +7,8 @@ use crate::response::BoxBody;
 
 use super::{BoxFuture, Middleware, Next};
 
+pub const TRACE_ID_HEADER: &'static str = "x-trace-id";
```
arferreira: I'm pretty sure clippy is gonna claim this @juv, did you run into clippy?

arferreira (dismissed):

rapina/src/middleware/trace_id.rs
juv: Funny. I extracted this with RustRover IDE built-in function "extract as constant" and the IDE is configured to run Clippy on each save but for some reason RustRover has problems indexing/analyzing this project. It can also not find the types in rapina-cli/src/commands/doctor.rs, showing errors for `to_string().green()`, `to_string().yellow()`,  `to_string().red()` despite having the "Expand macros" setting enabled. Let me check, something is off in my local environment

juv: Updating to RustRover 2025.3.3 helped. Please take a look once more. Checks are green now

arferreira (approved):
Looks good to me

rapina/src/middleware/trace_id.rs
arferreira: > Updating to RustRover 2025.3.3 helped. Please take a look once more. Checks are green now

hell yeah, I don't use RustRover. Great job!

PR #132, opened 2026-01-31
Add support for modular API grouping

### LINKED ISSUE #131: [Feature]: Add route nesting support for modular API grouping
### Problem Statement

Currently, managing a large number of API endpoints in a single router structure can become unwieldy. As the number of endpoints grows, readability and maintainability decrease.

For example, a standard REST API often splits resources into logical collections (e.g., `/api/users`, `/api/products`, `/api/invoices`). Without a nesting feature, we are often forced to:

- Manually repeat the path prefix for every single route definition
- Define all routes in a single large file, which hurts separation of concerns

I propose adding a mechanism to "group" or "nest" routers under a specific path segment. This would allow developers to split their API logic into multiple sub-routers (modules), where each module is responsible for its own sub-paths, independent of the parent prefix.

Taking inspiration from the **Axum** `.nest` [implementation](https://docs.rs/axum/latest/axum/), the main router setup would look like this:
```
//main.rs
    Router::new()
        .nest("/users", users::router())
        .nest("/products", products::router())
        .nest("/invoices", invoices::router())
```

Each module would then define its own router without needing to know about the global /users prefix:

``` 
//users.rs
pub fn router() -> Router<AppState> {
    Router::new()
        .route("", get(list_users)) 
        .route("/:id", get(get_user)) 
}
``` 

Benefits

- Managing large codebases with dozens or hundreds of endpoints will be easier
- Reduces noise in the main entry point of the application
- Makes it easier to move groups of endpoints to different path prefixes without touching the handlers themselves

### Proposed Solution

Rapina could offer a similar "grouping"/"nesting" functionality like Axum to compose of the list of routes on the main `Router` instance.

### Alternatives Considered

Another alternative I can think of - but can't really comment whether it is technically possible - is to have the Rapina macros take effect on the main `Router` even if they are declared in sub modules. But this could also impose some more challenges, for example how to declare the path prefixes for that group. 

### Contribution

juv:
Fixes #131

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -326,6 +326,20 @@ dependencies = [
 ]
 
 [[package]]
+name = "dashmap"
+version = "6.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5041cc499144891f3790297212f32a74fb938e5136a14943f338ef9e0ae276cf"
+dependencies = [
+ "cfg-if",
+ "crossbeam-utils",
+ "hashbrown 0.14.5",
+ "lock_api",
+ "once_cell",
+ "parking_lot_core",
+]
+
+[[package]]
 name = "der"
 version = "0.7.10"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -623,6 +637,12 @@ dependencies = [
 
 [[package]]
 name = "hashbrown"
+version = "0.14.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e5274423e17b7c9fc20b6e7e208532f9b19825d82dfd615708b70edd83df41f1"
+
+[[package]]
+name = "hashbrown"
 version = "0.16.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "841d1cc9bed7f9236f321df977030373f4a4163ae1a7dbfe1a51a2c1a51d9100"
@@ -855,7 +875,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7714e70437a7dc3ac8eb7e6f8df75fd8eb422675fc7678aff7364301092b1017"
 dependencies = [
  "equivalent",
- "hashbrown",
+ "hashbrown 0.16.1",
 ]
 
 [[package]]
@@ -1393,6 +1413,7 @@ name = "rapina"
 version = "0.2.1"
 dependencies = [
  "bytes",
+ "dashmap",
  "dotenvy",
  "http",
  "http-body-util",
--- a/README.md
+++ b/README.md
@@ -185,6 +185,20 @@ rapina openapi check                    # Verify spec matches code
 rapina openapi diff --base main         # Detect breaking changes
 ```
 
+### Rate Limiting
+
+Protect your API from abuse with token bucket rate limiting:
+
+```rust
+Rapina::new()
+    .with_rate_limit(RateLimitConfig::per_minute(100))
+    .router(router)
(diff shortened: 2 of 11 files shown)


rapina/examples/endpoint_groups.rs
```diff
@@ -0,0 +1,32 @@
...
+
+#[get("/")]
+async fn hello() -> &'static str {
+    "Hello, Rapina!"
+}
+
+#[get("/health")]
+async fn health() -> StatusCode {
+    StatusCode::OK
+}
```
arferreira: could we remove? non used in the router below

rapina/src/router.rs
```diff
@@ -261,6 +286,19 @@ impl Router {
...
+
+    fn join_group_route_pattern(prefix: &str, route_path: &str) -> String {
+        let prefix = prefix.trim_end_matches('/');
+        let route_path = route_path.trim_start_matches('/');
+
+        if prefix.is_empty() {
+            format!("/{}", route_path)
+        } else if route_path.is_empty() {
+            prefix.to_string()
+        } else {
```
arferreira, line 302: we should formalize this to always start with `/`, or is this intentional?
if prefix = "api" and route_path = "users" it's gonna be "api/users" instead of "/api/users"
maybe it was intentional for flexibility

rapina/src/router.rs
```diff
@@ -461,4 +499,103 @@ mod tests {
...
+    #[test]
+    fn test_multiple_router_groups() {
+        let users_router = Router::new()
+            .get_named("", "list_users", |_req, _params, _state| async {
+                StatusCode::OK
+            })
+            .post_named("", "create_user", |_req, _params, _state| async {
+                StatusCode::CREATED
+            })
+            .post_named("/:id", "get_user", |_req, _params, _state| async {
```
arferreira: this should probably be get_named, unless handler is wrong

rapina/src/router.rs
juv: Yes I tried this. Interestingly, even mixed combinations compile to a working webserver.
```
use rapina::prelude::*;

#[get("/")]
async fn hello() -> &'static str {
    "Hello, Rapina!"
}

#[get("")]
async fn list_users() -> String {
    "list_users() of users_router, this would typically be located in users.rs"
}

#[get("")]
async fn another_list_users() -> String {
    "another list_users()"
}

#[get("/:id")]
async fn get_user(id: Path<u64>) -> String {
    format!("ID: {}", id.into_inner())
}

#[tokio::main]
    let users_router = Router::new().get("what", another_list_users);
    let users_router_mixin = Router::new().get("", list_users).get("/:id", get_user);

    let api_router = Router::new()
        .get("", list_users)
        .group("users", users_router)
        .group("/users", users_router_mixin);

    let router = Router::new().get("/", hello).group("/api", api_router);

    Rapina::new().router(router).listen("127.0.0.1:3000").await
}

```

This webserver now serves: 
http://127.0.0.1:3000/
http://127.0.0.1:3000/api
http://127.0.0.1:3000/api/users
http://127.0.0.1:3000/api/users/123455
http://127.0.0.1:3000/api/users/what

I am totally fine with removing this and requiring a `/` prefix. It definitely reads a bit more idiomatic.
Do you want me to add it?

juv: I just saw that in this example, when changing it to 
`    let router = Router::new().get("/", hello).group("api", api_router);` 
will break the router. So I assume the first level of grouping requires a `/` prefix. I think this is what you meant. At that point we can also just make it a required `/` prefix. I will change it 👍

arferreira (approved):
LGTM

PR #133, opened 2026-02-02
fix: improve error messages for better developer feedback

### LINKED ISSUE #92: Improve error messages with suggestions and context
Our error messages are functional but not very helpful. Compare:

```
Before: "missing field: email"
After:  "Missing required field 'email' in JSON body"

Before: "invalid path param"
After:  "Path parameter 'id' must be a valid integer, got 'abc'"
```

Audit `rapina/src/error.rs`, `rapina/src/extract.rs`, and `rapina/src/config.rs`. Make every message tell you what went wrong and what was expected.

uemuradevexe:
Fixes #92

Improves error messages to clearly describe what went wrong
and what was expected, replacing generic errors with
actionable feedback

### DIFF
--- a/rapina/src/extract.rs
+++ b/rapina/src/extract.rs
@@ -473,7 +473,9 @@ where
         _state: &Arc<AppState>,
     ) -> Result<Self, Error> {
         let (param_name, value) = params.iter().next().ok_or_else(|| {
-            Error::bad_request("Missing path parameter. Ensure your route pattern includes a parameter like /:id")
+            Error::bad_request(
+                "Missing path parameter. Ensure your route pattern includes a parameter like /:id",
+            )
         })?;
 
         let parsed = value.parse::<T>().map_err(|e| {


rapina/src/extract.rs
```diff
@@ -462,14 +471,19 @@ where
...
     ) -> Result<Self, Error> {
-        let value = params
-            .values()
-            .next()
-            .ok_or_else(|| Error::bad_request("missing path param"))?;
-
-        let parsed = value
-            .parse::<T>()
-            .map_err(|e| Error::bad_request(format!("invalid path param: {}", e)))?;
+        let (param_name, value) = params.iter().next().ok_or_else(|| {
```
arferreira, line 475: this might not be a framework bug lol
could be user error if they user for example a Path<T> on a route without path params
consider something like: Missing path parameter. Ensure your route pattern includes a parameter like /:id

PR #139, opened 2026-02-10
Add Cookie extractor for typed cookie access

### LINKED ISSUE #85: Add Cookie extractor for typed cookie access
We have typed extractors for `Path`, `Query`, `Json`, `Form`, `Headers` — but not for cookies. You have to manually parse the `Cookie` header.

Should work like:

```rust
#[derive(Deserialize)]
struct Session {
    session_id: String,
}

#[get("/dashboard")]
async fn dashboard(session: Cookie<Session>) -> Result<Json<Dashboard>> {
    // ...
}
```

Returns 400 if a required cookie is missing. `Option<Cookie<T>>` for optional ones.

Look at `rapina/src/extract.rs` — follow the same pattern as `Headers` and `Query`.

arferreira:
## Summary

- Add `Cookie<T>` extractor for deserializing cookies into typed structs
- Parse `Cookie` header and deserialize to target type via serde
- Returns 400 Bad Request if required cookies are missing

 ## Usage
 ```rust
 #[derive(Deserialize)]
 struct Session {
     session_id: String,
 }

 #[get("/dashboard")]
 async fn dashboard(session: Cookie<Session>) -> Result<Json<Dashboard>> {
     let data = session.into_inner();
     // ...
 }
 ```

## Related Issues

Closes #85

## Checklist

- [x] `cargo fmt` passes
- [x] Tests pass

### DIFF
--- a/rapina/src/lib.rs
+++ b/rapina/src/lib.rs
@@ -132,5 +132,6 @@ pub mod prelude {
     pub use rapina_macros::{Config, delete, get, post, public, put};
 }
 
-// Re-export schemars so users don't need to add it to their Cargo.toml
+// Re-export schemars and http crates so users don't need to add them to their Cargo.toml
+pub use http;
 pub use schemars;


rapina/src/extract.rs
```diff
@@ -463,6 +496,44 @@ impl FromRequestParts for Headers {
...
+                    Some((key, value))
+                }
+            })
+            .collect();
+
+        // Serialize to JSON then deserialize to target type
+        let json = serde_json::to_string(&cookies)
+            .map_err(|e| Error::bad_request(format!("Failed to process cookies: {}", e)))?;
+
+        let value: T = serde_json::from_str(&json)
```
juv, line 530: Serializing to JSON and then deserializing from JSON to T looks interesting. Is this a good approach (genuinely asking)?

arferreira: fair point! it's a pragmatic shortcut, cookies are always k/v pairs, so the json roundtrip works, but yeah it's not the ideal world, I think maybe a direct Hashmap<String, String> -> T would be cleaner, good candidate for a follow-up improvement.

PR #144, opened 2026-02-10
Re-export http crate from lib.rs

### LINKED ISSUE #142: Re-export http crate from rapina
## Problem

When using Rapina with route macros, the generated code references `http::Request` internally. However, the `http` crate is not re-exported from rapina, forcing users to manually add it to their `Cargo.toml`:

```toml
[dependencies]
rapina = "0.2.1"
http = "1"  # Required because macro references http::Request
```

Without this, users get:
```
error[E0433]: failed to resolve: use of unresolved module or unlinked crate `http`
```

## Discovered via

Dogfooding exercise - building a blog API with SeaORM integration to test real-world DX.

## Solution

Re-export `http` from rapina's prelude or root module so users don't need to add it separately:

```rust
// In rapina/src/lib.rs
pub use http;
```

This follows the pattern of other frameworks like axum that re-export their dependencies.

ShiraiEd:
## Summary

What does this PR do?

## Related Issues

Closes #142

## Checklist

- [x] `cargo fmt` passes
- [x] Tests pass

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,12 @@
 version = 4
 
 [[package]]
+name = "adler2"
+version = "2.0.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
+
+[[package]]
 name = "aho-corasick"
 version = "1.1.4"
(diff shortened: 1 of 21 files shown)


arferreira (dismissed):
LGTM, please fix lint, clippy and stash your commits then we're good to merge

ShiraiEd:
I think this is becoming harder than it should be

PR #149, opened 2026-02-13
Add unit and integration tests for TraceIdMiddleware

### LINKED ISSUE #102: Add unit tests for TraceIdMiddleware
`TraceIdMiddleware` in `rapina/src/middleware/trace_id.rs` has zero tests.

It should be tested for:
- Generates a new UUID when no `x-trace-id` header is present
- Preserves the existing `x-trace-id` when one is already in the request
- Always includes `x-trace-id` in the response

Check `request_log.rs` in the same folder for a test pattern to follow.

yan-pi:
## Summary

Add unit and integration tests for `TraceIdMiddleware`, covering construction (`new`, `Default`) and verifying that an incoming `x-trace-id` header is preserved in the response.

## Related Issues

Closes #102

## Checklist

- [x] `cargo fmt` passes
- [x] Tests pass

### DIFF
--- a/rapina/tests/middleware.rs
+++ b/rapina/tests/middleware.rs
@@ -521,3 +521,25 @@ async fn test_compression_skips_without_accept_encoding() {
     assert_eq!(response.status(), StatusCode::OK);
     assert!(response.headers().get("content-encoding").is_none());
 }
+
+#[tokio::test]
+async fn test_trace_id_middleware_preserves_incoming_trace_id() {
+    let app = Rapina::new()
+        .with_introspection(false)
+        .middleware(TraceIdMiddleware::new())
+        .router(Router::new().route(http::Method::GET, "/health", |_, _, _| async { "ok" }));
+
+    let client = TestClient::new(app).await;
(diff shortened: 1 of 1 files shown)


rapina/src/middleware/trace_id.rs
```diff
@@ -59,3 +59,18 @@ impl Middleware for TraceIdMiddleware {
...
+
+    #[test]
+    fn test_request_log_middleware_new() {
+        let _mw = TraceIdMiddleware::new();
+    }
+
+    #[test]
+    fn test_request_log_middleware_default() {
+        let _mw: TraceIdMiddleware = Default::default();
+    }
```
arferreira: These say "request_log_middleware" but they're testing TraceIdMiddleware

yan-pi: I see, that was my mistake for not paying enough attention. I used those modules that Ive mentioned as references because they were already in the codebase and were similar to the `TraceIdMiddleware` that I should implement.

Issue #229, opened 2026-02-23
[Bug]: rapina dev fails to parse Cargo.toml


uemuradevexe:
### Description

Running rapina dev (or rapina test) inside a valid Rapina project fails immediately

### Steps to Reproduce

1. have rapina-cli installed
  2. navigate to any Rapina project directory
  3. run rapina dev

### Expected Behavior

rapina dev should start the development server successfully

### Code Example

```rust
Error: Failed to parse Cargo.toml: TOML parse error at line 1, column 10
    |
  1 | [package]
    |          ^
  unexpected content, expected nothing
```

### Environment

- Rapina version: 0.6.0
- Rust version: 1.93.1
- OS: macOS

Issue #207, opened 2026-02-22
Extract duplicate verify_rapina_project into shared CLI utility


arferreira:
The `verify_rapina_project()` function is copy-pasted identically across three files:

- `rapina-cli/src/commands/dev.rs`
- `rapina-cli/src/commands/add.rs`
- `rapina-cli/src/commands/test.rs`

**What to do:**

1. Create a shared utility module (e.g. `rapina-cli/src/util.rs`)
2. Move `verify_rapina_project()` there
3. Update all three files to use the shared version
4. Make sure `cargo test` and `cargo clippy` pass

PR #237, opened 2026-02-25
fix: fixed Cargo.toml parser error when run rapina dev and rapina tests

Ismaellima4:
## Summary

What does this PR do?

- fixed Cargo.toml parser error when run rapina dev and rapina tests
- fixed another parsing error that was in the get_binary_name function in commads/dev when run rapina dev

## Related Issues

#229 
#207

## Checklist

- [x] `cargo fmt` passes
- [x] Tests pass

### DIFF
--- a/rapina/src/introspection/endpoint.rs
+++ b/rapina/src/introspection/endpoint.rs
@@ -61,6 +61,11 @@ pub async fn list_routes(
 
 #[cfg(test)]
 mod tests {
+    use http::{HeaderValue, Method};
+    use serde_json::Value;
+
+    use crate::{app::Rapina, router::Router, testing::TestClient};
+
     use super::*;
 
     #[test]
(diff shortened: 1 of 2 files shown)


rapina-cli/src/commands/dev.rs
```diff
@@ -153,37 +155,8 @@ pub fn execute(config: DevConfig) -> Result<(), String> {
...
-            "This doesn't appear to be a Rapina project (no rapina dependency found)".to_string(),
-        );
-    }
-
-    Ok(())
-}
-
 /// Build the project and run the server.
-fn build_and_run(config: &DevConfig) -> Result<Child, String> {
+fn build_and_run(config: &DevConfig, binary_name: &String) -> Result<Child, String> {
```
arferreira: Should be `binary_name: &str`

rapina-cli/src/commands/mod.rs
```diff
@@ -8,3 +8,34 @@ pub mod new;
...
+            "This doesn't appear to be a Rapina project (no rapina dependency found)".to_string(),
+        );
+    }
+
+    Ok(parsed)
+}
+
+pub fn parse_cargo_toml(content: &str) -> Result<toml::Value, String> {
+    toml::from_str(&content).map_err(|e| format!("Failed to parse Cargo.toml: {}", e))
+}
```
arferreira: This is a one-liner used once, we drop it and call it directly

arferreira: BTW, `toml::from_str(&content).map_err(|e| format!("Failed to parse Cargo.toml: {}", e))`, `content` is already a `&str`, so if you use `&content`, is gonna be `&&str`. That works via auto-deref but unnecessary.

PR #253, opened 2026-02-27
Add `rapina import openapi` command

### LINKED ISSUE #169: rapina import openapi — generate project from OpenAPI spec
This is a design issue, not implementation-ready yet. Looking for feedback on the approach.

Most teams migrating to a new framework already have an API. They have an OpenAPI spec (or can export one). Right now they'd have to rewrite everything from scratch. That's a big ask.

The idea: `rapina import openapi spec.yaml` reads an OpenAPI 3.0/3.1 spec and generates a complete Rapina project — handlers with the correct HTTP methods and paths, request/response DTOs with serde derives, validation rules from the spec's constraints, error types, and route registration. The developer fills in business logic, everything else is scaffolded.

This is the same philosophy as \`rapina add resource\` but at a whole-project scale. You point at a spec, you get a working skeleton.

Supporting commands:
- `rapina import openapi --dry-run` — show what would be generated without writing files
- `rapina import openapi --diff` — compare an updated spec against existing generated code

The OpenAPI parsing infrastructure already exists in the codebase (we generate specs today), so reading them is the inverse operation. The code generation patterns from `rapina add resource` can be reused for the handler/DTO scaffolding.

Open questions:
- How to handle specs with features Rapina doesn't support yet (websocket callbacks, oneOf/anyOf schemas, etc.) — probably generate a TODO comment and move on
- Whether to generate into an existing project or always create a new one
- How to handle re-imports when the spec evolves — marker comments? separate generated vs hand-written directories?

Would love input from anyone who's done API migrations before.

arferreira:
Closes #169

Adds `rapina import openapi <file>` — reads an OpenAPI 3.0 spec (JSON or YAML) and generates handler stubs, DTOs, error types, and module structure following Rapina conventions. The use case is teams migrating an existing API to Rapina: instead of rewriting everything by hand, point the CLI at a spec file and get a working scaffold with `todo!()` bodies to fill in.

## How it works

Parses the spec with `openapiv3` v2, groups endpoints by tag (falls back to first path segment when untagged), and writes one feature module per group under `src/`. Each module gets `handlers.rs`, `dto.rs`, `error.rs`, and `mod.rs`.

Generated code follows existing Rapina patterns: `#[public]` on endpoints with no security requirement, `Validated<Json<T>>` when DTOs carry OpenAPI validation constraints (minLength, maxLength, email, range, pattern), `Path<T>`/`Query<T>` extractors for path and query params, typed error enums with `IntoApiError` + `DocumentedError`. PATCH operations map to `#[put]` with a comment since there's no `#[patch]` macro yet.

OpenAPI type mapping covers string formats (uuid → `Uuid`, date-time → `DateTimeUtc`), integer sizes (int32/int64), floats, booleans, arrays, and nested objects via `$ref` resolution. Unsupported constructs like `oneOf`/`anyOf` and external refs emit warnings and fall back to `serde_json::Value` rather than failing.

```
rapina import openapi spec.json
rapina import openapi spec.yaml --dry-run
rapina import openapi spec.json --tags users,posts
```

Feature-gated behind `import-openapi` in Cargo.toml so the base CLI compiles without `openapiv3`/`serde_yml` in the dependency tree. Uses `serde_yml` over `serde_yaml` since the latter is deprecated.

Separate file from the existing `import.rs` — the database import code has zero overlap with spec parsing and code generation.

This is V1: no re-import diffing, no `--diff` mode. Running it twice overwrites. That's a separate piece of work.

30 unit tests inline (no fixture files), manually tested against a multi-resource Petstore spec with dry-run, full import, and tag filtering.

### DIFF
--- a/rapina-cli/src/commands/codegen.rs
+++ b/rapina-cli/src/commands/codegen.rs
@@ -29,7 +29,6 @@ pub(crate) fn pluralize(s: &str) -> String {
     format!("{}s", s)
 }
 
-#[cfg(feature = "import")]
 pub(crate) fn singularize(s: &str) -> String {
     if let Some(stem) = s.strip_suffix("ies") {
         format!("{}y", stem)
--- a/rapina-cli/src/commands/import_openapi.rs
+++ b/rapina-cli/src/commands/import_openapi.rs
@@ -107,24 +107,7 @@ fn to_snake_case(s: &str) -> String {
 }
 
 /// Best-effort singularization for common English plurals.
-/// Handles regular cases (users->user, categories->category, boxes->box).
-/// Irregular plurals (analyses, people, data) are not covered — those are
-/// rare in API resource names and can be fixed manually.
-fn singularize(s: &str) -> String {
-    if let Some(stripped) = s.strip_suffix("ies") {
-        format!("{}y", stripped)
-    } else if s.ends_with("ses") || s.ends_with("xes") || s.ends_with("zes") {
-        // "boxes" -> "box", "buses" -> "bus", "fizzes" -> "fizz"
-        s[..s.len() - 2].to_string()
-    } else if s.ends_with("ss") {
-        // "class" stays "class"
-        s.to_string()
-    } else if let Some(stripped) = s.strip_suffix('s') {
-        stripped.to_string()
-    } else {
-        s.to_string()
-    }
-}
+use super::codegen::singularize;
 
 // ---------------------------------------------------------------------------
 // Path and naming utilities


rapina-cli/src/commands/codegen.rs
```diff
@@ -503,6 +504,7 @@ mod tests {
     use super::*;
 
     #[test]
+    #[cfg(feature = "import")]
```
ShiraiEd: This one too?

rapina-cli/src/commands/mod.rs
```diff
@@ -12,6 +12,9 @@ pub mod openapi;
 pub mod routes;
 pub mod test;
 
+#[cfg(feature = "import-openapi")]
+pub mod import_openapi;
+
 /// Verify that we're in a valid Rapina project directory.
 pub fn verify_rapina_project() -> Result<toml::Value, String> {
```
ShiraiEd, line 19: There is one function in codegen.rs with the same name but different outputs, this one can return a toml::Value inside a Result, the other a unit () in a Result,

To avoid future problems maybe consider to change the name

arferreira: Good catch. Removed the duplicate, the codegen one now delegates to the real implementation in `mod.rs`.

rapina-cli/src/commands/codegen.rs
```diff
@@ -29,6 +29,7 @@ pub(crate) fn pluralize(s: &str) -> String {
     format!("{}s", s)
 }
 
+#[cfg(feature = "import")]
```
ShiraiEd: maybe delete the ```#[cfg(feature = "import")]``` and use this fn in import_openapi.rs since it looks more complete and avoids creating 2 singularizes

rapina-cli/src/commands/import_openapi.rs
```diff
@@ -0,0 +1,1977 @@
...
+        println!("  {} {}", "warn:".yellow(), w);
+    }
+
+    if dry_run {
+        dry_run_report(&modules, &warnings);
+        return Ok(());
+    }
+
+    println!();
+    for module in &modules {
```
ShiraiEd, line 1122: If an error happens here, like writing an existing modulo, the ones after that wont be created and the ones before wont be deleted to facilitate the rerun of the fn,

Because checking if each modulo exists or not could be costly, something like a addition of  a 'ignore' arg in the Openapi struct would prevent that,

if ignore is true, instead of write_module returnan error, it would continue the loop with writing that one and print something like "{modulo_name} already exists, skipping that".

a 'force' arg is also a option, to overwrite the existing one

rapina-cli/src/commands/import_openapi.rs
```diff
@@ -0,0 +1,1977 @@
...
+}
+
+// ---------------------------------------------------------------------------
+// File I/O
+// ---------------------------------------------------------------------------
+
+fn write_module(module: &ModuleDefinition) -> Result<(), String> {
+    let module_dir = Path::new("src").join(&module.name);
+
+    if module_dir.exists() {
```
ShiraiEd, line 952: about the comment on line 1139

something like this:

```rust
fn write_module(module: &ModuleDefinition, ignore_existing: bool) -> Result<(), String> {
    let module_dir = Path::new("src").join(&module.name);

    if module_dir.exists() {
        if ignore_existing {
            println!(
                "  {} Skipping {} (already exists)",
                "warn:".yellow(),
                module.name
            );
            return Ok(());
        } else {
            return Err(format!(
                "Directory 'src/{}/' already exists. Use --ignore-existing to skip.",
                module.name
            ));
        }
    }

PR #267, opened 2026-02-28
Add built-in pagination

### LINKED ISSUE #265: Built-in pagination with Paginate extractor
Add a `Paginate` extractor and `Paginated<T>` response wrapper so pagination is a one-liner in handlers instead of manual boilerplate every time.

**Usage:**

```rust
#[get("/users")]
async fn list_users(db: Db, page: Paginate) -> Result<Json<Paginated<Model>>> {
    Ok(page.exec(User::find(), db.conn()).await?)
}
```

`Paginate` reads `?page=1&per_page=20` from the query string, validates, and applies defaults. `Paginated<T>` wraps the result with metadata:

```json
{
  "data": [...],
  "page": 1,
  "per_page": 20,
  "total": 150,
  "total_pages": 8
}
```

The `exec` method takes any SeaORM `Select<E>`, runs both the paginated fetch and count query, and returns `Paginated<T>`. Filters and ordering chain naturally before passing the select in:

```rust
let select = User::find()
    .filter(user::Column::Active.eq(true))
    .order_by_desc(user::Column::CreatedAt);

Ok(page.exec(select, db.conn()).await?)
```

**Defaults and guards:** page=1, per_page=20, max per_page=100. Invalid values return 422. Max configurable globally via `Rapina::new().with_pagination(PaginationConfig::default().max_per_page(50))`.

arferreira:
## Summary

Adds `Paginate` extractor and `Paginated<T>` response type so list endpoints don't need manual LIMIT/OFFSET boilerplate.

```rust
#[get("/users")]
async fn list_users(db: Db, page: Paginate) -> Result<Paginated<UserResponse>> {
    Ok(page.exec(User::find(), db.conn()).await?.map(UserResponse::from))
}
```

`exec()` runs fetch and count concurrently with `tokio::join!`. `Paginated<T>` implements `IntoResponse` and `JsonSchema` so it works as a direct return type. Optional `PaginationConfig` via `.state()` for custom defaults and max limits. Feature-gated behind `database`.

Docs page with examples added at `core-concepts/pagination.md`.

Closes #265


lucaserm (approved):
LGTM! It would be great add a CursorPaginate too, in case that you really need timestamp features, but it can be a single issue.

PR #268, opened 2026-02-28
Add response caching layer

### LINKED ISSUE #168: Add response caching layer
No built-in caching right now. Every request hits the handler, every time.

The idea is a caching layer with two backends, in-memory (default, no dependencies) and Redis (optional, for multi-instance deployments). Per-route caching via a macro attribute would be the cleanest API:

```rust
#[get("/products")]
#[cache(ttl = 60)]
async fn list_products(db: Db) -> Result<Json<Vec<Product>>> {
    // this only runs if the cache misses
}
```

Cache key should be derived from method + path + query params by default, with an option to customize it. Cache invalidation hooks for when data changes (POST/PUT/DELETE to the same resource path could auto-invalidate).

For the builder API, something like:
```rust
Rapina::new()
    .with_cache(CacheConfig::in_memory(1000))  // max entries
    // or
    .with_cache(CacheConfig::redis("redis://localhost:6379"))
```

This is a meaningful chunk of work, the macro integration, the backend abstraction, cache invalidation.

arferreira:
Per-route response caching via `#[cache(ttl = 60)]`. In-memory by default, Redis behind a feature flag.

The macro injects an internal header, the middleware reads it, caches the response, strips the header. Successful POST/PUT/DELETE/PATCH invalidates related GET entries by path prefix. No changes to existing APIs.

Tested both backends e2e with a fresh project. Redis tested against Docker with key/TTL inspection via redis-cli.

Closes #168

### DIFF
--- a/docs/content/docs/core-concepts/extractors.md
+++ b/docs/content/docs/core-concepts/extractors.md
@@ -21,6 +21,7 @@ Extractors automatically parse request data and inject it into your handlers. If
 | `Cookie<T>` | Typed cookie access |
 | `CurrentUser` | Authenticated user (JWT) |
 | `Validated<T>` | Validated extractor |
+| `Paginate` | Pagination params (requires feature) |
 | `Db` | Database connection (requires feature) |
 
 ## Path Parameters
--- a/docs/content/docs/core-concepts/pagination.md
+++ b/docs/content/docs/core-concepts/pagination.md
@@ -1,0 +1,239 @@
++++
+title = "Pagination"
+description = "Built-in pagination for list endpoints"
+weight = 6
+date = 2025-02-27
++++
+
+Every list endpoint needs the same boilerplate: LIMIT/OFFSET, a count query, and response metadata. Rapina handles this with a `Paginate` extractor and a `Paginated<T>` response wrapper.
+
+## Quick Start
+
+```rust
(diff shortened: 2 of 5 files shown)


rapina/src/cache_redis.rs
```diff
@@ -0,0 +1,190 @@
...
+/// Redis cache backend using multiplexed async connections.
+pub struct RedisCache {
+    conn: redis::aio::MultiplexedConnection,
+    prefix: String,
+}
+
+impl RedisCache {
+    /// Connects to Redis at the given URL.
+    pub async fn connect(url: &str) -> Result<Self, redis::RedisError> {
+        let client = redis::Client::open(url)?;
```
juv, line 56: I think a good follow-up issue here is to support custom CA certificates and optionally client certificate authentication for the Redis Client:
https://docs.rs/redis/latest/redis/struct.Client.html#method.build_with_tls 

Per default, only system root certificates are used in case a `rediss://` connection scheme is provided.
This is only supported with Redis' crate feature `rust-tls` though

juv (approved):
lgtm

rapina/src/cache_redis.rs
arferreira: Good catch man! I haven't thought about that honestly.

PR #282, opened 2026-03-03
Add raw WebSocket support behind feature flag

arferreira:
Adds WebSocketUpgrade extractor, WebSocket connection type, and Message enum behind the `websocket` feature flag. Uses hyper-tungstenite 0.19 for the upgrade handshake and tokio-tungstenite 0.28 for the stream.

The extractor works with both manual routes and proc macros — a handler with `WebSocketUpgrade` + `State<T>` just works through `#[get]` without any manual FromRequest calls.

Includes a chat example at `rapina/examples/websocket-chat/` and integration tests covering echo, binary messages, non-upgrade rejection, and concurrent connections.

Depends on #281.

### DIFF
--- a/rapina-cli/src/commands/mod.rs
+++ b/rapina-cli/src/commands/mod.rs
@@ -10,6 +10,7 @@ pub mod migrate;
 pub mod new;
 pub mod openapi;
 pub mod routes;
+pub mod templates;
 pub mod test;
 
 #[cfg(feature = "import-openapi")]
--- a/rapina-cli/src/commands/new.rs
+++ b/rapina-cli/src/commands/new.rs
@@ -4,12 +4,15 @@ use colored::Colorize;
 use std::fs;
 use std::path::Path;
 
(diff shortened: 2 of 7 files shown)


rapina/examples/websocket-chat/src/main.rs
juv, line 1: I feel like this file could use a bit more explanations as it reads much more complex than the other Rapina examples. Or at least a „warning“ that this is kinda an advanced topic and requires _some_ knowledge of Tokio‘s broadcast queue?

arferreira: Good call. This example is intentionally low-level since it's the raw WebSocket PR. PR #284 (relay + channel handlers) is where the ergonomic layer lives, a chat becomes a `#[relay("chat:*")]` handler without touching `tokio::sync::broadcast` directly. Once that lands, we'll add a relay-based chat example that's much more approachable.

PR #284, opened 2026-03-03
Add channel handlers and presence tracking

arferreira:
Adds server-side channel handlers to the relay system so topics aren't just dumb pipes. The `#[relay("pattern")]` macro registers handlers that receive `RelayEvent::Join`, `Message`, and `Leave` events — letting the server intercept subscriptions, validate messages, enrich payloads, and track presence.

Without this, every topic is a pass-through broadcast. With it, a handler like `#[relay("room:*")]` can reject unauthorized joins, persist messages, or push enriched responses back to subscribers.

The `#[relay]` macro generates a wrapper that matches the `ChannelHandlerFn` signature and uses `inventory` for compile-time discovery — same pattern as the route macros but without the struct/Handler impl (channel handlers aren't routes). First arg is always `RelayEvent`, remaining args go through `FromRequestParts` extraction with synthetic parts. `CurrentUser` is extracted before WebSocket upgrade and injected into the synthetic parts so auth works in channel handlers.

Pattern matching supports exact (`"chat:lobby"`) and prefix (`"room:*"`) patterns. Channels are sorted by specificity in `prepare()`: exact first, then prefix by length descending, so `"room:vip:*"` always wins over `"room:*"` regardless of inventory iteration order.

Subscribe flow subscribes to the backend *before* calling the Join handler so the joining client receives messages pushed during their own Join (matching Phoenix's semantics). If the handler rejects, the subscription is undone. Leave handlers are fire-and-forget via `tokio::spawn` so they don't block disconnect cleanup.

Presence tracking uses `DashMap<String, DashMap<u64, PresenceEntry>>` with atomic `remove_if` cleanup on the outer map when the inner map empties. The `Relay` extractor gains `track`, `untrack`, `presence`, and `presence_count` methods. The hub auto-untracks on unsubscribe and disconnect.

Stacked on #283.

### DIFF
--- a/rapina/tests/relay_channels.rs
+++ b/rapina/tests/relay_channels.rs
@@ -113,6 +113,16 @@ async fn exact_channel(event: RelayEvent, log: State<TestLog>) -> Result<()> {
     Ok(())
 }
 
+/// Handles "exact:*" topics — prefix handler that should lose to exact_channel
+/// for the literal "exact:topic" but win for anything else like "exact:other".
+#[rapina::relay("exact:*")]
+async fn exact_prefix_channel(event: RelayEvent, log: State<TestLog>) -> Result<()> {
+    if let RelayEvent::Join { topic, .. } = &event {
+        log.push(format!("exact-prefix-join:{topic}")).await;
+    }
+    Ok(())
+}
+
 /// Handles "auth:*" topics — requires CurrentUser, records user id.
 #[rapina::relay("auth:*")]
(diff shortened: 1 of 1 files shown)


ShiraiEd (commented):
the current tests to check the patterns order don't cover the specific over the wildcard *
Maybe a test to see if test:exact is prefered over test:* would be nice

ShiraiEd (approved):
LGTM

PR #285, opened 2026-03-03
Add group parameter to route macros for auto-discovery

### LINKED ISSUE #236: [Feature]: Support endpoint groups during auto discovery
### Problem Statement

The code changes in https://github.com/arferreira/rapina/pull/159 introduced the auto discovery feature, which is a great enhancement for DX. The current auto discovery implementation clashes with the endpoint group feature as the api group prefix is not known during auto discovery phase.

For example: in /examples/endpoint_group.rs, the auto discovery can not know about the "group prefix" ("/api/users") that was previously given with an explicit call to `.group("/api/users", users_router)` as there are multiple router instances involved.

### Proposed Solution

Taken from Rapina Discord discussion:

> .group() works at the router level, it takes a sub-Router and rewrites each route's pattern with the prefix. But .discover() bypasses that entirely because it registers routes directly onto self.router inside prepare().

>   So if you have #[get("/users")] and want the route at /api/users, you'd have to write #[get("/api/users")] in the macro, which defeats the purpose of grouping.

> A proper fix would probably look like a #[group("/api")] module-level attribute or a way for .discover() to accept prefix mappings. That's worth a design discussion, would love if you're wiling to open an issue to track it.

> For now, if you need grouped routes, stick with the manual Router::new().group() approach from the endpoint_groups example instead of .discover(). You can mix both: use .discover() for top-level routes and manual registration for grouped ones.

### Alternatives Considered
/

### Contribution

arferreira:
Closes #236.

Auto-discovered routes now accept an inline `group` parameter that joins a prefix with the handler path at compile time:

```rust
#[get("/users", group = "/api/v1")]
async fn list_users() -> Json<Vec<User>> { /* ... */ }
// registers at /api/v1/users
```

The prefix is resolved entirely in the proc macro during expansion. No changes to `RouteDescriptor`, `discovery.rs`, `app.rs`, or `router.rs`. Fully backward compatible.

Composes with `#[public]`, `#[cache(ttl = N)]`, and `#[errors(T)]`.

### DIFF
--- a/rapina-macros/src/lib.rs
+++ b/rapina-macros/src/lib.rs
@@ -338,6 +338,7 @@ fn is_parts_only_extractor(type_str: &str) -> bool {
         || type_str.contains("CurrentUser")
         || type_str.contains("Db")
         || type_str.contains("Cookie")
+        || type_str.contains("Relay")
 }
 
 /// Extracts the inner type from Json<T> wrapper for schema generation
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,4 @@
-/target
+**/target
 Cargo.lock
 
 # macOS
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -504,7 +504,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "0dc92fb57ca44df6db8059111ab3af99a63d5d0f8375d9972e319a379c6bab76"
 dependencies = [
  "generic-array",
- "rand_core",
+ "rand_core 0.6.4",
  "subtle",
  "zeroize",
 ]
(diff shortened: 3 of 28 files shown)


rapina-macros/src/lib.rs
```diff
@@ -2,23 +2,92 @@ use proc_macro::TokenStream;
...
 use syn::{FnArg, ItemFn, LitStr, Pat};
 
+/// Parsed route macro attribute: `"/path"` or `"/path", group = "/prefix"`.
+struct RouteAttr {
+    path: LitStr,
+    group: Option<LitStr>,
+}
+
+impl syn::parse::Parse for RouteAttr {
+    fn parse(input: syn::parse::ParseStream) -> syn::Result<Self> {
```
juv, line 12: What is the expected behaviour in case a route gets added more than once? Let’s say a user accidentally added the same route (method, path with/without group name) twice?

arferreira: Same as without groups, no duplicate detection, last one wins. Could be a good `rapina doctor` check down the road.

Issue #217, opened 2026-02-22
Add state and dependency injection documentation page


arferreira:
The `State<T>` extractor and `AppState` container are core DI mechanisms in Rapina but have no documentation page.

**What to cover:**

- Registering state with `.state(my_service)` on the app builder
- Extracting state in handlers with `State<T>`
- Using multiple state types
- Sharing database pools, config, and services across handlers
- How `AppState` works under the hood (type map)
- Example: injecting a database pool and a config struct

**Where:**

Create `docs/content/docs/core-concepts/state.md`. Source code is at `rapina/src/extractors/state.rs` and `rapina/src/app_state.rs`.

Issue #164, opened 2026-02-19
Documentation overhaul


arferreira:
The docs site exists but it's incomplete. Several core features are shipped but not documented, and the getting-started experience has gaps.

What's missing or thin right now:

- Middleware guide (#136) — all 7 middleware are shipped but there's no dedicated docs page showing how to use them, configure them, or write custom ones
- Migration/database guide (#153) — schema! macro, migrations, Db extractor, connection config are all shipped since v0.4.0 but undocumented on the site
- Getting started guide needs updating — configuration.md still shows the old .router() pattern, should show .discover()
- Auth guide could use more real-world examples (refresh tokens, logout, custom claims)
- No guide for the testing utilities (TestClient, TestRequest)
- No guide for OpenAPI generation and the CLI tooling around it

This is the umbrella issue. The individual doc pages can be separate PRs, but the goal is: someone should be able to build a complete API with auth, database, validation, and tests by following the docs alone, without reading source code.

Related: #136, #153

arferreira:
I covered some, going to grab openapi guide.

PR #302, opened 2026-03-06
add di docs

ShiraiEd:
## Summary
add the dependency injection docs
What does this PR do?
adds the state.md into docs/core-concepts
## Related Issues
#217 #164
Closes #
#217
## Checklist

- [x] `cargo fmt` passes
- [x] Tests pass

### DIFF
--- a/docs/content/docs/core-concepts/state.md
+++ b/docs/content/docs/core-concepts/state.md
@@ -1,0 +1,276 @@
++++
+title = "Dependency Injection"
+description = "Rapina AppState and dependency injections methods"
+weight = 3
+date = 2026-03-06
++++
+
+Rapina uses a type-safe container called `AppState` to share services across handlers. You register values once at startup — database pools, config structs, HTTP clients — and inject them anywhere via the `State<T>` extractor.
+
+`AppState` is backed by a `HashMap<TypeId, Arc<dyn Any + Send + Sync>>`. Each type gets one slot, keyed by its Rust type identity:
+
+```
+AppState
+  TypeId(AppConfig)          -> Arc<AppConfig>
+  TypeId(EmailClient)        -> Arc<EmailClient>
+  TypeId(DatabaseConnection) -> Arc<DatabaseConnection>
+  TypeId(...)                -> Arc<...>
+```
+
+At startup the state is wrapped in an `Arc` and shared across all requests. Per request, only the `Arc` is cloned — no data is ever copied.
+
+## State
+
+### Registering State
+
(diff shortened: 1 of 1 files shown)


docs/content/docs/core-concepts/state.md
```diff
@@ -0,0 +1,553 @@
...
+description = "Rapina AppState and dependency injections methods"
+weight = 3
+date = 2026-03-06
++++
+
+Rapina uses a type-safe container called `AppState` to share services across handlers. You register values once at startup: database pools, config structs, HTTP clients and inject them anywhere via the `State<T>` extractor.
+
+`AppState` is a `HashMap<TypeId, Arc<dyn Any + Send + Sync>>`. Each type gets one slot, keyed by its Rust type identity:
+
+```
```
arferreira, line 12: `AppState` is actually a newtype struct wrapping the HashMap, not the HashMap itself. Small distinction but worth being precise:

```suggestion
`AppState` is backed by a `HashMap<TypeId, Arc<dyn Any + Send + Sync>>`. Each type gets one slot, keyed by its Rust type identity:
```

docs/content/docs/core-concepts/state.md
```diff
@@ -0,0 +1,553 @@
...
+### Missing State
+
+If a handler requests `State<T>` but `T` was never registered, the request returns `500 Internal Server Error`:
+
+```
+State not registered for type 'my_crate::AppConfig'. Did you forget to call .state()?
+```
+
+### Overwriting State
+
```
arferreira, line 127: Everything from here down (Database, Middleware, Auth, Observability, WebSocket, Graceful Shutdown) already has dedicated pages in core-concepts. Having the same content in two places means they will drift apart. I would trim these sections and add links instead, the DI content above is the new stuff and it stands on its own.

docs/content/docs/core-concepts/state.md
```diff
@@ -0,0 +1,553 @@
++++
+title = "Dependency Injection"
+description = "Rapina AppState and dependency injections methods"
+weight = 3
+date = 2026-03-06
++++
+
+Rapina uses a type-safe container called `AppState` to share services across handlers. You register values once at startup: database pools, config structs, HTTP clients and inject them anywhere via the `State<T>` extractor.
```
arferreira:
```suggestion
Rapina uses a type-safe container called `AppState` to share services across handlers. You register values once at startup — database pools, config structs, HTTP clients — and inject them anywhere via the `State<T>` extractor.
```

arferreira (approved):

Issue #222, opened 2026-02-22
Improve naive pluralize function in add command


arferreira:
The `pluralize()` function in `rapina-cli/src/commands/add.rs` just appends "s" to any word:

```rust
fn pluralize(s: &str) -> String {
    format!("{s}s")
}
```

This produces wrong results for common resource names:
- "category" -> "categorys" (should be "categories")
- "status" -> "statuss" (should be "statuses") 
- "person" -> "persons" (should be "people")
- "address" -> "addresss" (should be "addresses")

**What to do:**

Handle the most common English pluralization rules:
- Words ending in "y" (preceded by a consonant) -> "ies"
- Words ending in "s", "x", "z", "ch", "sh" -> "es"
- Default -> "s"

No need to handle every irregular noun -- just the common patterns above. Add unit tests for the function.

PR #313, opened 2026-03-10
Add pluralized fn logic and tests

ShiraiEd:
## Summary
Improves the pluralize function and adds a test

## Related Issues
#222 

## Questions
Should I also improve the singularize and fix it to not return a warn on clippy?
Should I keep the redundant tests in add.rs?

## Checklist

- [x] `cargo fmt` passes
- [x] Tests pass

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2259,7 +2259,7 @@ dependencies = [
  "tracing-subscriber",
  "uuid",
  "validator",
- "windows",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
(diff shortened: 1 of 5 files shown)


rapina-cli/src/commands/codegen.rs
```diff
@@ -26,6 +26,20 @@ pub(crate) fn to_pascal_case(s: &str) -> String {
 }
 
 pub(crate) fn pluralize(s: &str) -> String {
+    let cases = [
+        ("y", "ies"),   //category -> categories
+        ("ss", "sses"), //adress -> addresses
```
arferreira:
```suggestion
        ("ss", "sses"), //address -> addresses
```

PR #314, opened 2026-03-10
Support UUID primary keys in schema! macro

### LINKED ISSUE #243: Support UUID primary keys in schema! macro
Follow-up from #170 / #240.

The schema! macro currently assumes `id: i32` for primary keys. Tables using UUID PKs are skipped during import with a warning. This is a common pattern (especially in distributed systems) and should be supported.

Ismaellima4:
## Summary

What does this PR do?
- Enhanced schema! macro to support UUID primary keys and resolve foreign key types dynamically.
- Updated rapina-cli import logic to correctly identify and validate UUID primary keys in SQLite, PostgreSQL, and MySQL.
- Modified code generation to handle non-i32 primary keys in handlers, DTOs, and migrations.
- Improved SQLite type mapping to support BLOB and Custom UUID types.
- Fixed a regression in composite primary key validation.
- Added comprehensive unit and integration tests for UUID support across macros and CLI.
## Related Issues

Closes #243 

## Checklist

- [x] `cargo fmt` passes
- [x] Tests pass

### DIFF
--- a/rapina-cli/src/commands/codegen.rs
+++ b/rapina-cli/src/commands/codegen.rs
@@ -94,10 +94,17 @@ pub(crate) fn generate_handlers(
         .collect();
     let update_body = update_checks.join("\n");
 
+    let uuid_import = if pk_type.to_lowercase().as_str() == "uuid" {
+        "use rapina::uuid::Uuid;"
+    } else {
+        ""
+    };
+
     format!(
         r#"use rapina::prelude::*;
 use rapina::database::{{Db, DbError}};
 use rapina::sea_orm::{{ActiveModelTrait, EntityTrait, IntoActiveModel, Set}};
+{uuid_import}
 
 use crate::entity::{pascal};
 use crate::entity::{singular}::{{ActiveModel, Model}};
@@ -174,6 +181,7 @@ pub async fn delete_{singular}(db: Db, id: Path<{pk_type}>) -> Result<Json<serde
         create_body = create_body,
         update_body = update_body,
         pk_type = pk_type,
+        uuid_import = uuid_import,
     )
 }
 
@@ -281,17 +289,18 @@ pub(crate) fn generate_schema_block(
     let mut attrs = String::new();
 
     if let Some(pk_cols) = primary_key {
-        attrs.push_str(&format!("\n    #[primary_key({})]\n", pk_cols.join(", ")));
+        attrs.push_str(&format!("\n    #[primary_key({})]", pk_cols.join(", ")));
     }
 
     if let Some(ts) = timestamps {
-        attrs.push_str(&format!("\n    #[timestamps({})]\n", ts));
+        attrs.push_str(&format!("\n    #[timestamps({})]", ts));
     }
 
     format!(
         r#"
 schema! {{
-    {pascal} {{{attrs}
+    {attrs}
+    {pascal} {{
 {fields}
     }}
 }}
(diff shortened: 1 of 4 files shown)


rapina-cli/src/commands/codegen.rs
```diff
@@ -341,13 +346,16 @@ impl MigrationTrait for Migration {{
...
+                    .col({{
+                        let mut col = ColumnDef::new({pascal_plural}::Id);
+                        match "{pk_type}".to_lowercase().as_str() {{
+                            "uuid" => col.uuid(),
+                            "i32" | "integer" => col.integer(),
+                            "i64" | "bigint" => col.big_integer(),
+                            _ => col.type_iden(rapina::migration::Alias::new("{pk_type}")),
+                        }};
+                        col.not_null().primary_key().to_owned()
+                    }})
```
arferreira: this drops `.auto_increment()` for i32 PKs too, the old template had it. without it, every insert on a regular integer PK table needs an explicit id value. shouldn't the integer branch still include `.auto_increment()`?

also, this match happens at runtime in the generated migration code, but the generator already knows the pk_type at codegen time. would it be simpler to just emit the right column definition directly instead of pushing the decision into the generated code?

rapina-cli/src/commands/codegen.rs
```diff
@@ -113,7 +114,7 @@ pub async fn list_{plural}(db: Db) -> Result<Json<Vec<Model>>> {{
 
 #[get("/{plural}/:id")]
 #[errors({pascal}Error)]
-pub async fn get_{singular}(db: Db, id: Path<i32>) -> Result<Json<Model>> {{
+pub async fn get_{singular}(db: Db, id: Path<{pk_type}>) -> Result<Json<Model>> {{
```
arferreira, line 143: when pk_type is `"Uuid"`, the generated file ends up with `Path<Uuid>` but there's no `use uuid::Uuid;` or `use rapina::uuid::Uuid;` anywhere in the template imports, wouldn't this fail to compile?

rapina-cli/src/commands/import.rs
```diff
@@ -98,6 +98,8 @@ fn map_mysql_type(col_type: &sea_schema::mysql::def::Type) -> NormalizedType {
         Type::Date => NormalizedType::Date,
         Type::Decimal(_) => NormalizedType::Decimal,
         Type::Json => NormalizedType::Json,
+        Type::Binary(_) => NormalizedType::Uuid,
+        Type::Varbinary(_) => NormalizedType::Uuid,
```
arferreira: are all binary columns really UUIDs though? `Binary` and `Varbinary` get used for hashes, encrypted data, serialized blobs, images, all sorts of things. wouldn't this mistype every binary column in someone's existing database?

maybe only `Binary(Some(16))` would be a reasonable guess? or leave these as `Unmappable` and let users annotate it manually.

rapina-cli/src/commands/import.rs
```diff
@@ -121,6 +123,8 @@ fn map_sqlite_type(col_type: &sea_schema::sea_query::ColumnType) -> NormalizedTy
         ColumnType::Date => NormalizedType::Date,
         ColumnType::Decimal(_) | ColumnType::Money(_) => NormalizedType::Decimal,
         ColumnType::Json | ColumnType::JsonBinary => NormalizedType::Json,
+        ColumnType::Binary(_) | ColumnType::Blob => NormalizedType::Uuid,
```
arferreira: same concern here, `Blob` is the most general binary type in SQLite. the `Custom(UUID)` check on the next line makes sense since it's explicit, but blanket `Blob` -> `Uuid` would produce wrong types for most real schemas, no?

rapina-cli/src/commands/import.rs
```diff
@@ -671,7 +703,7 @@ pub fn database(
 // Tests
```
arferreira, line 701: gating the whole test module on `import-postgres` means the pure-logic tests like `test_filter_accepts_uuid_pk` and `test_filter_skips_invalid_pk_type` won't run on a basic `cargo test`. those don't touch any postgres types at all, could the feature gate go on just the individual tests that actually call `map_pg_type`?

arferreira (changes requested):
nice work getting UUID support across the whole pipeline, the schema macro FK resolution is the part I like most, resolving the target entity's PK type for foreign keys is the right design. left some inline notes, mostly around a regression in the migration output and a couple of the type mappings.

arferreira (approved):

PR #330, opened 2026-03-12
Add seed database support

### LINKED ISSUE #146: Add database seeding support
## Summary

Add database seeding support to populate databases with initial or test data.

## Context

Builds on the database integration work in #143 which introduced the `schema!` macro for entity definitions.

## Proposed Features

- `rapina seed` CLI command
- Seed file structure (e.g., `seeds/` directory)
- Environment-aware seeding (dev vs production)
- Factory/fixture patterns for test data
- Idempotent seeding (don't duplicate data on re-run)

## Example Usage

```rust
// seeds/users.rs
use rapina::seed::Seed;

pub struct UserSeeder;

impl Seed for UserSeeder {
    async fn run(&self, db: &DatabaseConnection) -> Result<()> {
        let admin = user::ActiveModel {
            email: Set("admin@example.com".into()),
            name: Set("Admin".into()),
            ..Default::default()
        };
        admin.insert(db).await?;
        Ok(())
    }
}
```

## Open Questions

- Trait-based seeders vs declarative approach?
- How to handle seeding order with relationships?
- Integration with testing framework?

## References

- #143 - Database integration with schema! macro
- #145 - Database migrations support

yan-pi:
## Summary

Adds database seeding support to the Rapina CLI, closing #146.

Three new subcommands under `rapina seed`:

- **`rapina seed load [--entity NAME] [--fresh]`** - Load JSON seed files from `seeds/` into the database. Uses idempotent inserts (`ON CONFLICT DO NOTHING` for Postgres/SQLite, `INSERT IGNORE` for MySQL). The `--fresh` flag truncates tables before loading.
- **`rapina seed dump [--entity NAME]`** - Export database tables to JSON seed files in `seeds/`.
- **`rapina seed generate --count N [--entity NAME]`** - Generate fake seed data based on `schema!` macro definitions in `src/entity.rs`, with type-aware values (emails, names, UUIDs, dates).

### Key design decisions

- **Cross-database support**: All generated SQL is backend-aware (Postgres, SQLite, MySQL) for identifier quoting, placeholder syntax, conflict handling, and foreign key management.
- **SQL injection prevention**: All table/column names are quoted via `quote_ident()` with proper escaping per backend.
- **Foreign key safety**: `--fresh` disables FK checks before deletion (`PRAGMA foreign_keys = OFF` / `SET session_replication_role` / `SET FOREIGN_KEY_CHECKS`), deletes in reverse order, then re-enables.
- **Fake data generation**: Type-aware values using `fastrand`, with record index for sequential fields like emails.

### AI usage disclaimer

Parts of this PR were written with the help of an AI tool. I reviewed, adapted, and manually tested all changes. I also added the AI as a co-author here for transparency, since I was unsure if that was the preferred practice in this project.

## Related Issues

Closes #146

## Checklist

- [x] `cargo fmt` passes
- [x] Tests pass

### DIFF
--- a/rapina-cli/Cargo.toml
+++ b/rapina-cli/Cargo.toml
@@ -22,12 +22,16 @@ notify = { version = "8", default-features = false, features = [
   "macos_kqueue",
 ] }
 notify-debouncer-mini = "0.4"
+sea-orm = { version = "1", features = ["runtime-tokio-rustls"], optional = true }
+tokio = { version = "1", features = ["full"], optional = true }
+dotenvy = { version = "0.15", optional = true }
 ctrlc = "3"
 toml = "1.0"
 serde = { version = "1", features = ["derive"] }
 serde_json = "1"
 chrono = "0.4"
-tokio = { version = "1", features = ["full"], optional = true }
+fastrand = { version = "2.3.0", optional = true }
+uuid = { version = "1", features = ["v4"], optional = true }
 sea-schema = { version = "0.16", optional = true, default-features = false }
 sqlx = { version = "0.8", optional = true }
 openapiv3 = { version = "2", optional = true }
@@ -35,7 +39,11 @@ serde_yaml_ng = { version = "0.10", optional = true }
 
 [features]
 default = []
-import = ["tokio"]
+seed = ["dep:sea-orm", "dep:tokio", "dep:dotenvy", "dep:fastrand", "dep:uuid"]
+seed-postgres = ["seed", "sea-orm/sqlx-postgres"]
+seed-mysql = ["seed", "sea-orm/sqlx-mysql"]
+seed-sqlite = ["seed", "sea-orm/sqlx-sqlite"]
+import = ["dep:tokio"]
 import-postgres = [
     "import",
     "sea-schema/postgres",
@@ -67,3 +75,5 @@ import-openapi = ["dep:openapiv3", "dep:serde_yaml_ng"]
 
 [dev-dependencies]
 tempfile = "3.27.0"
+tokio = { version = "1", features = ["rt", "macros"] }  # for #[tokio::test]
+sea-orm = { version = "1", features = ["runtime-tokio-rustls", "sqlx-sqlite"] }
--- a/rapina-cli/src/commands/seed.rs
+++ b/rapina-cli/src/commands/seed.rs
@@ -1,0 +1,1040 @@
+//! Database seeding commands for loading, dumping, and generating seed data.
+
+use colored::Colorize;
+use sea_orm::{
+    ConnectionTrait, DatabaseBackend, DatabaseConnection, FromQueryResult, Statement,
+    TransactionTrait,
+};
+use std::fs;
+use std::path::Path;
+
+const SEEDS_DIR: &str = "seeds";
+
+/// An entity definition: entity name paired with its fields (name, type).
+type EntitySchema = Vec<(String, Vec<(String, String)>)>;
+
+// -- Public API --
+
+/// Load seed data from JSON files in the `seeds/` directory into the database.
+///
+/// Reads each `.json` file, parses it as an array of objects, and inserts records
+/// using idempotent `ON CONFLICT DO NOTHING` (Postgres/SQLite) or `INSERT IGNORE` (MySQL).
+///
+/// **Note:** Without `--fresh`, rows whose primary key already exists in the
+/// database are silently skipped (not updated). Use `--fresh` to truncate all
+/// target tables before loading. Full upsert (update-on-conflict) requires
+/// per-table primary key discovery and is planned for a future iteration.
+///
+/// # Errors
+///
+/// Returns an error if the seeds directory is missing, the database is unreachable,
+/// or any seed file contains invalid JSON.
+pub async fn load(entity: Option<String>, fresh: bool) -> Result<(), String> {
+    let seeds_path = Path::new(SEEDS_DIR);
+    verify_seeds_dir(seeds_path)?;
+    let conn = connect_to_db().await?;
+    let seed_files = discover_seed_files(seeds_path, entity.as_deref())?;
+
+    // Use a transaction so FK disable + inserts share the same connection
+    let txn = conn
+        .begin()
+        .await
+        .map_err(|e| format!("Failed to begin transaction: {}", e))?;
+
+    disable_fk_checks(&txn).await?;
+
+    if fresh {
+        truncate_tables(&txn, &seed_files).await?;
+    }
+
+    for (table_name, file_path) in &seed_files {
+        let content = fs::read_to_string(file_path)
+            .map_err(|e| format!("Failed to read seed file '{}': {}", file_path.display(), e))?;
+
+        let records: Vec<serde_json::Value> = serde_json::from_str(&content)
+            .map_err(|e| format!("Failed to parse JSON in '{}': {}", file_path.display(), e))?;
+
+        insert_records(&txn, table_name, &records).await?;
+
+        println!(
+            "{}: Inserted {} records into '{}'",
+            "Success".green(),
+            records.len(),
+            table_name.cyan()
+        );
+    }
+
+    enable_fk_checks(&txn).await?;
+
+    txn.commit()
+        .await
+        .map_err(|e| format!("Failed to commit transaction: {}", e))?;
+
+    Ok(())
+}
+
+/// Dump all database tables (or a specific entity) to JSON seed files.
+///
+/// Queries each table and writes the rows as a pretty-printed JSON array
+/// to `seeds/{table_name}.json`.
+///
+/// # Errors
+///
+/// Returns an error if the database is unreachable or a table cannot be read.
+pub async fn dump(entity: Option<String>) -> Result<(), String> {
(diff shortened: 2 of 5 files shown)


rapina-cli/Cargo.toml
```diff
@@ -22,20 +22,27 @@ notify = { version = "8", default-features = false, features = [
   "macos_kqueue",
 ] }
 notify-debouncer-mini = "0.4"
+sea-orm = { version = "1", features = ["runtime-tokio-rustls"] }
+tokio = { version = "1", features = ["full"] }
+dotenvy = "0.15"
```
arferreira: `sea-orm` and `tokio` were optional or absent before this PR. Now every CLI invocation (`rapina new`, `rapina routes`, `rapina doctor`) pulls in sea-orm, sqlx, and a full tokio runtime even when the user has no database. These should be behind the `seed-*` feature gates, same pattern as the import features.

rapina-cli/src/main.rs
```diff
@@ -184,7 +217,8 @@ enum OpenapiCommands {
     },
 }
 
-fn main() {
+#[tokio::main]
+async fn main() {
```
arferreira: This forces a tokio runtime on every CLI invocation, including purely sync commands like `rapina new` and `rapina routes`. The seed commands should spin up a runtime only when needed, `tokio::runtime::Runtime::new().unwrap().block_on(...)` inside the `Seed` match arm, keeping `main` synchronous.

rapina-cli/src/commands/seed.rs
```diff
@@ -0,0 +1,965 @@
...
+    let txn = conn
+        .begin()
+        .await
+        .map_err(|e| format!("Failed to begin transaction: {}", e))?;
+
+    if fresh {
+        truncate_tables(&txn, &seed_files).await?;
+    }
+
+    disable_fk_checks(&txn).await?;
```
arferreira: This is called right after `truncate_tables`, but `truncate_tables` already calls `disable_fk_checks` / `enable_fk_checks` internally. So FK checks go disabled -> enabled -> disabled -> enabled. Either pull the FK logic out of `truncate_tables` and manage it once in `load`, or remove this standalone call.

rapina-cli/src/commands/seed.rs
```diff
@@ -0,0 +1,965 @@
...
+                quoted_table, cols, placeholders
+            )
+        }
+        DatabaseBackend::Postgres | DatabaseBackend::Sqlite => {
+            format!(
+                "INSERT INTO {} ({}) VALUES ({}) ON CONFLICT DO NOTHING",
+                quoted_table, cols, placeholders
+            )
+        }
+    }
```
arferreira, line 331: Without `--fresh`, existing rows with matching primary keys are silently skipped. That's probably not what the user expects — they'd want an upsert (`ON CONFLICT DO UPDATE`). At minimum document the behavior, but upsert semantics would be more useful as the default.

rapina-cli/src/commands/seed.rs
```diff
@@ -0,0 +1,965 @@
...
+                serde_json::json!(format!("https://example.com/{}", n % 10000))
+            } else {
+                serde_json::json!(format!("{}_{}", field_name, index))
+            }
+        }
+        "i32" => serde_json::json!((n % 1000) as i32),
+        "i64" => serde_json::json!((n % 100000) as i64),
+        "f32" | "f64" => serde_json::json!((n % 10000) as f64 / 100.0),
+        "bool" => serde_json::json!(n % 2 == 0),
+        "Uuid" => serde_json::json!(uuid::Uuid::new_v4().to_string()),
```
arferreira, line 583: You mentioned in the description deterministic generation using `rand::Rng`, but `Uuid::new_v4()` hits the OS RNG and produces different values on every run. What do you think about droping the determinism claim?

rapina-cli/Cargo.toml
```diff
@@ -22,20 +22,27 @@ notify = { version = "8", default-features = false, features = [
...
 ctrlc = "3"
 toml = "1.0"
 serde = { version = "1", features = ["derive"] }
 serde_json = "1"
 chrono = "0.4"
-tokio = { version = "1", features = ["full"], optional = true }
+rand = "0.10"
```
arferreira: This is a heavy dependency for what amounts to `rng.random::<u32>()`. `fastrand` (already pulled in transitively) would cover this with a smaller footprint.

rapina-cli/src/commands/seed.rs
```diff
@@ -0,0 +1,965 @@
...
+            entities.push((entity_name, fields));
+        }
+    }
+
+    if entities.is_empty() {
+        return Err("No schema! blocks found in src/entity.rs".to_string());
+    }
+
+    Ok(entities)
+}
```
arferreira, line 518: The line-by-line string matching will break on comments inside a `schema!` block, attributes on fields, or multi-line type annotations. Fine for a v1 since `schema!` has a known format, but worth a doc comment noting the limitation.

rapina-cli/src/commands/seed.rs
```diff
@@ -0,0 +1,965 @@
...
+        "Date" => serde_json::json!(format!("2025-01-{:02}", (n % 28) + 1)),
+        "Text" => serde_json::json!(format!("Sample {} text for record {}", field_name, index)),
+        _ => serde_json::Value::Null,
+    }
+}
+
+// -- Tests --
+
+#[cfg(test)]
+mod tests {
```
arferreira, line 599: Good coverage on the pure functions. Missing integration tests for the actual database path, `load`, `dump`, and `insert_records` are untested. A SQLite in-memory test would catch regressions in the generated SQL without needing external infrastructure.

arferreira (changes requested):
Solid work for a first contribution, and the scope is well chosen, thanks!
Just some nits to address and we can merge, once those are addressed, we'll also want to update the docs with a dedicated page for `rapina seed` covering all three subcommands, the expected seeds/ directory structure, the `schema!` format requirements, and the `DATABASE_URL` setup.

rapina-cli/src/commands/seed.rs
yan-pi: Fair point. should drop the determinism claim from the PR description.
The generate command should be treated as convenience tool for bootstrapping test data, not a reproducible fixture generator.

rapina-cli/src/commands/seed.rs
yan-pi: Yeah, you are right. Ive added a doc comment on load() explaining that without `--fresh`, existing rows with matching PKs are skipped (not updated). 

Full upsert needs PK discovery per table which adds non-trivial complexity. I can open an follow-up issue for upsert support. 

Does that work for v1?

arferreira: Works for v1, go ahead and open the follow-up issue.

rapina-cli/Cargo.toml
```diff
@@ -22,20 +22,28 @@ notify = { version = "8", default-features = false, features = [
...
 serde_yaml_ng = { version = "0.10", optional = true }
 
 [features]
 default = []
-import = ["tokio"]
+seed = ["dep:sea-orm", "dep:tokio", "dep:dotenvy"]
+seed-postgres = ["seed", "sea-orm/sqlx-postgres"]
+seed-mysql = ["seed", "sea-orm/sqlx-mysql"]
+seed-sqlite = ["seed", "sea-orm/sqlx-sqlite"]
+import = []
```
arferreira: Anyone doing `cargo install rapina-cli --features import-postgres` would fail to compile.
```suggestion
import = ["dep:tokio"]
```

rapina-cli/Cargo.toml
```diff
@@ -22,20 +22,28 @@ notify = { version = "8", default-features = false, features = [
...
+uuid = { version = "1", features = ["v4"] }
 sea-schema = { version = "0.16", optional = true, default-features = false }
 sqlx = { version = "0.8", optional = true }
 openapiv3 = { version = "2", optional = true }
 serde_yaml_ng = { version = "0.10", optional = true }
 
 [features]
 default = []
-import = ["tokio"]
+seed = ["dep:sea-orm", "dep:tokio", "dep:dotenvy"]
```
arferreira: As they're only used inside #[cfg(feature = "seed")], right?
```suggestion
seed = ["dep:sea-orm", "dep:tokio", "dep:dotenvy", "dep:fastrand", "dep:uuid"]
```

rapina-cli/Cargo.toml
```diff
@@ -22,20 +22,28 @@ notify = { version = "8", default-features = false, features = [
...
+tokio = { version = "1", features = ["full"], optional = true }
+dotenvy = { version = "0.15", optional = true }
 ctrlc = "3"
 toml = "1.0"
 serde = { version = "1", features = ["derive"] }
 serde_json = "1"
 chrono = "0.4"
-tokio = { version = "1", features = ["full"], optional = true }
+fastrand = "2.3.0"
+uuid = { version = "1", features = ["v4"] }
```
arferreira: Every CLI install pulls them in for nothing, right?
```suggestion
fastrand = { version = "2.3.0", optional = true }
uuid = { version = "1", features = ["v4"], optional = true }
```

PR #331, opened 2026-03-12
Add #[patch] proc-macro and Router::patch() shorthand

### LINKED ISSUE #323: Add #[patch] route macro and Router::patch() shorthand
The router supports GET, POST, PUT, and DELETE as both proc macro attributes and shorthand methods, but PATCH is missing from the surface API. Every layer underneath already handles it correctly: the trie maps it to index 6, the static map accepts any `http::Method`, CORS includes it in default allowed methods, the cache layer treats it as a mutation, `TestClient` has `.patch()`, and the OpenAPI generator already iterates over it.

The only things missing are `Router::patch()` and `Router::patch_named()` in `router/mod.rs`, the `#[patch]` attribute macro in `rapina-macros/src/lib.rs`, and the prelude export in `lib.rs`. The OpenAPI importer currently works around this by mapping PATCH to PUT at `import_openapi.rs:689`, which should be updated once the shorthand exists.

uemuradevexe:
## Summary

wxposes PATCH to the public API: adds `#[patch]` macro and `Router::patch()` / `Router::patch_named()`

Removes the PATCH→PUT workaround from the OpenAPI importer

## Related Issues

Closes #323 

## Checklist

- [x] `cargo fmt` passes
- [x] Tests pass

### DIFF
--- a/rapina-macros/src/lib.rs
+++ b/rapina-macros/src/lib.rs
@@ -87,6 +87,13 @@ pub fn put(attr: TokenStream, item: TokenStream) -> TokenStream {
 
 /// Registers a PATCH route handler.
 ///
+/// # Example
+///
+/// ```ignore
+/// #[patch("/users/:id")]
+/// async fn update_user(Path(id): Path<u64>) -> Json<User> { /* ... */ }
+/// ```
+///
 /// See [`get`] for syntax details including the optional `group` parameter.
 #[proc_macro_attribute]
 pub fn patch(attr: TokenStream, item: TokenStream) -> TokenStream {
--- a/docs/content/docs/core-concepts/routing.md
+++ b/docs/content/docs/core-concepts/routing.md
@@ -19,6 +19,7 @@ let router = Router::new()
     .get("/about", about)
     .post("/users", create_user)
     .put("/users/:id", update_user)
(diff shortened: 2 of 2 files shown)


rapina/src/router/mod.rs
```diff
@@ -206,6 +206,38 @@ impl Router {
         )
     }
 
+    /// Adds a PATCH route with a handler name.
+    pub fn patch_named<F, Fut, Out>(self, pattern: &str, handler_name: &str, handler: F) -> Self
```
arferreira, line 210: Nice that you added both `patch` and `patch_named`. Worth noting that `put_named` and `delete_named` don't exist either, same gap, different methods. Not something to fix in this PR but might be worth a follow-up issue.

rapina-macros/src/lib.rs
```diff
@@ -85,6 +85,14 @@ pub fn put(attr: TokenStream, item: TokenStream) -> TokenStream {
     route_macro("PUT", attr, item)
 }
 
+/// Registers a PATCH route handler.
+///
```
arferreira, line 89: Worth adding a `#[patch("/users/:id")]` handler example.

arferreira (approved):

PR #333, opened 2026-03-12
Add feature flag for compression

### LINKED ISSUE #219: Gate compression middleware behind a feature flag
The `flate2` dependency (used by compression middleware) is always compiled in, even for users who don't need compression. This adds compile time and binary size unnecessarily.

**What to do:**

1. Add a `compression` feature flag to `rapina/Cargo.toml`
2. Make `flate2` an optional dependency gated behind it
3. Gate the compression middleware module with `#[cfg(feature = "compression")]`
4. Gate the `with_compression()` builder method on `Rapina`
5. Add `compression` to the default features so existing users aren't broken
6. Update any docs or examples that reference compression
7. Make sure `cargo test`, `cargo test --no-default-features`, and `cargo clippy` all pass

juv:
## Summary

Adds a feature flag for compression, enabled by default

## Related Issues

Closes #219

## Checklist

- [ ] `cargo fmt` passes
- [ ] Tests pass

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -59,7 +59,6 @@ jobs:
           - "--no-default-features"
           - "--features database"
           - "--features metrics"
-          - "--features compression"
     steps:
       - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5  # v4
       - uses: dtolnay/rust-toolchain@efa25f7f19611383d5b0ccf2d1c8914531636bf9
--- a/rapina/Cargo.toml
+++ b/rapina/Cargo.toml
@@ -55,16 +55,11 @@ jsonwebtoken = { version = "10.3.0", features = ["rust_crypto"] }
 # Rate Limiting
 dashmap = "6.1.0"
 
-<<<<<<< feature/add-compression-feature-flag
-# Compression (optional)
-flate2 =  { version = "1.1", optional = true }
-=======
 # SmallVec for stack-allocated path params
 smallvec = "1"
 
-# Compression
-flate2 = "1.1"
->>>>>>> main
+# Compression (optional)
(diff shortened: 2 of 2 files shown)


rapina/Cargo.toml
```diff
@@ -91,11 +91,12 @@ serial_test = "3"
...
+default = ["compression"]
+compression = ["flate2"]
 database = ["sea-orm", "sea-orm-migration", "async-trait"]
 postgres = ["database", "sea-orm/sqlx-postgres", "sea-orm-migration/sqlx-postgres"]
 mysql = ["database", "sea-orm/sqlx-mysql", "sea-orm-migration/sqlx-mysql"]
 sqlite = ["database", "sea-orm/sqlx-sqlite", "sea-orm-migration/sqlx-sqlite"]
 metrics = ["prometheus"]
 cache-redis = ["redis"]
-websocket = ["hyper-tungstenite", "tokio-tungstenite", "futures-util"]
+websocket = ["hyper-tungstenite", "tokio-tungstenite", "futures-util"]
```
arferreira: Looks like the trailing newline got dropped here, mind adding it back? some tooling gets weird without it
```suggestion
websocket = ["hyper-tungstenite", "tokio-tungstenite", "futures-util"]

```

.github/workflows/ci.yml
```diff
@@ -59,6 +59,7 @@ jobs:
           - "--no-default-features"
           - "--features database"
           - "--features metrics"
+          - "--features compression"
```
arferreira: Since compression is in `default` now, doesn't the "" entry already cover this? what does `--features compression` on top of defaults give us that we're not already testing?
```suggestion
```

juv: good catch, i think it can be removed

PR #347, opened 2026-03-13
Use RFC7807 Problem Details for error responses

### LINKED ISSUE #171: Use RFC 7807 Problem Details for error responses
Our error envelope is currently:

```json
{
  "error": { "code": "NOT_FOUND", "message": "user not found" },
  "trace_id": "550e8400-..."
}
```

RFC 7807 (Problem Details for HTTP APIs) is the standard for this. It looks like:

```json
{
  "type": "https://userapina.com/errors/not-found",
  "title": "Not Found",
  "status": 404,
  "detail": "user not found",
  "instance": "/users/123",
  "trace_id": "550e8400-..."
}
```

The Content-Type becomes `application/problem+json`. Most API consumers and tooling recognize this format. It also plays well with OpenAPI since there's a standard schema for it.

This would be a breaking change to the error format, so it should land before a stable version. We can keep the existing error constructors (`Error::not_found()`, etc.) and just change what gets serialized. The `trace_id` field can be an extension property which RFC 7807 explicitly supports.

Worth considering whether to make this opt-in (`.with_rfc7807_errors(true)`) for backwards compatibility during the alpha/beta period, or just switch the default.

Ismaellima4:
## Summary

What does this PR do?
- Introduced opt-in support for RFC 7807 Problem Details error format.
- Refactored Error struct to use Box<ErrorInner> to optimize size.
- Modularized error handling into standard and rfc7807 submodules.
- Updated into_response to conditionally return RFC 7807 format.
- Added enable_rfc7807_errors() to the application builder.
## Related Issues

Closes #171 

## Checklist

- [x] `cargo fmt` passes
- [x] Tests pass


rapina/src/app.rs
```diff
@@ -72,6 +72,8 @@ pub struct Rapina {
     /// Relay configuration (if enabled)
     #[cfg(feature = "websocket")]
     pub(crate) relay_config: Option<crate::relay::RelayConfig>,
+    /// Whether to use RFC 7807 Problem Details for error responses (default: true)
```
arferreira: Is this doc accurate? Looks like `rfc7807_errors` defaults to `false` in `new()` and `enable_rfc7807_errors()` says "disabled by default." Should this say `(default: false)`?

rapina/src/error.rs
```diff
@@ -27,31 +27,116 @@
...
 use http_body_util::Full;
 
-/// The JSON structure returned for error responses.
-#[derive(Debug, Serialize)]
-pub struct ErrorResponse {
-    /// The error details.
-    pub error: ErrorDetail,
-    /// Unique identifier for request tracing.
-    pub trace_id: String,
+static USE_RFC7807: AtomicBool = AtomicBool::new(false);
```
arferreira: Have you considered storing this on `AppState` instead of a process-wide static? Right now if someone spins up two `Rapina` instances in the same process (or in tests), the last `prepare()` call wins for both. It also means every integration test needs `#[serial]` even if it has nothing to do with error formatting, that's a pretty big blast radius. Not blocking, but worth thinking about for v2.

rapina/src/error.rs
```diff
@@ -27,31 +27,116 @@
...
+        #[serde(skip_serializing_if = "Option::is_none")]
+        pub instance: Option<String>,
+        /// Unique identifier for request tracing.
+        pub trace_id: String,
+        /// Optional additional error details.
+        #[serde(flatten, skip_serializing_if = "Option::is_none")]
+        pub extensions: Option<serde_json::Value>,
+    }
+
+    const BASE_ERROR_URI: &str = "https://userapina.com/errors";
```
arferreira: Should this be configurable? Apps built on Rapina will all have `type` URIs pointing to `userapina.com` instead of their own domain. RFC 7807 expects the `type` URI to be dereferenceable documentation. Maybe something like `.rfc7807_base_uri("https://myapp.com/errors")` on the builder?

arferreira (commented):
Left three minor notes, the doc comment typo is the only one that should be fixed before merge, the other two are future considerations, also `fmt` and `clippy`.

arferreira (approved):

PR #348, opened 2026-03-13
Wrap State<T> extractor in Arc<T> to remove Clone bound

### LINKED ISSUE #327: State<T> deep clones on every request, should expose the inner Arc
Values registered in AppState are already stored as `Arc<dyn Any>` in the internal map. But `AppState::get<T>()` calls `downcast_ref()`, which returns a plain `&T`, and then `State<T>`'s extractor calls `value.clone()` on that reference. Every request that extracts `State<AppConfig>` deep-clones the entire config struct, even though the data is already behind an Arc that could be cheaply shared.

For a small struct this is negligible. For a `State<AppConfig>` with connection pools, feature flags, or any non-trivial data, it's a wasteful deep copy on every request. The Clone bound on T is also unnecessarily restrictive since it prevents using State with types that are expensive or impossible to clone.

The fix is to either have `State<T>` wrap an `Arc<T>` instead of a `T` (so extraction is an Arc clone, one atomic increment), or to add a companion method on AppState like `get_arc<T>() -> Option<Arc<T>>` that hands out the Arc directly. The current `get<T>() -> Option<&T>` can stay for convenience, but the extractor should take the cheap path.

This is the same pattern axum uses with its `State<T>`,  the value is wrapped in an Arc at the framework level, extraction is always a reference count bump, never a deep copy.

lucaserm:
## Summary

State extraction now bumps an atomic reference count instead of deep-cloning the inner value.  This drops the `Clone` requirement on types stored via `.state()` and adds `AppState::get_arc` for the underlying lookup.

## Related Issues

Closes #327

## Checklist

- [x] `cargo fmt` passes
- [x] Tests pass


arferreira (commented):
Clean change, while you're here, there are a few docs and examples still showing `#[derive(Clone)]` on state structs that no longer need it:

- `docs/content/docs/getting-started/configuration`
- `rapina/examples/auth`
- `rapina/examples/todo_api`
- `rapina/examples/macros`
- `rapina/examples/websocket-chat/src/main`

Would be good to clean those up in this same PR so the examples match the new behavior, then we can merge.

PR #351, opened 2026-03-14
Add --force flag to `import database` for re-importing

### LINKED ISSUE #241: Add --force flag for re-importing over existing files
Follow-up from #170 / #240.

Currently `rapina import database` fails silently or overwrites if files already exist. A `--force` flag would let users explicitly re-import, useful when the upstream database schema changes.

lucaserm:
## Summary

Added a --force flag to rapina import database that allows re-importing tables when generated files already exist. Previously the command would  hard-error if src/<plural>/ already existed.

What --force does:
  - Removes and re-creates existing feature module directories (src/<plural>/)
  - Strips duplicate schema! blocks from entity.rs before appending the updated one
  - Without --force, behavior is unchanged (errors on conflict)

Tests added:
  - 3 tests for remove_schema_block (match, no-match, with attributes)
  - 2 tests for create_feature_module (error without force, overwrite with force)
  - 2 tests for update_entity_file (dedup with force, append without force)

## Related Issues

Closes #241

## Checklist

- [x] `cargo fmt` passes
- [x] Tests pass

PS: Documented the rapina import database command (was previously undocumented entirely)

### DIFF
--- a/rapina/src/middleware/cors.rs
+++ b/rapina/src/middleware/cors.rs
@@ -188,3 +188,236 @@ impl Middleware for CorsMiddleware {
         })
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use http::{HeaderValue, Method, StatusCode, header};
+
+    // --- Config constructors ---
+
+    #[test]
+    fn test_permissive_config() {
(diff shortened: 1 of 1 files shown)


arferreira (approved):
LGTM

PR #358, opened 2026-03-14
Enables multiple path parameters and updates the docs

### LINKED ISSUE #324: Support Path<(A, B)> tuple and Path<MyStruct> extraction for multiple path params
`Path<T>` currently calls `params.iter().next()` and parses that single value. Works fine for one param, undefined behavior for two or more. Anyone building nested resource APIs (`/orgs/:org_id/teams/:team_id/members/:member_id`) has to drop down to raw `PathParams` and call `.get()` manually.

Two extraction modes should work:

**Tuple extraction** follows declaration order in the route pattern. For `#[get("/users/:uid/posts/:pid")]`, `Path<(u64, u64)>` gives you `(uid_value, pid_value)` in the order the params appear in the pattern. This works because PathParams is now backed by SmallVec (PR #321), which preserves insertion order, and the trie inserts left to right during traversal.

**Struct extraction** matches by field name. For the same route, `Path<MyParams>` where `MyParams { uid: u64, pid: u64 }` would look up each field name in the params map. This requires the struct to implement `Deserialize` or a custom trait that maps field names to param keys. axum uses `serde::Deserialize` for this, which is reasonable since serde is already a dependency.

The ordering question for tuples: yes, it follows declaration order in the pattern, not alphabetical, not argument order in the handler signature. The trie walks the path left to right and pushes params in that order, SmallVec preserves it, so tuple index 0 is the first `:param` in the pattern.

The single param case (`Path<u64>`) should keep working exactly as it does today.

ShiraiEd:
### Add multiple path parameters support with tupple extraction

**Problem:** `Path<T>` only supported a single URL segment. Routes with multiple dynamic segments (e.g. `/orgs/:org_id/repos/:repo_id`) had no extraction mechanism.

## Changes

- **`rapina/src/extract.rs`** — Added `impl_multiple_params!` macro that implements `FromRequestParts` for `Path<(T1, T2, ...)>` tuples up to 5 parameters. Parameters are extracted positionally in URL order, each parsed independently via `FromStr`. Also introduced a sealed trait (`SingleParam`) to resolve Rust coherence conflicts between the existing single-param blanket impl and the new tuple impls.

- **`rapina-macros/src/lib.rs`** — Fixed `route_macro_core` to support any pattern in handler arguments, not just `Pat::Ident`. Previously, destructuring patterns like `Path((org_id, team_id)): Path<(u64, u64)>` would panic. Fix extracts to a temp variable `__rapina_arg_N` then binds the original pattern.

- **`rapina/examples/multi_path_params.rs`** — New example demonstrating single, 2-param, 3-param, and mixed-type extraction.

- **`docs/content/docs/core-concepts/routing.md`** — New "Multiple Path Parameters" subsection with usage examples.

## Usage

```rust
#[get("/orgs/:org_id/repos/:repo_id/issues/:issue_id")]
async fn get_issue(Path((org_id, repo_id, issue_id)): Path<(u64, u64, u64)>) -> String {
    format!("org={} repo={} issue={}", org_id, repo_id, issue_id)
}
```
Closes #324

## Checklist

- [x] `cargo fmt` passes
- [x] Tests pass

### DIFF
--- a/rapina/src/extract.rs
+++ b/rapina/src/extract.rs
@@ -66,7 +66,7 @@ pub struct Json<T>(pub T);
 /// }
 /// ```
 ///
-/// **Three or more** — same tuple syntax, up to 5 parameters:
+/// **Three or more** — same tuple syntax:
 /// ```ignore
 /// #[get("/orgs/:org_id/teams/:team_id/members/:member_id")]
 /// async fn get_member(Path((org_id, team_id, member_id)): Path<(u64, u64, u64)>) -> String {


rapina/src/extract.rs
```diff
@@ -611,7 +625,17 @@ impl<T: DeserializeOwned + Send> FromRequestParts for Cookie<T> {
...
+    pub trait SingleParam: std::str::FromStr + Send {}
+}
+macro_rules! impl_single_param {
+    ($($t:ty),+) => { $(impl sealed::SingleParam for $t {})+ };
+}
+impl_single_param!(
+    u8, u16, u32, u64, u128, i8, i16, i32, i64, i128, f32, f64, bool, String
+);
+
+impl<T: sealed::SingleParam> FromRequestParts for Path<T>
```
arferreira: This is a regression, the sealed trait restricts `Path<T>` to only the listed primitives, which breaks `Path<Uuid>` and any other custom type that implements `FromStr`.
You can just keep the original bound:
```suggestion
impl<T: FromStr + Send> FromRequestParts for Path<T>
```

ShiraiEd: In this way the code don't compiles, the impls conflict,
I can try the serde method instead of this macro one

arferreira: Alright let me check this brb

rapina/src/extract.rs
```diff
@@ -46,21 +46,35 @@ use http::header::CONTENT_TYPE;
...
+///
+/// **Tuple** — extracts multiple parameters in declaration order (left to right in the pattern):
+/// ```ignore
+/// #[get("/orgs/:org_id/teams/:team_id")]
+/// async fn get_team(Path((org_id, team_id)): Path<(u64, u64)>) -> String {
+///     format!("org={} team={}", org_id, team_id)
+/// }
+/// ```
+///
+/// **Three or more** — same tuple syntax, up to 5 parameters:
```
arferreira:
```suggestion
/// **Three or more** — same tuple syntax:
```

PR #361, opened 2026-03-14
Add configurable verbosity and header redaction to RequestLogMiddleware

### LINKED ISSUE #165: Improve request/response logging
RequestLogMiddleware currently logs method, path, status, and duration at INFO level. That's fine for development but not enough for production.

Two things would make this much more useful:

1. **Configurable verbosity** — option to log request/response headers, body size, query params. Maybe something like `RequestLogMiddleware::verbose()` vs the current `::new()`. You don't want full body logging by default, but sometimes you really need it when debugging.

2. **Sensitive field redaction** — when logging at higher verbosity, automatically redact Authorization headers, cookie values, and fields like "password" or "token" in request bodies. Nobody wants credentials in their log aggregator.

The tracing infrastructure is already there, this is mostly about adding configuration options to the existing middleware.

arferreira:
Closes #165.

`RequestLogMiddleware` was a zero-config unit struct that logged method, path, status, and duration. Useful for dev, not enough for production debugging.

This adds `RequestLogConfig` with three opt-in fields: header logging (request and response, with case-insensitive redaction), query parameter logging, and body size via `content-length`. Sensitive headers (`authorization`, `cookie`, `set-cookie`, `x-api-key`) are redacted by default when using `RequestLogConfig::verbose()`. All extra fields are emitted as structured tracing fields so they flow through whatever subscriber is configured.

`RequestLogMiddleware::new()` still compiles and behaves exactly as before, no extra output unless you ask for it. For richer logging there's `RequestLogMiddleware::verbose()`, `RequestLogMiddleware::with_config(config)`, and a `Rapina::with_request_log(config)` convenience method that follows the same pattern as `with_cors` and `with_rate_limit`.

Body field redaction is intentionally out of scope. The request body is `hyper::body::Incoming`, inspecting it means buffering and consuming the stream before the handler sees it. That's a full copy per request, breaks non-JSON content types, and adds latency on every request. Body-level redaction belongs in structured logging serializers or the application layer.

Documentation updated with a dedicated Request Logging section in the middleware guide, and the middleware example now uses `verbose()` to demonstrate the new API.

### DIFF
--- a/docs/content/docs/core-concepts/database.md
+++ b/docs/content/docs/core-concepts/database.md
@@ -192,6 +192,9 @@ Relationships are inferred from types:
 | `#[timestamps(created_at)]` | Only include `created_at` timestamp |
 | `#[timestamps(updated_at)]` | Only include `updated_at` timestamp |
 | `#[timestamps(none)]` | No automatic timestamps |
+| `#[primary_key(col1, col2)]` | Define a custom primary key (single or composite) |
+
+When `#[primary_key(...)]` is set, the auto-generated `id: i32` field is **not** added — you must declare the primary key columns yourself as regular fields.
 
 ```rust
 #[table_name = "people"]
@@ -203,6 +206,22 @@ Person {
 AuditLog {
     action: String,
     timestamp: NaiveDateTime,  // manage your own timestamp (no timezone)
+}
+
+// UUID primary key instead of auto-increment i32
+#[primary_key(id)]
+Product {
+    id: Uuid,
+    name: String,
+    price: f64,
+}
+
+// Composite primary key (join table)
+#[primary_key(user_id, role_id)]
+#[timestamps(none)]
+UserRole {
+    user_id: User,     // belongs_to User
+    role_id: Role,     // belongs_to Role
 }
 ```
 
--- a/rapina-cli/src/commands/codegen.rs
+++ b/rapina-cli/src/commands/codegen.rs
@@ -26,7 +26,109 @@ pub(crate) fn to_pascal_case(s: &str) -> String {
         .collect()
 }
 
+/// Irregular plural forms: (singular, plural)
+const IRREGULARS: &[(&str, &str)] = &[
+    // Common irregular plurals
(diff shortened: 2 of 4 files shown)


rapina/src/middleware/request_log.rs
```diff
@@ -46,9 +167,33 @@ impl Middleware for RequestLogMiddleware {
...
+                        .map(|s| s.to_string())
+                } else {
+                    None
+                };
+
                 info!(
                     status = status,
                     duration_ms = duration.as_millis() as u64,
+                    request_headers = req_headers.as_deref().unwrap_or_default(),
+                    response_headers = res_headers.as_deref().unwrap_or_default(),
```
lucaserm: When a flag is disabled, the fields are logged as empty strings ("") rather than being omitted entirely, this means every request emits 5 extra tracing fields even in the default zero-config mode, just with empty values. This adds noise to structured log output (JSON subscribers will emit "request_headers":"" etc). Not a bug, but worth remember.

rapina/src/middleware/request_log.rs
```diff
@@ -7,12 +7,112 @@ use crate::response::BoxBody;
...
+    }
+
+    /// Append a header name to the redaction list.
+    pub fn redact_header(mut self, name: impl Into<String>) -> Self {
+        self.redacted_headers.push(name.into());
+        self
+    }
+}
+
+fn default_redacted_headers() -> Vec<String> {
```
lucaserm, line 63: The proxy-authorization header also carries credentials and is missing, not a big deal but worth to know.

rapina/src/middleware/request_log.rs
arferreira: Good catch. Default mode now uses a separate info! call with just status and duration_ms, no empty string fields. The extra fields only show up when at least one flag is enabled.

rapina/src/middleware/request_log.rs
arferreira: Added!

PR #362, opened 2026-03-14
Add --bless snapshot testing for API responses

### LINKED ISSUE #263: rapina test --bless for snapshot testing
Add a --bless flag to rapina test that captures API response bodies as golden files and compares against them on subsequent runs.

When you have dozens of endpoints, asserting on individual fields gets tedious. Snapshot testing catches unexpected changes in response shape without per-field assertions — similar to insta, but built into the framework's test runner.

How it would work:

`rapina test` --bless runs tests, saves response snapshots to a snapshots/ directory
`rapina test` compares responses against saved snapshots, fails on diff with a clear output
Dynamic fields like UUIDs, timestamps, and trace_ids get auto-redacted so snapshots stay stable

lucaserm:
## Summary

It adds --bless snapshot testing to rapina, spanning both the CLI and framework crates.

  CLI side (rapina-cli):
  - Adds --bless flag to rapina test
  - When set, passes RAPINA_BLESS=1 env var to the cargo test subprocess
  - Prints an info message when blessing mode is active

  Framework side (rapina):
  - New rapina/src/testing/snapshot.rs module with:
    - Redaction — walks JSON trees replacing UUIDs with [UUID], ISO 8601 timestamps with [TIMESTAMP], and trace_id keys with [UUID] regardless of value
    - Snapshot format — HTTP {status} {reason}\nContent-Type: ...\n\n{body}\n
    - Bless mode — writes .snap files to a snapshots/ directory
    - Compare mode — reads existing .snap files, panics with a line-by-line diff on mismatch
    - 18 unit tests covering redaction, formatting, diffing, and file I/O (bless creates file, match passes, mismatch panics, missing panics)
  - Adds .assert_snapshot("name") method to TestResponse so users can write:
  let response = client.get("/users/1").send().await;
  response.assert_snapshot("get_user");
  - Adds regex as a dependency, tempfile as a dev-dependency

## Related Issues

Closes #263

## Checklist

- [x] `cargo fmt` passes
- [x] Tests pass


rapina/src/testing/snapshot.rs
```diff
@@ -0,0 +1,362 @@
...
+}
+
+/// Format a snapshot with a status line, content type, and body.
+pub fn format_snapshot(status: u16, body: &str) -> String {
+    let reason = reason_phrase(status);
+    let content_type = if body.starts_with('{') || body.starts_with('[') {
+        "application/json"
+    } else {
+        "text/plain"
+    };
```
arferreira: Hmm, this is a bit ugly. Pretty sure you can pass content type here and get that from the real headers.

rapina/src/testing/snapshot.rs
```diff
@@ -0,0 +1,362 @@
...
+        405 => "Method Not Allowed",
+        409 => "Conflict",
+        422 => "Unprocessable Entity",
+        429 => "Too Many Requests",
+        500 => "Internal Server Error",
+        502 => "Bad Gateway",
+        503 => "Service Unavailable",
+        _ => "Unknown",
+    }
+}
```
arferreira, line 157: Okay, this is also odd, first because it's not covering all the available status code, If someone gets a `418 I'm a Teapot` or `307 Temporary Redirect`, the snapshot says HTTP 418 Unknown instead of the real reason phrase.

The `http` crate (which rapina already depends on) has this built in. What about this?

```suggestion
fn reason_phrase(status: u16) -> &'static str {
    http::StatusCode::from_u16(status)
          .ok()
          .and_then(|s| s.canonical_reason())
          .unwrap_or("Unknown")
}
```

rapina/Cargo.toml
```diff
@@ -41,6 +41,7 @@ tracing = "0.1"
 tracing-subscriber = { version = "0.3", features = ["json", "env-filter"] }
 
 uuid = { version = "1", features = ["v4"] }
+regex = "1"
```
arferreira, line 45: Worth a `FIXME:` for a follow-up, so users who don't use snapshot testing don't pay the compile cost.
```suggestion
# FIXME: move behind a `snapshot` feature flag
regex = "1"
```

arferreira (changes requested):
Just small things to be addressed and we can merge. Good work so far!

rapina/src/testing/snapshot.rs
lucaserm: done!

rapina/src/testing/snapshot.rs
lucaserm: done!

rapina/Cargo.toml
lucaserm: done!

PR #363, opened 2026-03-14
Multipart file upload extractor

### LINKED ISSUE #329: Multipart file upload extractor
There's no way to handle file uploads in rapina. No multipart extractor, no streaming field access, nothing. This is the most common feature request for any web framework and a hard blocker for anyone building forms, file storage, or media upload APIs.

multer is the right crate for this. It's async-native, well-tested, and what axum uses internally. It works directly with hyper's streaming body so files don't need to be buffered entirely in memory before processing.

A `Multipart` extractor that implements `FromRequest` (body-consuming) would give access to individual fields as an async stream. Each field exposes its name, content type, filename, and a streaming reader for the bytes. Something like:

```rust
#[post("/upload")]
async fn upload(mut form: Multipart) -> StatusCode {
    while let Some(field) = form.next_field().await? {
        let name = field.name().unwrap_or_default();
        let data = field.bytes().await?;
        // ...
    }
    StatusCode::OK
}
```

multer should be an optional dependency behind a feature gate since not every app needs multipart parsing.

Ismaellima4:
## Summary

What does this PR do?

- rapina/src/extract/multipart.rs: Implemented [Multipart] extractor based on `multer` for efficient streaming of multipart form-data.
- rapina-macros/src/lib.rs: Fixed a bug where the `mut` keyword was stripped from handler arguments, enabling the use of mutable extractors.
- rapina/src/extract/multipart.rs: Added comprehensive documentation and usage examples for [Multipart]
- rapina/Cargo.toml: Configured to specify `required-features` for the multipart example.
- rapina/examples/multipart.rs: Enhanced example to demonstrate reading fields and file data.

## Related Issues

Closes #329 

## Checklist

- [x] `cargo fmt` passes
- [x] Tests pass


arferreira (commented):
Nice work! The `chunk()` method exists for streaming, but a default max size would be safer. Worth a follow-up issue, if you can write down, please.
Also, just needs a rebase against main to fix conflicts, then we can merge.

arferreira (approved):

PR #366, opened 2026-03-16
feat: add built-in health check endpoints

### LINKED ISSUE #162: Add health check endpoint
Right now there's no built-in health check endpoint. Anyone deploying Rapina behind a load balancer or on Kubernetes needs one.

The idea is a `/__rapina/health` endpoint (following the existing `/__rapina/` convention) that returns 200 when the server is healthy. Should be enabled with something like `.with_health_check(true)` on the builder.

Basic version just returns `{"status": "ok"}`. But ideally it also checks database connectivity when the database feature is enabled, and supports registering custom health checks for things like Redis, external services, etc.

Kubernetes expects both liveness and readiness probes — worth considering whether those should be separate endpoints (`/__rapina/health/live`, `/__rapina/health/ready`) or a single endpoint with different semantics.

ShiraiEd:
Adds `/__rapina/health`, `/__rapina/health/live`, and `/__rapina/health/ready` endpoints enabled via `.with_health_check(true)` on the builder.

- Liveness probe always returns 200 — never checks external deps
- Readiness probe runs DB connectivity check (when `database` feature is enabled) and any custom checks registered via `.add_health_check()`
- Failed checks return 503 with a `"checks"` map identifying which ones failed
- All `/__rapina/*` routes are automatically public (bypass JWT auth)
- All project templates (`rest-api`, `auth`, `crud`) now include `.with_health_check(true)` by default
- Updated examples, docs, and removed the now-redundant manual `/health` handlers throughout the codebase

Closes #162 

## Checklist

- [x] `cargo fmt` passes
- [x] Tests pass

### DIFF
--- a/rapina/src/app.rs
+++ b/rapina/src/app.rs
@@ -5,6 +5,7 @@ use std::net::SocketAddr;
 use std::time::Duration;
 
 use crate::auth::{AuthConfig, AuthMiddleware, PublicRoutes};
+use crate::health::{HealthRegistry, health_check, liveness_check, readiness_check};
 use crate::introspection::{RouteRegistry, list_routes};
 #[cfg(feature = "metrics")]
 use crate::metrics::{MetricsMiddleware, MetricsRegistry, metrics_handler};
@@ -55,6 +56,10 @@ pub struct Rapina {
     pub(crate) middlewares: MiddlewareStack,
     /// Whether introspection is enabled.
     pub(crate) introspection: bool,
+    /// Whether health check endpoint is enabled
+    pub(crate) health_check: bool,
+    /// Custom health checks
+    pub(crate) health_registry: HealthRegistry,
     /// Whether metrics is enabled.
     pub(crate) metrics: bool,
     /// Whether OpenAPI is enabled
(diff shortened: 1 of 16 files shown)


rapina/src/app.rs
```diff
@@ -310,6 +316,32 @@ impl Rapina {
         self
     }
 
+    /// Enables or disables the health check endpoint.
+    ///
+    /// When enabled, a `GET /__rapina/health` endpoint is registered
+    /// that returns a simple `200 OK` response.
```
arferreira: Perhaps, registers readiness and liveness health check endpoints, because is an alias for `readiness_check`?

rapina/src/health/endpoint.rs
```diff
@@ -0,0 +1,263 @@
...
+            let db_ok = conn
+                .execute(Statement::from_string(backend, "SELECT 1"))
+                .await
+                .is_ok();
+            checks.insert(
+                "db".to_string(),
+                if db_ok {
+                    "ok".into()
+                } else {
+                    "unreachable".into()
```
arferreira: Let's return just `error` for consistency, DB failure is returning `unreachable`, for redis as well.

arferreira: I mean, a boolean check returning `false` doesn't mean `unreachable`, just mean unhealthy right?

rapina/src/health/endpoint.rs
```diff
@@ -0,0 +1,263 @@
...
+                if ok { "ok".into() } else { "error".into() },
+            );
+            if !ok {
+                all_ok = false;
+            }
+        }
+    }
+
+    #[cfg(not(feature = "database"))]
+    let _ = state;
```
arferreira: Dead code?

PR #368, opened 2026-03-16
feat: gate rate limiting middleware behind `rate-limit` feature flag

### LINKED ISSUE #220: Gate rate limiting behind a feature flag
The `dashmap` dependency (used by rate limiting middleware) is always compiled in, even for users who don't need rate limiting.

**What to do:**

1. Add a `rate-limit` feature flag to `rapina/Cargo.toml`
2. Make `dashmap` an optional dependency gated behind it
3. Gate the rate limit middleware module with `#[cfg(feature = "rate-limit")]`
4. Gate `with_rate_limit()` and related builder methods
5. Add `rate-limit` to the default features so existing users aren't broken
6. Update any docs or examples that reference rate limiting
7. Make sure `cargo test`, `cargo test --no-default-features`, and `cargo clippy` all pass

lucaserm:
## Summary

Add a `rate-limit` feature (enabled by default) that gates the rate limiting module, `with_rate_limit()` builder method, `Error::rate_limited()`, and related prelude re-exports. Users who don't need rate limiting can disable the feature to exclude that code path.

## Related Issues

Closes #220

## Checklist

- [x] `cargo fmt` passes
- [x] Tests pass

### DIFF
--- a/rapina/Cargo.toml
+++ b/rapina/Cargo.toml
@@ -54,7 +54,7 @@ schemars = { version = "1.2.0", features = ["chrono04", "uuid1", "rust_decimal1"
 # JWT Authentication
 jsonwebtoken = { version = "10.3.0", features = ["rust_crypto"] }
 
-# Concurrent map (used by cache + optional rate limiting)
+# Concurrent map (used by cache, relay, and optional rate limiting)
 dashmap = "6.1.0"
 
 # SmallVec for stack-allocated path params
--- a/rapina/src/lib.rs
+++ b/rapina/src/lib.rs
@@ -67,7 +67,7 @@
 //! - [`BodyLimitMiddleware`](middleware::BodyLimitMiddleware) - Limit request body size
 //! - [`TraceIdMiddleware`](middleware::TraceIdMiddleware) - Add trace IDs to requests
 //! - [`RequestLogMiddleware`](middleware::RequestLogMiddleware) - Structured request logging
-//! - [`RateLimitMiddleware`](middleware::RateLimitMiddleware) - Token bucket rate limiting (requires `rate-limit` feature)
+//! - `RateLimitMiddleware` - Token bucket rate limiting (requires `rate-limit` feature)
 //!
 //! ## Introspection
 //!


rapina/Cargo.toml
```diff
@@ -103,8 +107,9 @@ harness = false
 windows-sys = { version = "0.61.2", features = ["Win32_System_Console"] }
 
 [features]
-default = ["compression"]
+default = ["compression", "rate-limit"]
 compression = ["flate2"]
+rate-limit = []
```
ShiraiEd, line 119:
```toml
rate-limit = ["dashmap"]
```

rapina/Cargo.toml
```diff
@@ -54,7 +54,7 @@ schemars = { version = "1.2.0", features = ["chrono04", "uuid1", "rust_decimal1"
 # JWT Authentication
 jsonwebtoken = { version = "10.3.0", features = ["rust_crypto"] }
 
-# Rate Limiting
+# Concurrent map (used by cache + optional rate limiting)
 dashmap = "6.1.0"
```
ShiraiEd, line 58:
```toml
dashmap = { version = "6.1.0", optional = true }
```

rapina/src/middleware/mod.rs
```diff
@@ -14,6 +14,7 @@ mod body_limit;
 #[cfg(feature = "compression")]
 mod compression;
 mod cors;
+#[cfg(feature = "rate-limit")]
```
arferreira, line 17: Hmmm, I don't know, but our error constructor is a pure function that just creates a `429` response, I mean, it has zero dependency on the rate limiting middleware code.

rapina/Cargo.toml
arferreira: `dashmap` stays unconditional for now since cache always needs it, and making cache optional is a separate issue.

lucaserm: done!

rapina/src/middleware/mod.rs
lucaserm: done!

rapina/src/lib.rs
```diff
@@ -65,7 +65,7 @@
 //! - [`BodyLimitMiddleware`](middleware::BodyLimitMiddleware) - Limit request body size
 //! - [`TraceIdMiddleware`](middleware::TraceIdMiddleware) - Add trace IDs to requests
 //! - [`RequestLogMiddleware`](middleware::RequestLogMiddleware) - Structured request logging
-//! - [`RateLimitMiddleware`](middleware::RateLimitMiddleware) - Token bucket rate limiting
+//! - [`RateLimitMiddleware`](middleware::RateLimitMiddleware) - Token bucket rate limiting (requires `rate-limit` feature)
```
arferreira: When built without `rate-limit`, `middleware::RateLimitMiddleware` doesn't exist and this intra-doc link resolves to nothing. `rustdoc` will warn about it. Adding "(requires `rate-limit` feature)" in prose doesn't fix the broken link, the link itself needs to go. Either remove the backtick-link and write it as inline code without the link target, or restructure the section.

rapina/Cargo.toml
```diff
@@ -54,7 +54,7 @@ schemars = { version = "1.2.0", features = ["chrono04", "uuid1", "rust_decimal1"
 # JWT Authentication
 jsonwebtoken = { version = "10.3.0", features = ["rust_crypto"] }
 
-# Rate Limiting
+# Concurrent map (used by cache + optional rate limiting)
```
arferreira: This is inaccurate, it's also used by relay/backend and relay/channel!

rapina/src/middleware/mod.rs
```diff
@@ -23,6 +24,7 @@ pub use body_limit::BodyLimitMiddleware;
 #[cfg(feature = "compression")]
 pub use compression::{CompressionConfig, CompressionMiddleware};
 pub use cors::{AllowedHeaders, AllowedMethods, AllowedOrigins, CorsConfig, CorsMiddleware};
+#[cfg(feature = "rate-limit")]
```
arferreira, line 27: It's unfair.. but we need to add docs here, not only here but in all the gated features. Would you mind mapping out and opening issues for a follow-up?
Something like:

```rust
#[cfg(feature = "rate-limit")]
#[cfg_attr(docsrs, doc(cfg(feature = "rate-limit")))]
pub use rate_limit::{KeyExtractor, RateLimitConfig, RateLimitMiddleware};
```

PR #369, opened 2026-03-16
feat(cli): add `import database --diff` for schema drift detection

### LINKED ISSUE #245: Add --diff mode to compare entities against live database
Follow-up from #170 / #240.

`rapina import database --diff` should compare current entity definitions against the live database and report drift — columns added/removed/changed that aren't reflected in code. Useful for catching manual DB changes.

lucaserm:
## Summary

Compare entity definitions in src/entity.rs against a live database and report drift — columns added/removed/changed, type mismatches, nullability differences, untracked tables, and missing tables. Returns non-zero exit code when drift is detected for CI use.

- Add entity_parser module to parse schema! blocks from entity files
- Add diff engine (compute_drift) comparing parsed entities vs DB
- Add --diff flag (conflicts_with --force) to import database command
- Update CLI docs with --diff flag and usage examples
- 49 new tests covering parser, diff engine, and edge cases

## Related Issues

Closes #245

## Checklist

- [x] `cargo fmt` passes
- [x] Tests pass

### DIFF
--- a/rapina-cli/src/commands/entity_parser.rs
+++ b/rapina-cli/src/commands/entity_parser.rs
@@ -28,7 +28,6 @@ pub struct ParsedEntity {
 
 /// A field extracted from an entity definition.
 #[derive(Debug, Clone)]
-#[allow(dead_code)]
 pub struct ParsedField {
     /// Field name as written in schema!
     pub name: String,
@@ -81,6 +80,9 @@ pub fn parse_entity_source(content: &str) -> Result<Vec<ParsedEntity>, String> {
 }
 
 /// Extract the text content of all `schema! { ... }` blocks.
+///
+/// NOTE: brace matching is done on raw text. Unbalanced braces inside
+/// inline comments or string literals will produce wrong results.
 fn extract_schema_blocks(content: &str) -> Vec<String> {
     let mut blocks = Vec::new();
     let mut lines = content.lines().peekable();
--- a/rapina-cli/src/commands/import.rs
+++ b/rapina-cli/src/commands/import.rs
@@ -1151,9 +1151,17 @@ pub fn database_diff(
 
     let tables = introspect_tables(url, schema_name)?;
 
+    if entities.is_empty() {
+        return Err("No entities found in src/entity.rs — nothing to diff".to_string());
+    }
+
     let total = tables.len();
     let db_tables = filter_tables_for_diff(tables, table_filter);
     println!("  {} Discovered {} table(s)", "✓".green(), total);
+
+    if db_tables.is_empty() {
+        return Err("No tables found in database — nothing to diff".to_string());
+    }
 
     let report = compute_drift(&entities, &db_tables);
     print_drift_report(&report);
@@ -2071,6 +2079,40 @@ mod tests {
             missing_tables: vec![],
         };
         assert!(!report.has_drift());
+    }
+
+    /// Empty entities + real tables → every table appears untracked (false positive).
+    /// `database_diff` guards against this with an early return.
+    #[test]
+    fn test_drift_empty_entities_reports_all_untracked() {
+        let entities: Vec<ParsedEntity> = vec![];
+        let tables = vec![make_table(
+            "users",
+            vec![
+                ("id", NormalizedType::I32, false),
+                ("email", NormalizedType::Str, false),
+            ],
+        )];
+
+        let report = compute_drift(&entities, &tables);
+        assert!(report.has_drift());
+        assert_eq!(report.untracked_tables, vec!["users"]);
+    }
+
+    /// Real entities + empty tables → every entity appears missing (false positive).
+    /// `database_diff` guards against this with an early return.
+    #[test]
+    fn test_drift_empty_tables_reports_all_missing() {
+        let entities = vec![make_entity(
+            "User",
+            "users",
+            vec![("email", "String", false)],
+        )];
+        let tables: Vec<IntrospectedTable> = vec![];
+
+        let report = compute_drift(&entities, &tables);
+        assert!(report.has_drift());
+        assert_eq!(report.missing_tables, vec!["users"]);
     }
 
     #[test]
--- a/rapina-cli/src/commands/mod.rs
+++ b/rapina-cli/src/commands/mod.rs
@@ -4,6 +4,7 @@ pub mod add;
 pub(crate) mod codegen;
 pub mod dev;
 pub mod doctor;
+#[cfg(feature = "import")]
 pub(crate) mod entity_parser;
 #[cfg(feature = "import")]
 pub mod import;


rapina-cli/src/commands/entity_parser.rs
```diff
@@ -0,0 +1,964 @@
...
+        if c.is_uppercase() {
+            if i > 0 {
+                result.push('_');
+            }
+            result.push(c.to_lowercase().next().unwrap());
+        } else {
+            result.push(c);
+        }
+    }
+    result
```
arferreira, line 453: This inserts an underscore before every uppercase char, so `HTMLParser` becomes `h_t_m_l_parser` and `APIToken` becomes `a_p_i_token`. The `schema!` macro uses `heck` for snake_case conversion internally, so the parser and compiler will disagree on table names for any entity with consecutive uppercase letters. That means false drift on every one of those entities.

Needs to handle the consecutive uppercase case. Something like:

```suggestion
fn to_snake_case(s: &str) -> String {
    let mut result = String::new();
    let chars: Vec<char> = s.chars().collect();
    for (i, &c) in chars.iter().enumerate() {
        if c.is_uppercase() {
            let next_is_lower = chars.get(i + 1).is_some_and(|n| n.is_lowercase());
            if i > 0 && (next_is_lower || !chars[i - 1].is_uppercase()) {
                result.push('_');
            }
            result.push(c.to_lowercase().next().unwrap());
        } else {
            result.push(c);
        }
    }
    result
}
```

And update the test below to assert `html_parser` instead of `h_t_m_l_parser`.

rapina-cli/src/commands/import.rs
```diff
@@ -699,6 +699,477 @@ pub fn database(
...
+        // has_many fields don't correspond to DB columns
+        if field.is_has_many {
+            continue;
+        }
+
+        if field.is_belongs_to {
+            // belongs_to generates a {name}_id FK column, assumed i32
+            columns.push(ExpectedColumn {
+                name: field.column_name.clone(),
+                expected_type: NormalizedType::I32,
```
arferreira: This hardcodes i32 for all FK columns but if the referenced entity uses a UUID primary key the actual DB column will be UUID. That produces a false type mismatch on every belongs_to pointing at a UUID entity.

At minimum document the limitation here with a comment. Ideally you could look up the referenced entity's PK type from the parsed entities list since you already have them in scope when calling compute_drift.

rapina-cli/src/commands/import.rs
```diff
@@ -699,6 +699,477 @@ pub fn database(
...
+    table_filter: Option<&[String]>,
+    schema_name: Option<&str>,
+) -> Result<(), String> {
+    codegen::verify_rapina_project()?;
+
+    println!();
+    println!("  {} Parsing entity definitions...", "→".bright_cyan());
+
+    let entities =
+        super::entity_parser::parse_entity_file_at(std::path::Path::new("src/entity.rs"))?;
```
arferreira, line 1142: `import database` generates entities into feature module directories like `src/users/entity.rs`, not a single `src/entity.rs`. So the diff tool cannot find entities created by rapina's own import command.

Should either glob for `**/entity.rs` files or accept a path flag like `--entity-file` so users can point it at the right location.

rapina-cli/src/commands/import.rs
```diff
@@ -699,6 +699,477 @@ pub fn database(
...
+        entities.len(),
+        "src/entity.rs".cyan()
+    );
+
+    println!("  {} Connecting to database...", "→".bright_cyan());
+
+    let rt = tokio::runtime::Runtime::new()
+        .map_err(|e| format!("Failed to create async runtime: {}", e))?;
+
+    let tables: Vec<IntrospectedTable> = rt.block_on(async {
```
arferreira: This entire block (runtime creation, URL scheme dispatch, introspect calls, feature gate error messages) is duplicated from `database()` above. Extract a shared helper like `fn introspect_tables(url: &str, schema_name: Option<&str>) -> Result<Vec<IntrospectedTable>>` so both code paths stay in sync when you add a new DB backend or change error messages.

rapina-cli/src/commands/import.rs
```diff
@@ -598,7 +598,7 @@ pub fn database(
     let rt = tokio::runtime::Runtime::new()
         .map_err(|e| format!("Failed to create async runtime: {}", e))?;
 
-    let tables = rt.block_on(async {
+    let tables: Vec<IntrospectedTable> = rt.block_on(async {
```
arferreira: This type annotation was already inferred by the compiler. No reason to touch this line in this PR.

rapina-cli/src/commands/entity_parser.rs
lucaserm: done!

rapina-cli/src/commands/import.rs
lucaserm: done!

rapina-cli/src/commands/import.rs
lucaserm: done!

rapina-cli/src/commands/import.rs
lucaserm: done!

rapina-cli/src/commands/import.rs
lucaserm: But I have a question, import database writes all entities into a single flat file src/entity.rs (codegen.rs:597). Every entity gets appended as a schema! {} block in that one file, right?

import database --diff reads from the same src/entity.rs (import.rs:1093).

import database does generate feature modules for other things — handlers, DTOs, routes go into directories like src/users/ — but the entity definitions all live in src/entity.rs.

It should not turns into a new issue?

rapina-cli/src/commands/mod.rs
```diff
@@ -4,6 +4,7 @@ pub mod add;
 pub(crate) mod codegen;
 pub mod dev;
 pub mod doctor;
+pub(crate) mod entity_parser;
```
arferreira, line 8: This is only used by import feature, which IS gated. So when `import` is disabled, this is gonna be compiled as dead code.

rapina-cli/src/commands/entity_parser.rs
```diff
@@ -0,0 +1,969 @@
...
+    /// Whether created_at is expected
+    pub has_created_at: bool,
+    /// Whether updated_at is expected
+    pub has_updated_at: bool,
+    /// Primary key columns (None = default auto i32 "id")
+    pub primary_key: Option<Vec<String>>,
+}
+
+/// A field extracted from an entity definition.
+#[derive(Debug, Clone)]
```
arferreira, line 30: The way you put this, I mean the blanket on the struct silences everything, this should be in the exact field where needs to sit allow dead code.

rapina-cli/src/commands/entity_parser.rs
```diff
@@ -0,0 +1,969 @@
...
+    let mut entities = Vec::new();
+
+    for block in blocks {
+        entities.extend(parse_schema_block(&block)?);
+    }
+
+    Ok(entities)
+}
+
+/// Extract the text content of all `schema! { ... }` blocks.
```
arferreira, line 82:
```suggestion
/// Extract the text content of all `schema! { ... }` blocks.
///
/// NOTE: brace matching is done on raw text. Unbalanced braces inside
/// inline comments or string literals will produce wrong results.
```

rapina-cli/src/commands/import.rs
```diff
@@ -699,6 +714,457 @@ pub fn database(
...
+    }
+
+    if !report.missing_tables.is_empty() {
+        println!("  {} Missing tables (in entity, not in DB):", "⚠".yellow());
+        for t in &report.missing_tables {
+            println!("    {} {}", "•".yellow(), t);
+        }
+        println!();
+    }
+
```
ShiraiEd: warnings report

```rust
if !report.warnings.is_empty() {
    println!("  {} Warnings:", "⚠".yellow());
    for w in &report.warnings {
        println!("    {} {}", "•".yellow(), w);
    }
    println!();
}
```

rapina-cli/src/commands/import.rs
```diff
@@ -699,6 +714,457 @@ pub fn database(
...
+    table_filter: Option<&[String]>,
+    schema_name: Option<&str>,
+) -> Result<(), String> {
+    codegen::verify_rapina_project()?;
+
+    println!();
+    println!("  {} Parsing entity definitions...", "→".bright_cyan());
+
+    let entities =
+        super::entity_parser::parse_entity_file_at(std::path::Path::new("src/entity.rs"))?;
```
ShiraiEd, line 1142: check if there is nothing in the entities for a clearer report
```rust
 if entities.is_empty() {
        return Err("No entities found in src/entity.rs — nothing to diff".to_string());
    }
```

rapina-cli/src/commands/import.rs
```diff
@@ -699,6 +714,457 @@ pub fn database(
...
+        entities.len(),
+        "src/entity.rs".cyan()
+    );
+
+    println!("  {} Connecting to database...", "→".bright_cyan());
+
+    let tables = introspect_tables(url, schema_name)?;
+
+    let total = tables.len();
+    let db_tables = filter_tables_for_diff(tables, table_filter);
```
ShiraiEd, line 1159:
```rust
  if db_tables.is_empty() {
        return Err("No tables found in database — nothing to diff".to_string());
    }
```

PR #373, opened 2026-03-18
feat: unify port env var to RAPINA_PORT and add .env scanning

### LINKED ISSUE #352: `rapina openapi export` should identify which port is being used, either via a --port flag or by scanning the current project
## Description

Currently, when exporting OpenAPI with Rapina, the tool does not identify which port the application is running on, running always in `3000`. This can lead to incorrect server URLs in the generated specification.

It would be helpful if the export process could determine the port automatically or allow the user to specify it manually.

## Suggested Improvement

Add a way for the OpenAPI export to determine the port used by the project. This could work in two ways:

- Allow specifying a port via a CLI flag (e.g., --port 3000)
- Attempt to detect the port by scanning the current project configuration

> This would ensure the generated OpenAPI specification contains the correct server URL and would improve the developer experience by avoiding manual adjustments after exporting.

### Context

_No response_

### Scope

Tooling (CLI, macros)

### Examples

```rust
rapina openapi export -o openapi.json --port 3000
```

lucaserm:
## Summary

Unify OpenAPI commands (export, check, diff) to use RAPINA_PORT instead of SERVER_PORT, matching dev/routes/doctor. SERVER_PORT and PORT are preserved as backwards-compatible fallbacks.

Load .env file before clap parsing so all port-using commands auto-detect the port from project configuration. Make dotenvy a non-optional dependency to support this.

## Related Issues

Closes #352

## Checklist

- [x] `cargo fmt` passes
- [x] Tests pass


docs/content/docs/cli/commands.md
```diff
@@ -287,7 +287,9 @@ Options:
 
 | Flag | Description | Default |
 |------|-------------|---------|
-| `-o, --output <FILE>` | Output file | openapi.json |
+| `-o, --output <FILE>` | Output file | stdout |
```
lucaserm, line 306: If I specify `rapina openapi export -o openapi.json` or rapina openapi export --output openapi.json` it will output into openapi.json not in stdout, so it be better if there was any example of that

docs/content/docs/core-concepts/openapi.md
```diff
@@ -184,7 +184,7 @@ Internal routes under `/__rapina/` are excluded from the generated spec, so the
 
 ## CLI Tools
 
-The `rapina` CLI ships three subcommands for working with OpenAPI specs. All three require a running development server and accept `--host` (default `127.0.0.1`) and `--port` / `-p` (default `3000`, also reads `$SERVER_PORT`).
+The `rapina` CLI ships three subcommands for working with OpenAPI specs. All three require a running development server and accept `--host` (default `127.0.0.1`) and `--port` / `-p` (default `3000`, also reads `$RAPINA_PORT` or `$SERVER_PORT`).
```
lucaserm, line 187: Why  $RAPINA_PORT **and** $SERVER_PORT? Some pattern could be useful.

rapina-cli/src/common/env.rs
```diff
@@ -0,0 +1,103 @@
...
+///   3. `PORT` set → copy to `RAPINA_PORT` (generic convention)
+///
+/// # Safety
+/// Must be called before any threads are spawned (i.e. at the top of `main`).
+pub fn propagate_port_env() {
+    if std::env::var("RAPINA_PORT").is_ok() {
+        return;
+    }
+
+    if let Ok(port) = std::env::var("SERVER_PORT") {
```
lucaserm, line 16: This should be only RAPINA_PORT or we can use this approach too @arferreira?

lucaserm (commented):
Leave a few nits

docs/content/docs/cli/commands.md
arferreira: Yes, worth adding. The default changed from `openapi.json` to stdout which is a behavior change, so showing the explicit file output case removes any ambiguity. Something like `rapina openapi export -o openapi.json` right below the options table is enough.

docs/content/docs/core-concepts/openapi.md
arferreira: The docs mention both because both actually work. `RAPINA_PORT` is the canonical var going forward, `SERVER_PORT` is preserved so existing projects don't break silently when they upgrade. Documenting the fallback tells users what to expect rather than hiding it. If you want to make the priority order explicit you could write "reads `$RAPINA_PORT`, falling back to `$SERVER_PORT`".

rapina-cli/src/common/env.rs
arferreira: Keep the fallback chain. The whole point of `propagate_port_env` is backwards compatibility. If you drop `SERVER_PORT` support here, existing users who already have that set in their `.env` or shell will get silent breakage on upgrade with no obvious reason why the port changed back to 3000. Same for `PORT`: platforms like Railway and Render inject it automatically, so picking it up for free is a net win. `RAPINA_PORT` wins when set, everything else is just a migration path.

docs/content/docs/cli/commands.md
lucaserm: done!

docs/content/docs/core-concepts/openapi.md
lucaserm: done!

rapina-cli/src/common/env.rs
lucaserm: done!

PR #378, opened 2026-03-18
Streaming body support

### LINKED ISSUE #326: Support streaming and SSE responses
`BoxBody` is `Full<Bytes>`, which means every response must be fully materialized in memory before hyper sends a single byte. This rules out SSE, LLM token streaming, large file downloads, and any chunked transfer pattern. In 2026 this is a hard limitation.

The lock is at the type alias in `response.rs`. `Full<Bytes>` flows through `IntoResponse`, the `Handler` trait return type, `HandlerFn`, every middleware hop, and the `service_fn` closure. But hyper itself doesn't care — `serve_connection` only requires the `http_body::Body` trait, which `Full<Bytes>` happens to implement but so does `StreamBody`, channel-based bodies, and anything else that yields frames incrementally.

The fix is replacing `BoxBody` with an actual boxed trait object (`http_body_util::combinators::BoxBody<Bytes, Error>` or equivalent) so both `Full<Bytes>` and streaming body types can coexist behind the same type. `IntoResponse` stays the same for non-streaming handlers, they just wrap their `Full<Bytes>` into the boxed body. Streaming handlers return a `StreamBody` wrapping a `Stream<Item = Result<Frame<Bytes>, Error>>`.

On top of that, two concrete response types: `StreamResponse` for raw chunked byte streams, and `SseResponse` for Server-Sent Events with proper `text/event-stream` content type, keep-alive, and the `data:` / `event:` / `id:` framing. Both implement `IntoResponse` by producing a streaming body.

The middleware layer needs consideration too. The compression middleware currently calls `body.collect().await` to buffer the full response, which defeats streaming. For streaming responses, compression should either be skipped or applied per-chunk.

lucaserm:
## Summary

- Replaced BoxBody with a custom struct wrapping Pin<Box<dyn http_body::Body + Send>>. This is a true trait object so both Full<Bytes> (buffered) and StreamBody (streaming) coexist behind the same type. Added full_body() and empty_body() helpers, then mechanically updated ~20 files that constructed Full::new() directly.
- Added StreamResponse and SseResponse — two new response types implementing IntoResponse. StreamResponse wraps any Stream<Item = Result<Bytes, E>> for chunked transfers. SseResponse produces text/event-stream with proper data:/event:/id: framing and optional keep-alive. Both insert a StreamingMarker into response extensions.
- Made middleware streaming-aware — compression and cache middlewares check for StreamingMarker and skip body buffering. Non-streaming responses continue to be compressed/cached normally.
- Docs — new "Streaming Responses" page covering SSE, chunked streams, JSON events, keep-alive, and custom streaming types. Added notes to middleware and caching docs.

Key design decision: Couldn't use http_body_util::combinators::BoxBody because it requires Send + Sync, but StreamBody<Pin<Box<dyn Stream + Send>>> is only Send. So I defined a custom BoxBody struct that only requires Send. The name stays the same, so the proc-macro and all downstream code works unchanged.

## Related Issues

Closes #326

## Checklist

- [x] `cargo fmt` passes
- [x] Tests pass

### DIFF
--- a/rapina/src/streaming.rs
+++ b/rapina/src/streaming.rs
@@ -167,18 +167,16 @@ impl IntoResponse for SseResponse {
                 let merged = futures_util::stream::unfold(
                     (event_stream.fuse(), interval),
                     |(mut events, interval)| async move {
-                        loop {
-                            let sleep = tokio::time::sleep(interval);
-                            tokio::pin!(sleep);
-                            tokio::select! {
-                                biased;
-                                item = futures_util::StreamExt::next(&mut events) => {
-                                    return item.map(|result| (result, (events, interval)));
-                                }
-                                _ = &mut sleep => {
-                                    let comment = Bytes::from_static(b": keep-alive\n\n");
-                                    return Some((Ok(Frame::data(comment)), (events, interval)));
-                                }
+                        let sleep = tokio::time::sleep(interval);
+                        tokio::pin!(sleep);
+                        tokio::select! {
+                            biased;
+                            item = futures_util::StreamExt::next(&mut events) => {
+                                item.map(|result| (result, (events, interval)))
+                            }
+                            _ = &mut sleep => {
+                                let comment = Bytes::from_static(b": keep-alive\n\n");
+                                Some((Ok(Frame::data(comment)), (events, interval)))
                             }
                         }
                     },
--- a/rapina/src/health/endpoint.rs
+++ b/rapina/src/health/endpoint.rs
@@ -36,9 +36,9 @@ pub async fn liveness_check(
     Response::builder()
         .status(StatusCode::OK)
         .header(CONTENT_TYPE, APPLICATION_JSON)
-        .body(crate::response::full_body(http_body_util::Full::new(bytes::Bytes::from(
-            serde_json::to_vec(&body).unwrap_or_default(),
-        ))))
+        .body(crate::response::full_body(http_body_util::Full::new(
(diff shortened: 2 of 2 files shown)


docs/content/docs/core-concepts/streaming.md
```diff
@@ -0,0 +1,180 @@
...
+            })
+        );
+
+        use http_body_util::BodyExt;
+        let body = BoxBody::new(body.map_err(|e| -> BoxBodyError { e }));
+
+        let mut response = http::Response::builder()
+            .status(http::StatusCode::OK)
+            .body(body)
+            .unwrap();
```
richasdev: I think using .expect("message") is better because the build will fail

rapina/src/streaming.rs
```diff
@@ -0,0 +1,416 @@
...
+        let stream: Pin<Box<dyn Stream<Item = Result<Frame<Bytes>, BoxBodyError>> + Send>> =
+            if let Some(interval) = self.keep_alive {
+                let keep_alive_stream = futures_util::stream::unfold((), move |()| async move {
+                    tokio::time::sleep(interval).await;
+                    let comment = Bytes::from_static(b": keep-alive\n\n");
+                    Some((Ok(Frame::data(comment)), ()))
+                });
+                Box::pin(futures_util::stream::select(
+                    event_stream,
+                    keep_alive_stream,
```
richasdev: `stream::select` with an infinite `keep_alive_stream` means the SSE connection never closes after the event stream is exhausted,`select` only terminates when both sides end

PR #379, opened 2026-03-18
feat: generate BelongsTo/HasMany relationship fields from foreign keys during import

### LINKED ISSUE #244: Generate relationship fields in schema! block from foreign keys
Follow-up from #170 / #240.

The import command detects foreign keys but currently keeps them as plain integer fields. It should generate BelongsTo/HasMany relationship fields in the schema! block when both sides of the FK are being imported.

lucaserm:
## Summary

The import command now converts FK columns into proper relationship fields in the schema! block when both sides of the FK are being imported. All imported tables are placed in a single combined schema! block.

- FK columns (e.g., user_id) replaced with BelongsTo fields (e.g., user: User)
- Nullable FKs become Option<Entity> (optional BelongsTo)
- Referenced tables get HasMany fields (e.g., posts: Vec<Post>)
- Migrations still use raw scalar columns as before

## Related Issues

Closes #244

## Checklist

- [x] `cargo fmt` passes
- [x] Tests pass

### DIFF
--- a/rapina-cli/src/commands/import.rs
+++ b/rapina-cli/src/commands/import.rs
@@ -493,21 +493,10 @@ fn detect_timestamps(table: &IntrospectedTable) -> Option<&'static str> {
 // Per-table generation
 // ---------------------------------------------------------------------------
 
-fn generate_for_table(
-    table: &IntrospectedTable,
-    _relationships: &HashMap<String, Vec<RelationshipInfo>>,
-    force: bool,
-) -> Result<(), String> {
-    let singular = codegen::singularize(&table.name);
-    let plural = &table.name;
-    let pascal = codegen::to_pascal_case(&singular);
-    let pascal_plural = codegen::to_pascal_case(plural);
-
+/// Build all scalar fields for a table (including FK columns), for use by
+/// migration and feature-module generation which need the raw database columns.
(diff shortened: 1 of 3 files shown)


rapina-cli/src/commands/import.rs
```diff
@@ -520,14 +558,28 @@ fn generate_for_table(
...
     };
 
+    // Collect FK column names so we can skip them as scalar fields
+    let fk_columns: std::collections::HashSet<String> = relationships
+        .get(&table.name)
+        .map(|rels| {
+            rels.iter()
+                .filter(|r| matches!(r.kind, RelationKind::BelongsTo))
+                .map(|r| format!("{}_id", r.field_name))
+                .collect()
```
arferreira, line 568: This reconstruction only holds when the original FK column ended in `_id`. `resolve_relationships` does `strip_suffix("_id").unwrap_or(fk_column)`, so for a column named `owner` the `field_name` is `"owner"` and this produces `"owner_id"` — a name that does not exist in the table. The column never gets excluded from scalar fields, and the entity ends up with both the raw field and the relation. Storing the original FK column name in `RelationshipInfo` and using it here directly avoids the reconstruction entirely.

PR #380, opened 2026-03-18
perf: tfb hot path

### LINKED ISSUE #322: TechEmpower Framework Benchmarks submission
Both rapina and axum sit on hyper, so the only thing TFB measures is framework overhead. After the radix trie and SmallVec PathParams work, routing is no longer the bottleneck, but the rest of the request pipeline still has unnecessary allocations that would show up under TFB's plaintext and JSON tests.

The main offender is `RequestContext::new()` in `server.rs`, which generates a UUID v4 and allocates a 36-byte String on every single request, unconditionally, before middleware even runs. There's no way to opt out. For a TFB plaintext handler that just returns `Hello, World!`, this dominates the profile. The fix is making context creation lazy, only pay for it when tracing middleware is actually registered.

Second issue: `&'static str` responses go through `.to_owned()` before becoming `Bytes`. The bytes crate already has `Bytes::from_static()` for exactly this case, zero copy, zero alloc. We should detect static lifetimes in `IntoResponse` and take that path.

Third: TFB requires a `Date` response header. Every competitive entry caches this value and updates it once per second from a background task instead of formatting it per request.

The plan is to start with plaintext and JSON (no database), fix the hot path issues above, fork TechEmpower/FrameworkBenchmarks, add `frameworks/Rust/rapina/` with the required config and Dockerfile, validate locally with `./tfb --test rapina`, and PR upstream. Database tests come as a follow up.

lucaserm:
## Summary

Three optimizations to rapina's request hot path for TechEmpower benchmarks:

- Lazy trace_id — UUID v4 is no longer generated on every request. It uses OnceLock and only allocates when tracing middleware actually reads it. Zero cost when no tracing is registered.
- StaticStr + String fix — Removed a redundant .to_owned() on String responses. Added a StaticStr newtype that uses Bytes::from_static() for zero-copy, zero-alloc &'static str responses.
- Cached Date header — A background task formats the HTTP Date header once per second and shares it via a watch channel. Every response gets the header injected automatically without per-request formatting.

## Related Issues

Closes #322

## Checklist

- [x] `cargo fmt` passes
- [x] Tests pass

### DIFF
--- a/rapina/src/context.rs
+++ b/rapina/src/context.rs
@@ -1,4 +1,4 @@
-use std::sync::OnceLock;
+use std::sync::{Arc, OnceLock};
 use std::time::Instant;
 
 /// Per-request context passed through the middleware stack and into handlers.
@@ -11,7 +11,7 @@ use std::time::Instant;
 /// the UUID v4 allocation entirely.
 #[derive(Debug, Clone)]
 pub struct RequestContext {
-    trace_id: OnceLock<String>,
+    trace_id: Arc<OnceLock<String>>,
     /// Timestamp recorded when the request context was created, used to
     /// calculate request duration.
     pub start_time: Instant,
@@ -20,13 +20,13 @@ pub struct RequestContext {
 impl RequestContext {
     pub fn new() -> Self {
         Self {
-            trace_id: OnceLock::new(),
+            trace_id: Arc::new(OnceLock::new()),
             start_time: Instant::now(),
         }
     }
 
     pub fn with_trace_id(trace_id: String) -> Self {
-        let cell = OnceLock::new();
+        let cell = Arc::new(OnceLock::new());
         let _ = cell.set(trace_id);
         Self {
             trace_id: cell,
@@ -95,11 +95,11 @@ mod tests {
     }
 
     #[test]
-    fn test_clone() {
+    fn test_clone_shares_trace_id() {
         let ctx1 = RequestContext::new();
-        // Access trace_id to initialize it before cloning
-        let _ = ctx1.trace_id();
         let ctx2 = ctx1.clone();
+        // Arc<OnceLock> ensures both sides share the same cell,
+        // so whichever initializes first wins.
         assert_eq!(ctx1.trace_id(), ctx2.trace_id());
     }
 
--- a/rapina/src/date_cache.rs
+++ b/rapina/src/date_cache.rs
@@ -17,7 +17,7 @@ pub(crate) struct DateHeaderCache {
 fn format_date_header() -> HeaderValue {
     let now = SystemTime::now();
     let formatted = httpdate::HttpDate::from(now).to_string();
-    // SAFETY: httpdate always produces valid ASCII
+    // httpdate always produces valid ASCII, so this never panics
     HeaderValue::from_str(&formatted).unwrap()
 }
 


lucaserm:
Needs to submit to TechEmpower #322. Planning to send a main.rs like:

```rust
use rapina::prelude::*;

use mimalloc::MiMalloc;

#[global_allocator]
static GLOBAL: MiMalloc = MiMalloc;

#[derive(Serialize)]
struct Message {
    message: &'static str,
}

#[get("/plaintext")]
async fn plaintext() -> StaticStr {
    StaticStr("Hello, World!")
}

#[get("/json")]
async fn json() -> Json<Message> {
    Json(Message {
        message: "Hello, World!",
    })
}

#[tokio::main]
    Rapina::new()
        .discover()
        .listen("0.0.0.0:8080")
        .await
}
```

All required files already has been written — just waiting merge.

rapina/src/context.rs
```diff
@@ -1,15 +1,17 @@
+use std::sync::OnceLock;
```
arferreira:
```suggestion
use std::sync::{Arc, OnceLock};
```

rapina/src/context.rs
```diff
@@ -1,15 +1,17 @@
...
+/// The trace ID is generated lazily on first access, so requests that never
+/// read the trace ID (e.g. when no tracing middleware is registered) avoid
+/// the UUID v4 allocation entirely.
 #[derive(Debug, Clone)]
 pub struct RequestContext {
-    /// Unique identifier for this request, used for distributed tracing.
-    /// Populated from the incoming `x-trace-id` header if present, otherwise
-    /// a new UUID v4 is generated.
-    pub trace_id: String,
+    trace_id: OnceLock<String>,
```
arferreira: `OnceLock<String>` clones as an empty cell. In `server.rs` line 81, `ctx.clone()` is inserted into request extensions before trace_id is ever accessed. When middleware later calls `trace_id()` on the cloned context from extensions, and a handler calls `trace_id()` on the original `ctx`, both generate independent UUIDs. They are supposed to be the same request.

`Arc<OnceLock<String>>` fixes this: original and clone share one cell, so whichever side initializes first wins and the other side reads the same value.

```suggestion
    trace_id: Arc<OnceLock<String>>,
```

rapina/src/context.rs
```diff
@@ -18,18 +20,26 @@ pub struct RequestContext {
 impl RequestContext {
     pub fn new() -> Self {
         Self {
-            trace_id: uuid::Uuid::new_v4().to_string(),
+            trace_id: OnceLock::new(),
```
arferreira:
```suggestion
            trace_id: Arc::new(OnceLock::new()),
```

rapina/src/context.rs
```diff
@@ -18,18 +20,26 @@ pub struct RequestContext {
...
             start_time: Instant::now(),
         }
     }
 
     pub fn with_trace_id(trace_id: String) -> Self {
+        let cell = OnceLock::new();
```
arferreira:
```suggestion
        let cell = Arc::new(OnceLock::new());
```

rapina/src/date_cache.rs
```diff
@@ -0,0 +1,67 @@
...
+/// backed by reference-counted `Bytes`).
+#[derive(Clone)]
+pub(crate) struct DateHeaderCache {
+    rx: watch::Receiver<HeaderValue>,
+}
+
+fn format_date_header() -> HeaderValue {
+    let now = SystemTime::now();
+    let formatted = httpdate::HttpDate::from(now).to_string();
+    // SAFETY: httpdate always produces valid ASCII
```
arferreira: `// SAFETY:` is the convention for `unsafe` blocks. This is safe code calling `unwrap()`. Just use a regular comment.

```suggestion
    // httpdate always produces valid ASCII, so this never panics
```

PR #394, opened 2026-03-20
Add `rapina jobs list` CLI command

### LINKED ISSUE #340: Background jobs: rapina jobs list CLI command
Basic CLI visibility into the job queue. Queries `rapina_jobs` and displays counts by status (pending, running, completed, failed) and optionally lists individual jobs with `--failed` to show error details.

Follow the existing CLI command structure in `rapina-cli/src/commands/` — add a `jobs.rs` module with the subcommand. The command needs a database connection, same as `rapina import` uses. Output format should match the style of `rapina routes` — clean, colored, tabular.

`rapina jobs retry <id>` and `rapina jobs purge` come in Phase 2, not here. This is just read-only visibility.

See [RFC: Background Jobs (discussion #334)](https://github.com/rapina-rs/rapina/discussions/334), "CLI" section.

Depends on #335.

uemuradevexe:
## Summary

add `rapina jobs list` CLI command for read-only visibility into the background job queue                                               
                                                                                                                                                
  - `rapina jobs list` queries `rapina_jobs` and displays counts by status (`pending`, `running`, `completed`, `failed`), including any unexpected
   statuses                                                                                                                                     
  - `rapina jobs list --failed` additionally lists individual failed jobs with ID, queue, job type, attempt count, and truncated error message    
  - gated behind `jobs-postgres` feature flag following the same pattern as `seed-postgres`; without the feature, prints an actionable reinstall  
  hint 

## Related Issues

Closes #340 

## Checklist

- [x] `cargo fmt` passes
- [x] Tests pass

### DIFF
--- a/docs/content/docs/cli/commands.md
+++ b/docs/content/docs/cli/commands.md
@@ -289,6 +289,47 @@ The migration creates the `rapina_jobs` table used by the background jobs system
 
 > **Note:** The jobs migration requires PostgreSQL. It uses `gen_random_uuid()` and partial indexes, which are not available in MySQL or SQLite.
 
+## rapina jobs list
+
+Show job counts grouped by status:
+
+```bash
+rapina jobs list
+```
+
+Output:
+
+```
+  STATUS        COUNT
+  ────────────  ─────
+  pending       3
+  running       1
+  completed     42
+  failed        2
+
+  ✓ 48 total job(s)
+```
+
+Options:
+
+| Flag | Description |
+|------|-------------|
+| `--failed` | Also list individual failed jobs with error details |
+
+With `--failed`:
+
+```bash
+rapina jobs list --failed
+```
+
+This appends a table of failed jobs showing ID, queue, job type, attempt count (`attempts/max_retries`), and the last error message.
+
+Requires the `jobs` feature:
+
+```bash
+cargo install rapina-cli --features jobs-postgres
+```
+
 ## rapina openapi export
 
 Export the OpenAPI specification to a file:


rapina-cli/Cargo.toml
```diff
@@ -72,6 +72,8 @@ import-sqlite = [
     "sqlx/sqlite",
 ]
 import-openapi = ["dep:openapiv3", "dep:serde_yaml_ng"]
+jobs = ["dep:sea-orm", "dep:tokio", "dep:dotenvy"]
+jobs-postgres = ["jobs", "sea-orm/sqlx-postgres"]
```
arferreira: `dotenvy` is declared as a mandatory dependency on line 27 of this file (no `optional = true`). The `dep:` prefix only works for optional deps, so Cargo rejects this and refuses to parse the workspace. Just remove it from the list.

```suggestion
jobs = ["dep:sea-orm", "dep:tokio"]
```

rapina-cli/src/commands/jobs.rs
```diff
@@ -72,10 +75,224 @@ fn fresh_mod_rs() -> &'static str {
...
+        .map_err(|_| "DATABASE_URL environment variable is not set".to_string())?;
+
+    sea_orm::Database::connect(&database_url)
+        .await
+        .map_err(|e| format!("Failed to connect to database: {e}"))
+}
+
+/// Truncate a string to `max` visible characters, appending "..." if it exceeds
+/// the limit. Safe for multi-byte UTF-8 — never splits a codepoint.
+fn truncate_chars(s: &str, max: usize) -> String {
```
arferreira, line 236: `truncate_chars` is called only from `list()`, which is `#[cfg(feature = "jobs")]`, but the function itself has no gate. Without the feature enabled it becomes dead code and rustc will warn. If clippy is in the error gate this becomes a compile failure.

```suggestion
#[cfg(feature = "jobs")]
fn truncate_chars(s: &str, max: usize) -> String {
```

PR #396, opened 2026-03-20
Auto-generate OpenAPI requestBody from extractors

### LINKED ISSUE #395: [Feature]: Auto-generate OpenAPI requestBody from Json<T> and Validated<T> extractors
### Problem Statement

Currently, Rapina does an excellent job of automatically generating the response schema for OpenAPI documentation based on a handler's return type (e.g., -> Json<MyResponse>). However, it does not automatically document the incoming request payloads for POST, PUT, and PATCH endpoints.

When a handler expects a Json<CreateUser> or Validated<Json<CreateUser>>, the resulting /__rapina/openapi.json spec does not include a requestBody definition. This leaves the API documentation incomplete, meaning frontend developers and API consumers don't know what fields are required to make mutations unless they read the Rust source code.

### Proposed Solution

The routing macros (#[post], #[put], #[patch]) should inspect the function arguments at compile time. If an argument uses the Json<T>, Validated<Json<T>>, or Form<T> extractors, the macro should generate the JSON schema for the inner type T (which already implements schemars::JsonSchema via DTOs) and attach it to the route's metadata.

### Alternatives Considered

_No response_

Auditability (contracts, errors, observability)

### Contribution

sarpedondev:
## Summary

Auto-generate OpenAPI `requestBody` documentation for POST, PUT, and PATCH endpoints when handlers use `Json<T>`, `Validated<Json<T>>`, or `Form<T>` extractors. The routing macros now inspect function arguments at compile  
  time and generate JSON schemas for inner types that implement `JsonSchema`.

## Related Issues

Closes #395

## Checklist

- [x] `cargo fmt` passes
- [x] Tests pass

### DIFF
--- a/rapina-macros/src/lib.rs
+++ b/rapina-macros/src/lib.rs
@@ -214,25 +214,36 @@ fn route_macro_core(
         quote! {}
     };
 
-    // Extract request body type and content type for schema generation
-    let (request_schema_impl, request_content_type_impl) =
-        if let Some(meta) = extract_request_body_meta(&func.sig.inputs) {
-            let inner_type = meta.inner_type;
-            let content_type = meta.content_type;
-            (
-                quote! {
-                    fn request_schema() -> Option<serde_json::Value> {
-                        Some(rapina::openapi_schema_for::<#inner_type>())
-                    }
-                },
-                quote! {
-                    fn request_content_type() -> Option<&'static str> {
-                        Some(#content_type)
-                    }
-                },
-            )
+    // Extract request body type and content type for schema generation.
+    // Only generate requestBody for POST, PUT, and PATCH methods per OpenAPI spec.
+    let (request_schema_impl, request_content_type_impl, request_body_required_impl) =
+        if matches!(method, "POST" | "PUT" | "PATCH") {
+            if let Some(meta) = extract_request_body_meta(&func.sig.inputs) {
+                let inner_type = meta.inner_type;
+                let content_type = meta.content_type;
+                let required = meta.required;
+                (
+                    quote! {
+                        fn request_schema() -> Option<serde_json::Value> {
+                            Some(rapina::openapi_schema_for::<#inner_type>())
+                        }
+                    },
+                    quote! {
+                        fn request_content_type() -> Option<&'static str> {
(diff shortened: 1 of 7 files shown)


rapina-macros/src/lib.rs
```diff
@@ -214,6 +214,27 @@ fn route_macro_core(
         quote! {}
     };
 
+    // Extract request body type and content type for schema generation
+    let (request_schema_impl, request_content_type_impl) =
+        if let Some(meta) = extract_request_body_meta(&func.sig.inputs) {
```
arferreira: This runs for every HTTP method. A `#[delete(...)]` handler that takes `Json<T>` would produce a `requestBody` in the OpenAPI output, which the spec forbids for GET and strongly discourages for DELETE. `route_macro_core` already receives `method` as its first argument, would it make sense to gate this on `matches!(method, "POST" | "PUT" | "PATCH")`?

rapina/src/openapi/spec.rs
```diff
@@ -292,6 +292,26 @@ pub fn build_openapi_spec(
...
+            let mut content = BTreeMap::new();
+            content.insert(
+                content_type.to_string(),
+                MediaType {
+                    schema: Schema::Inline(schema.clone()),
+                },
+            );
+            operation.request_body = Some(RequestBody {
+                description: None,
+                required: true,
```
arferreira: `required: true` is hardcoded here. That's correct for every extractor this PR supports, but what's the plan once `Option<Json<T>>` style optional bodies come along, should the macro pass this information through, or would that be handled separately?

rapina-macros/src/lib.rs
sarpedondev: requestBody is now only generated for POST, PUT, and PATCH handlers. GET and DELETE handlers with body extractors won't produce OpenAPI requestBody definitions

rapina/src/openapi/spec.rs
sarpedondev: Option<Json<T>> and Option<Form<T>> are now supported and produce required: false in the OpenAPI output. The macro passes this information through the request_body_required field.

Issue #289, opened 2026-03-05
[Bug]: Rapina dev localhost static host/port


ShiraiEd:
### Description

If the `host:port(127.0.0.1:3000)` is changed in `Rapina::new().listen(here)`
the localhost address in Rapina Cli doesn't print the used one, instead it always prints the default 127.0.0.1:3000 address

### Steps to Reproduce

1. Create a new rapina project or use an existing one
2. Change the address inside `Rapina::new().listen()` in src/main.rs to something like 127.0.0.2:3000 or 127.0.0.1:8000
3. Run rapina dev inside the project's dir to create a local server
4. See the incorrect cli output 127.0.0.1:3000

### Expected Behavior

The address printed out should match the one in `Rapina::new().listen(here)`

### Code Example

```rust
Rapina::new()
        .state(ChatRoom { tx })
        .router(Router::new().route(Method::GET, "/", |_, _, _| async {
            rapina::http::Response::builder()
                .header("content-type", "text/html")
                .body(Full::new(Bytes::from(include_str!("index.html"))))
                .unwrap()
        }))
        .discover()
        .listen("127.0.0.1:8000") <- Here
        .await
}

//should return 
╭───────────────────────────────────────────────────────────╮
 │               Rapina CLI - Development Mode               │
 ├───────────────────────────────────────────────────────────┤
 │                                                           │
 │ Serving at: http://127.0.0.1:8000                          │
 │ Routes:     http://127.0.0.1:8000/.__rapina/routes        │
 │                                                           │
 │ For production: cargo build --release                     │
 │                                                           │
 ╰───────────────────────────────────────────────────────────╯
```

### Environment

- Rapina version:0.8.0
- Rust version:rustc 1.95.0-nightly
- OS:Omarchy

### Error Output / Logs

```shell
incorrect: INFO Server started on http://127.0.0.1:3000 (Press CTRL+C to quit)
correct: INFO Server started on http:// your input (Press CTRL+C to quit)
```

PR #406, opened 2026-03-23
Add host and port verification

ShiraiEd:
## Summary

The `print_banner` function in `dev.rs` always uses the CLI's `--host`/`--port` flags (or their defaults `127.0.0.1:3000`) for the banner output, never the address passed to `.listen()`, which is the actual address the server binds to as shown in the tracing logs.

The default case (`127.0.0.1:3000`) never surfaced this bug because the CLI defaults happen to match. It only appears when the developer changes `.listen()` to a different address.

If two different addresses are used, the program runs normally but the app doesn't work — the CLI tries to connect to its own address while the server is listening on a different one.

**What this PR does:**
- Adds a `check_listen_addr()` helper in `app.rs` that panics with a clear message if `RAPINA_HOST`/`RAPINA_PORT` (set by `rapina dev`) don't match the address passed to `.listen()`, failing fast instead of silently misbehaving
- Adds 4 serial tests covering: no env vars, matching addresses, mismatched port, mismatched host

**Root causes identified (not fully fixed yet):**
- `print_banner` is called before `build_and_run`, so it always shows the CLI defaults regardless of what `.listen()` uses
- There is no mechanism for the CLI to know what address `.listen()` will use before the server starts

**Open questions:**
An alternative approach was explored: piping the server's stdout/stderr and reading the actual bound address from the `"Rapina listening on"` tracing log line, then buffering output and printing the banner after with the correct address. This requires no extra steps from the developer but adds background threads for I/O forwarding.

Should `.listen()` be the source of truth and override `RAPINA_HOST`/`RAPINA_PORT`, or should the CLI flags be authoritative and `.listen()` serve as the production fallback only?
## Related Issues
#289 
## Checklist

- [x] `cargo fmt` passes
- [x] Tests pass

### DIFF
--- a/rapina/src/app.rs
+++ b/rapina/src/app.rs
@@ -85,6 +85,16 @@ pub struct Rapina {
     pub(crate) rfc7807_base_uri: String,
 }
 
+// Resolves the listen address, preferring RAPINA_HOST/RAPINA_PORT over addr when both are set.
+pub(crate) fn resolve_listen_addr(addr: &str) -> SocketAddr {
+    match (std::env::var("RAPINA_HOST"), std::env::var("RAPINA_PORT")) {
+        (Ok(host), Ok(port)) => format!("{host}:{port}")
+            .parse()
+            .expect("invalid RAPINA_HOST/RAPINA_PORT"),
+        _ => addr.parse().expect("invalid address"),
+    }
+}
+
 impl Rapina {
     /// Creates a new Rapina application builder.
     ///
@@ -661,9 +671,13 @@ impl Rapina {
     ///
     /// # Panics
     ///
-    /// Panics if the address cannot be parsed.
+    /// Panics if the address is invalid.
+    ///
+    /// When `RAPINA_HOST`/`RAPINA_PORT` env vars are set (e.g. under `rapina dev`),
+    /// they take priority over the `addr` argument so the CLI is always authoritative
+    /// in dev mode and the banner is correct by construction.
     pub async fn listen(self, addr: &str) -> std::io::Result<()> {
-        let addr: SocketAddr = addr.parse().expect("invalid address");
+        let addr: SocketAddr = resolve_listen_addr(addr);
         let app = self.prepare();
         serve(
             app.router,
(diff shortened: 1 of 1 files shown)


rapina/src/app.rs
```diff
@@ -659,11 +659,16 @@ impl Rapina {
 
     /// Starts the HTTP server on the given address.
     ///
-    /// # Panics
-    ///
-    /// Panics if the address cannot be parsed.
+    /// When `RAPINA_HOST`/`RAPINA_PORT` env vars are set (e.g. under `rapina dev`),
+    /// they take priority over the `addr` argument so the CLI is always authoritative
+    /// in dev mode and the banner is correct by construction.
     pub async fn listen(self, addr: &str) -> std::io::Result<()> {
```
arferreira, line 679: `listen()` still panics on an invalid address, worth keeping the `# Panics` note.

rapina/src/app.rs
```diff
@@ -702,6 +707,85 @@ mod tests {
...
     }
 
+    fn resolve_listen_addr(addr: &str) -> SocketAddr {
+        match (std::env::var("RAPINA_HOST"), std::env::var("RAPINA_PORT")) {
+            (Ok(host), Ok(port)) => format!("{host}:{port}")
+                .parse()
+                .expect("invalid RAPINA_HOST/RAPINA_PORT"),
+            _ => addr.parse().expect("invalid address"),
+        }
+    }
```
arferreira: Move this outside the test module so `listen()` can call it and the tests cover the real code path. Something like `pub(crate) fn resolve_listen_addr` just above the `impl Rapina` block.

rapina/src/app.rs
```diff
@@ -659,11 +659,16 @@ impl Rapina {
...
+    /// they take priority over the `addr` argument so the CLI is always authoritative
+    /// in dev mode and the banner is correct by construction.
     pub async fn listen(self, addr: &str) -> std::io::Result<()> {
-        let addr: SocketAddr = addr.parse().expect("invalid address");
+        let addr: SocketAddr = match (std::env::var("RAPINA_HOST"), std::env::var("RAPINA_PORT")) {
+            (Ok(host), Ok(port)) => format!("{host}:{port}")
+                .parse()
+                .expect("invalid RAPINA_HOST/RAPINA_PORT"),
+            _ => addr.parse().expect("invalid address"),
+        };
```
arferreira: Also, this duplicate `resolve_listen_addr` in the test module below. The tests call that local copy, not this, so they don't cover the actual production path.

rapina/src/app.rs
```diff
@@ -85,6 +85,16 @@ pub struct Rapina {
     pub(crate) rfc7807_base_uri: String,
 }
 
+// Verifies that the listen address is valid.
```
arferreira: Sorry, this is wrong, function doesn't just verify, it resolves which address to use based on env vars, right?
```suggestion
// Resolves the listen address, preferring RAPINA_HOST/RAPINA_PORT over addr when both are set.
```

PR #408, opened 2026-03-24
Add JWT validation based on Json Web Key Sets

### LINKED ISSUE #115: [Feature]: Verify JWT from JWKS url
### Problem Statement

When using machine-to-machine communication, incoming JWTs are typically signed by an Authorization server.
Rapina needs to verify that incoming JWTs are issued by the trusted Authorization server.

More information: https://auth0.com/docs/secure/tokens/json-web-tokens/json-web-key-sets 

### Proposed Solution

Rapina should ideally be configurable to load Json Web Key Sets (JWKS) from

1. Direct JWKS URL: Fetch keys directly from a specific endpoint 
2. OIDC Discovery endpoint: Fetch the configuration from the OIDC discovery endpoint 

Option 2. is similar to 1. but requires downloading the JSON provided in the OIDC discovery endpoint, finding the JWKS url and then fetching the JWKS data. 

The signature of incoming JWTs can be verified by using the public key from a matching key id (`kid`) in the JWKS data.

JWKS data should be held in a cache and must be refreshed somewhat frequently as keys can be rotated (configurable refresh?)

### Alternatives Considered

_No response_

Security (guardrails, validation)

### Contribution

juv:
## Summary

Adds Json Web Token extractor that validates the tokens based on a configured JWKS url or OIDC discovery endpoint

## Related Issues

Closes #115

## Checklist

- [ ] `cargo fmt` passes
- [ ] Tests pass

### DIFF
--- a/docs/content/docs/core-concepts/jwks.md
+++ b/docs/content/docs/core-concepts/jwks.md
@@ -61,7 +61,7 @@ This pulls in Rapina's `cron-scheduler` for automatic periodic cache refresh and
 
 ## JWKS Client
 
-`JwksClient` is responsible for fetching and caching the JSON Web Key Set from the identity provider. It is registered as application state and used automatically by the `JsonWebToken` extractor on every request.
+Rapina's `JwksClient` is responsible for fetching and caching the JSON Web Key Set from the identity provider. It is registered as application state and used automatically by the `JsonWebToken` extractor on every request.
 
 ### Caching and Automatic Refresh
 
@@ -88,6 +88,8 @@ let jwks_client = JwksClient::direct(
 );
 ```
 
+> ⚠️ **The JWKS endpoint url must contain the HTTPS scheme, i.e. start with `https://`**. The lack of transport-layer security can have a severe impact on the security of the Rapina backend and its protected resources. Rapina will reject urls with plain HTTP scheme during startup.
+
 ### OIDC Discovery
 
 Use this when the provider publishes an OpenID Connect discovery document. Rapina will first fetch the discovery document, extract the `jwks_uri` field, and then fetch the actual JWKS:
@@ -102,6 +104,8 @@ let jwks_client = JwksClient::oidc(
 ```
 
 This is the **recommended approach** for standard OIDC providers, as it is more robust: if the provider rotates their JWKS URL, the discovery document is updated automatically and your application continues to work.
+
+> ⚠️ **The OIDC discovery url must contain the HTTPS scheme, i.e. start with `https://`**. The lack of transport-layer security can have a severe impact on the security of the Rapina backend and its protected resources. Rapina will reject urls with plain HTTP scheme during startup.
 
 **OIDC discovery flow:**
 1. Fetch `discovery_url` → parse `jwks_uri`


rapina/src/jwt/extract.rs
```diff
@@ -0,0 +1,350 @@
...
+use serde::de::DeserializeOwned;
+use std::sync::Arc;
+
+#[derive(Debug, Deserialize)]
+pub struct DefaultClaims {}
+
+/// Extracts a JSON Web Token (JWT) from the `Authorization` request header. The header value can optionally be prefixed with `Bearer `.
+///
+/// Parses the `Authorization` header and validates the structure of the header value to be a JSON Web Token
+/// Returns 400 Bad Request if the header value is empty or parsing fails.
```
juv, line 19:
```suggestion
/// Returns 400 Bad Request if the header value is empty or parsing fails.
/// Returns 401 Unauthorized if the `kid` of the JWT is missing in the configured JWKS or the JWT fails the validation based on the configured `validation` (covering token expiration, token audience, ...)
```

rapina/src/jwt/extract.rs
```diff
@@ -0,0 +1,350 @@
...
+/// ```ignore
+/// use rapina::prelude::*;
+///
+/// #[derive(Clone)]
+/// struct AppConfig {
+///     db_url: String,
+/// }
+///
+/// #[get("/config")]
+/// async fn get_config(state: JsonWebToken) -> StatusCode {
```
juv:
```suggestion
///
/// #[get("/config")]
/// async fn get_config(token: JsonWebToken) -> StatusCode {
```

rapina/src/jwt/extract.rs
```diff
@@ -0,0 +1,350 @@
...
+            .ok_or_else(|| Error::unauthorized("missing authorization header"))?
+            .to_str()
+            .map_err(|_| {
+                Error::unauthorized("Authorization header could not be parsed as String")
+            })?;
+
+        let jwks_client: Option<&JwksClient> = state.get::<JwksClient>();
+        let validation: Option<&Validation> = state.get::<Validation>();
+
+        println!("JWKS config: {:?}", jwks_client.is_some());
```
juv:
```suggestion
```

rapina/src/jwt/extract.rs
```diff
@@ -0,0 +1,350 @@
...
+        let jwks: Option<JwkSet> = match jwks_client {
+            Some(client) => Some(client.get_jwks_content().await?),
+            None => None,
+        };
+
+        let Some(jwks) = jwks else {
+            return Err(Error::internal("Internal authentication error"));
+        };
+
+        println!("JWKS: {:?}", jwks);
```
juv:
```suggestion
```

rapina/src/jwt/extract.rs
```diff
@@ -0,0 +1,350 @@
...
+                .route(
+                    http::Method::GET,
+                    "/realms/master/protocol/openid-connect/cert",
+                    |_, _, _| async { AUTH0_SAMPLE_JWKS },
+                )
+                .route(
+                    http::Method::GET,
+                    "/realms/master/.well-known/openid-configuration",
+                    |req, _, _| async move {
+                        println!("{:?}", req);
```
juv:
```suggestion
```

rapina/src/jwt/extract.rs
```diff
@@ -0,0 +1,350 @@
...
+
+        let jwks_server = TestClient::new(setup_jwks_server()).await;
+        let jwks_client = setup_jwks_client(jwks_server.addr());
+
+        let state = AppState::new().with(jwks_client).with(custom_validation);
+
+        let result: Result<JsonWebToken<DefaultClaims>, Error> =
+            JsonWebToken::from_request_parts(&parts, &empty_params(), &Arc::new(state)).await;
+
+        println!("{:?}", result);
```
juv:
```suggestion
```

rapina/src/jwt/extract.rs
```diff
@@ -0,0 +1,350 @@
...
+
+        let jwks_server = TestClient::new(setup_jwks_server_oidc_discovery()).await;
+        let jwks_client = setup_jwks_client_oidc_discovery(jwks_server.addr());
+
+        let state = AppState::new().with(jwks_client).with(custom_validation);
+
+        let result: Result<JsonWebToken<DefaultClaims>, Error> =
+            JsonWebToken::from_request_parts(&parts, &empty_params(), &Arc::new(state)).await;
+
+        println!("{:?}", result);
```
juv:
```suggestion
```

rapina/src/jwt/extract.rs
```diff
@@ -0,0 +1,350 @@
...
+
+        let jwks_server = TestClient::new(setup_jwks_server()).await;
+        let jwks_client = setup_jwks_client(jwks_server.addr());
+
+        let state = AppState::new().with(jwks_client);
+
+        let result: Result<JsonWebToken<DefaultClaims>, Error> =
+            JsonWebToken::from_request_parts(&parts, &empty_params(), &Arc::new(state)).await;
+
+        println!("{:?}", result);
```
juv:
```suggestion
```

rapina/src/jwt/extract.rs
```diff
@@ -0,0 +1,350 @@
...
```
juv:
```suggestion
```

rapina/src/jwt/jwks_client.rs
juv, line 1: This feature must be covered in the docs too

rapina/examples/jwt-validation/src/main.rs
```diff
@@ -0,0 +1,64 @@
...
+use rapina::jwt::JwksProvider;
+use rapina::jwt::{JsonWebToken, JwksClient};
+use rapina::prelude::*;
+
+#[derive(Deserialize)]
+struct GoogleClaims {
+    pub email: String,
+}
+#[get("/users")]
+async fn list_users(token: JsonWebToken<GoogleClaims>) -> Json<String> {
```
juv: the function name is kind of misleading

rapina/src/jwt/extract.rs
```diff
@@ -0,0 +1,350 @@
...
+{
+    pub fn new(
+        jwks: JwkSet,
+        validation: Option<&Validation>,
+        token: String,
+    ) -> Result<Self, Error> {
+        let token = token.trim_start_matches("Bearer ");
+        let jwt_header = Self::parse_header(token)?;
+
+        tracing::debug!("jwt header: {:?}", jwt_header);
```
juv: Can potentially be removed? Or do we want to add some debug logging for the token validation?

rapina/src/jwt/extract.rs
arferreira: Both printlns in from_request_parts should go.

rapina/src/jwt/extract.rs
```diff
@@ -0,0 +1,350 @@
...
+        let jwks_client: Option<&JwksClient> = state.get::<JwksClient>();
+        let validation: Option<&Validation> = state.get::<Validation>();
+
+        println!("JWKS config: {:?}", jwks_client.is_some());
+
+        let jwks: Option<JwkSet> = match jwks_client {
+            Some(client) => Some(client.get_jwks_content().await?),
+            None => None,
+        };
+
```
arferreira, line 144: This fires a network request to your identity provider on every single authenticated request. The JWKS endpoint is designed to be fetched once and cached with a TTL, not polled per-request. Most providers will start rate-limiting you quickly under any real traffic.

rapina/src/jwt/jwks_client.rs
```diff
@@ -0,0 +1,110 @@
...
+pub fn default_validation() -> Validation {
+    let mut validation = Validation::default();
+
+    // account for 10 seconds of clock skew per default
+    validation.leeway = 10;
+
+    // enable aud (audience), exp (expiration) and nbf (not before) field validation
+    validation.validate_aud = true;
+    validation.validate_exp = true;
+    validation.validate_nbf = true;
```
arferreira, line 132: `nbf` is an optional JWT claim per [RFC 7519](https://www.rfc-editor.org/rfc/rfc7519#section-4.1.5). Setting as true means the validator will reject any token that doesn't include the claim at all, which is probably not what you want. Most tokens in the wild, especially from machine-to-machine flows, don't set nbf. I'd leave this false unless you have a specific reason to require it.
```suggestion
    validation.validate_nbf = false;
```

rapina/src/jwt/mod.rs
```diff
@@ -0,0 +1,9 @@
+mod extract;
+mod jwks_client;
+
+pub use extract::JsonWebToken;
+pub use jwks_client::HttpsClient;
+pub use jwks_client::JwksClient;
+pub use jwks_client::JwksProvider;
+pub use jwks_client::build_http_client;
```
arferreira: `HttpsClient`, `JwksProvider`, and `build_http_client` are implementation details of `JwksClient`. By re-exporting them from the crate root they become part of rapina's public API surface, meaning any future rename or removal is a semver-breaking change. I'd keep them `pub(crate)` and only expose what callers actually need to configure things: `JsonWebToken`, `JwksClient`, and `default_validation`.

rapina/src/jwt/jwks_client.rs
juv: It is implemented differently in the `jsonwebtoken` crate.
https://github.com/Keats/jsonwebtoken/blob/master/src/validation.rs#L57-L65 

```
    /// Validation only happens if `nbf` claim is present in the token.
    /// Adding `nbf` to `required_spec_claims` will make it required.
```

In case we want Rapina to enforce the presence of any field, i.e. `exp` and `aud`, we need to put it into them into `required_spec_claims`: `validation.set_required_spec_claims(&["exp", "aud"]);`

juv: regarding the `nbf = false`, the docs of the `jsonwebtoken` crate mention this:
`    /// Defaults to `true`. Very insecure to turn this off. Only do this if you know what you are doing.`

For this implementation, I have chosen to agree with that statement and set it to `true` per default here.
This forces the developer to actually call the `validation.set_audience` like in the example: https://github.com/rapina-rs/rapina/pull/408/changes#diff-dbfd7c708292946e1472281f08eea1fe72b122b0f3af80df0d4edc4982e4b51fR50-R51 

Otherwise, the JWT validation will fail per default! In case the developer wants to explicitly turn `aud` validation off, he can just override it by setting `validate_aud = false`. 

This must be explained properly in the docs of course. But I'd rather have a secure implementation per default instead of not applying best practices (although in this case it comes with a few-lines-of-code overhead for the developer). Lmk what you think

docs/content/docs/core-concepts/jwks.md
juv, line 1: This docs file needs to be updated when the cronjob scheduling for refreshing the JWKS content is built in

rapina/src/jwt/extract.rs
juv: 100%, this is a must have feature before this PR can be merged.
I have created #420 to track the progress for this topic as discussed on Discord. 

How do we proceed with this pr in the meantime? Do I wait for that #420 to be finished?

rapina/src/jwt/jwks_client.rs
arferreira: Sure, good call! Just remember mentioning this in our docs so it can be changed if desired.

rapina/src/jwt/extract.rs
arferreira: Yeahh we can't merge without that.

rapina/src/jwt/jwks_client.rs
```diff
@@ -0,0 +1,345 @@
...
+
+    validation
+}
+
+pub(crate) fn build_http_client() -> HttpsClient {
+    let http_client = hyper_rustls::HttpsConnectorBuilder::new()
+        .with_native_roots()
+        .expect("no native root CA certificates found")
+        .https_or_http()
```
arferreira: What happens if an attacker an http endpoint and inject their own pks and sign tokens? Does the application will accept that? Should it be https only?

rapina/src/jwt/jwks_client.rs
```diff
@@ -0,0 +1,345 @@
...
+        refresh_schedule: String,
+        cache: Arc<RwLock<Option<JwkSet>>>,
+    },
+    Oidc {
+        client: HttpsClient,
+        discovery_url: String,
+        refresh_schedule: String,
+        cache: Arc<RwLock<Option<JwkSet>>>,
+    },
+}
```
arferreira, line 32: I think we could wrap the inner representation in a private type like:

``` rust
#[derive(Clone)]
pub struct JwksClient(Arc<JwksClientInner>);

enum JwksClientInner {
      Direct {
          jwks_url: String,
          refresh_schedule: String,
          cache: Arc<RwLock<Option<JwkSet>>>,
      },
      Oidc {
          discovery_url: String,
          refresh_schedule: String,
          cache: Arc<RwLock<Option<JwkSet>>>,
      },
}
```
The `Arc` wrapping also makes `clone()` a single atomic increment instead of cloning the `HttpsClient`.

rapina/src/jwt/extract.rs
```diff
@@ -0,0 +1,348 @@
...
+            v.algorithms = vec![jwt_header.alg];
+            v
+        } else {
+            let mut v = jwt::default_validation();
+            v.algorithms = vec![jwt_header.alg];
+            v
+        };
+
+        let decoding_key =
+            DecodingKey::from_jwk(jwk).map_err(|_| Error::unauthorized("Failed to decode JWKS"));
```
arferreira: Sorry, we could add a tracing here.

rapina/src/jwt/extract.rs
```diff
@@ -0,0 +1,348 @@
...
+
+        match decode::<JsonWebToken<T>>(token, &decoding_key?, &validation) {
+            Ok(decoded_token) => Ok(decoded_token.claims),
+            Err(e) => Err(Error::unauthorized(format!(
+                "failed to decode token: {:?}",
```
arferreira: This leaks `jsonwebtoken`, right?
```suggestion
                "failed to decode token: {}",
```

rapina/examples/jwt-validation/src/main.rs
```diff
@@ -0,0 +1,66 @@
...
+     */
+    tracing_subscriber::fmt().init();
+
+    let router = Router::new().get("/email", get_email);
+
+    // OIDC Discovery endpoint of Google Accounts API
+    let discovery_url = "https://accounts.google.com/.well-known/openid-configuration";
+
+    // Cron schedule of 5 minutes to periodically refresh the JWKS content
+    let cron_refresh_schedule = "* */5 * * * *";
```
arferreira: This fires every second of every 5th minute, not once every 5 minutes.
```suggestion
    let cron_refresh_schedule = "0 */5 * * * *";
```

rapina/examples/jwt-validation/src/main.rs
```diff
@@ -0,0 +1,66 @@
...
+use rapina::jwt::{JsonWebToken, JwksClient};
+use rapina::prelude::*;
+
+#[derive(Deserialize)]
+struct GoogleClaims {
+    pub email: String,
+}
+#[get("/email")]
+async fn get_email(token: JsonWebToken<GoogleClaims>) -> Json<String> {
+    println!("Token subject: {}", token.sub);
```
arferreira: What you think?
```suggestion
    tracing::info!(sub = %token.sub, "authenticated request");
```

arferreira (commented):
Looks great, I think I found some parts we could improve before merging.

rapina/src/jwt/extract.rs
juv: before: `"message": "failed to decode token: Error(InvalidSignature)"`
after: `"message": "failed to decode token: InvalidSignature"`

rapina/src/jwt/jwks_client.rs
juv: Is your main concern the two string heap allocations for each cron tick? 

As far as I understand, the `HttpsClient` should be cheap to clone, because hyper's `Client` struct has this comment:
```
A Client to make outgoing HTTP requests.
Client is cheap to clone and cloning is the recommended way to share a Client. The underlying connection pool will be reused.
```

Maybe we can simply [change this line ](https://github.com/rapina-rs/rapina/pull/408/changes#diff-5b6e1671fc646273b6ae243e206d37b4bfd9de71107e1b5b300a97a00f27025aR442)to `let jwks_client = Arc::new(jwks_client)`?

arferreira: Sorry for my late answer 😅, fair point you're right, `HttpsClient` is cheap to clone, that concern will be more relevant once the API stabilizes, not a blocker, let's get this merged.

Issue #339, opened 2026-03-13
Background jobs: in-process worker with job dispatch


arferreira:
The worker is a background task that polls `rapina_jobs` for claimable work, dispatches to registered handlers, and manages job lifecycle (pending → running → completed/failed). In Phase 1 it runs in-process alongside the HTTP server — no separate binary.

The poll loop runs on a configurable interval (default 5s), executes the claiming query from the RFC to grab a batch of jobs, and for each claimed job: sets `status = 'running'` and `locked_until = NOW() + timeout`, looks up the handler by `job_type` name in the `inventory` registry, deserializes the payload, injects dependencies from `AppState`, and calls the handler. On success, sets `status = 'completed'` and `finished_at`. On failure, delegates to the retry logic (#338).

The worker spawns via `tokio::spawn` during server startup when jobs are configured. It needs to respect graceful shutdown — when the server receives SIGTERM/SIGINT, the worker finishes its current job and stops claiming new ones. See `rapina/src/server.rs` for how the shutdown signal propagates; the worker should use the same `tokio::select!` pattern with a shutdown channel.

Single-queue polling in Phase 1. Multi-queue with weighted round-robin comes in Phase 2.

`trace_id` restoration: when the worker picks up a job that has a `trace_id`, it should set that trace_id on the worker's tracing span so logs from the job execution are correlated with the original HTTP request.

See [RFC: Background Jobs (discussion #334)](https://github.com/rapina-rs/rapina/discussions/334), "Worker Model" section.

Depends on #335, #336, #338.

PR #426, opened 2026-03-29
Implement RetryPolicy with exponential backoff

### LINKED ISSUE #338: Background jobs: retry with exponential backoff
When a job handler returns an error, the worker needs to decide whether to retry or mark the job as permanently failed. The default strategy is exponential backoff with jitter: attempt 1 immediate, attempt 2 at 1s + jitter, attempt 3 at 4s + jitter, attempt 4 at 16s + jitter. The formula is `base_delay * 4^(attempt - 1) + random(0..base_delay)`.

On failure: increment `attempts`, set `last_error` to the error message. If `attempts < max_retries`, set `status = 'pending'` and `run_at = NOW() + backoff_delay` — the job goes back into the claimable pool after the delay. If `attempts >= max_retries`, set `status = 'failed'` and `finished_at = NOW()`. Failed jobs stay in the table for inspection and manual retry via CLI.

The `RetryPolicy` enum lives in `rapina/src/jobs/`:

```rust
RetryPolicy::exponential(max_retries, base_delay)
RetryPolicy::fixed(max_retries, delay)
RetryPolicy::none()
```

Phase 1 only needs `exponential` as the default. `fixed` and `none` are trivial to add but can wait if scope needs trimming.

See [RFC: Background Jobs (discussion #334)](https://github.com/rapina-rs/rapina/discussions/334), "Retry Strategy" section.

Depends on #335.

ShiraiEd:
## Summary

- Adds `RetryPolicy` enum with `exponential`, `fixed`, and `none` variants
- Implements exponential backoff with jitter: attempt 1 immediate, attempt 2 at 1s+jitter, attempt 3 at 4s+jitter, attempt 4 at 16s+jitter, capped at 1 week
- Adds `apply_failure` — increments attempts, reschedules as `pending` with backoff or permanently marks as `failed`
- Adds `apply_success` — marks job as `completed`

The exponent uses `attempts - 2` instead of `attempts - 1` from the issue formula. With `attempts - 1`, attempt 2 would produce `4^1 = 4s` instead of `1s`. Subtracting 2 aligns the math with the concrete examples in the spec.

Anyone doing issue #339 needs to delete the `#[allow(dead_code)]` in retry.rs

## Related Issues 

Closes #338 

## Checklist

- [x] `cargo fmt` passes
- [x] Tests pass

### DIFF
--- a/docs/content/docs/core-concepts/caching.md
+++ b/docs/content/docs/core-concepts/caching.md
@@ -111,7 +111,7 @@ For multi-instance deployments where all instances need to share the same cache 
 
 ```toml
 [dependencies]
-rapina = { version = "0.7", features = ["cache-redis"] }
+rapina = { version = "0.10.0", features = ["cache-redis"] }
 ```
 
 ```rust
--- a/docs/content/docs/core-concepts/database.md
+++ b/docs/content/docs/core-concepts/database.md
@@ -13,7 +13,7 @@ Add the database feature to your `Cargo.toml`:
 
 ```toml
 [dependencies]
-rapina = { version = "0.6.0", features = ["postgres"] }
+rapina = { version = "0.10.0", features = ["postgres"] }
 # or "mysql", "sqlite"
 ```
 
--- a/docs/content/docs/core-concepts/metrics.md
+++ b/docs/content/docs/core-concepts/metrics.md
@@ -13,7 +13,7 @@ Add the feature to your `Cargo.toml`:
 
 ```toml
 [dependencies]
-rapina = { version = "0.6.0", features = ["metrics"] }
+rapina = { version = "0.10.0", features = ["metrics"] }
 ```
 
 Enable the endpoint in your application:
--- a/docs/content/docs/core-concepts/migrations.md
+++ b/docs/content/docs/core-concepts/migrations.md
@@ -15,7 +15,7 @@ Your `Cargo.toml` needs the `database` feature and a database driver:
 
 ```toml
 [dependencies]
-rapina = { version = "0.8", features = ["sqlite"] }
+rapina = { version = "0.10.0", features = ["sqlite"] }
 ```
 
 Replace `sqlite` with `postgres` or `mysql` depending on your database. You also need a database connection configured in your app — see the [Database](/docs/core-concepts/database/) page if you haven't set that up yet.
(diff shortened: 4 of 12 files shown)


rapina/src/jobs/retry.rs
```diff
@@ -0,0 +1,263 @@
...
+/// # Example
+///
+/// ```rust,ignore
+/// use rapina::jobs::RetryPolicy;
+/// use std::time::Duration;
+///
+/// // Exponential backoff: immediate, 1s, 4s, 16s, then fail.
+/// let policy = RetryPolicy::exponential(5, Duration::from_secs(1));
+/// ```
+pub enum RetryPolicy {
```
arferreira, line 23: This is public but doesn't derive `Clone` or `Debug`, someone storing this in an app config struct or logging it will hit an immediate wall.
```suggestion
#[derive(Clone, Debug)]
pub enum RetryPolicy {
```

rapina/src/jobs/retry.rs
```diff
@@ -0,0 +1,263 @@
...
+    if base.is_zero() {
+        return Duration::ZERO;
+    }
+    let nanos = std::time::SystemTime::now()
+        .duration_since(std::time::UNIX_EPOCH)
+        .unwrap_or_default()
+        .subsec_nanos() as u64;
+    let base_nanos = base.as_nanos().min(u128::from(u64::MAX)) as u64;
+    Duration::from_nanos(nanos % base_nanos)
+}
```
arferreira: The whole point of jitter is to spread retries from workers that failed around the same time. Workers failing within the same millisecond will read nearly identical `subsec_nanos` values and produce nearly identical jitter, exactly the thundering herd you're trying to avoid. Seed from `job_id` instead, which is unique per job regardless of timing and requires no new dependency.

rapina/src/jobs/retry.rs
```diff
@@ -0,0 +1,263 @@
...
+
+/// Computes the exponential backoff delay for a given attempt count.
+///
+/// Capped at one week so `run_at` never becomes absurdly far in the future.
+#[allow(dead_code)]
```
arferreira, line 112: Leave a `// TODO: remove once the worker lands` comment on each, or better, track it in the PR description so the worker PR knows to clean these up.

PR #431, opened 2026-03-31
add new worker.rs file, new jobs() app method, updates retry.rs mod.r…

### LINKED ISSUE #339: Background jobs: in-process worker with job dispatch

ShiraiEd:
## Summary
**Worker** (`rapina/src/jobs/worker.rs`)
- `JobConfig`  configurable poll interval (5s), batch size (10), queues (`["default"]`), and job lock timeout (30s)
- Poll loop that claims jobs atomically via a CTE `UPDATE … FROM … RETURNING` with `FOR UPDATE SKIP LOCKED`, so concurrent workers never process the same row
- Handler dispatch via `inventory::iter::<JobDescriptor>`  looks up by `job_type`, no manual registration needed
- Trace ID restoration: restores the enqueuing request's `trace_id` into the worker's tracing span for log correlation
- Graceful shutdown via independent SIGINT/SIGTERM listeners  same `tokio::select!` pattern as `server.rs`, finishes the current batch before stopping
- **Wiring** (`app.rs`)
- `.jobs(JobConfig)` builder method on `Rapina`
- Worker spawns via `tokio::spawn` in `listen()` when jobs are configured, sharing `AppState` with the HTTP server 
**Cleanup** (`retry.rs`)
- Removed all `#[allow(dead_code)]` and `// TODO: remove once the worker lands` markers  `apply_failure` and `apply_success` are now used
**Docs**
- `jobs/mod.rs` module doc updated to reflect the full system (worker, lifecycle, trace propagation)
- `docs/content/docs/core-concepts/background-jobs.md`  added sections for `#[job]` macro, `Jobs` extractor, worker configuration, job lifecycle, retry policy, DI limitations, and trace propagation
## Related Issues

Closes #339

## Checklist

- [x] `cargo fmt` passes
- [x] Tests pass

### DIFF
--- a/rapina/src/jobs/worker.rs
+++ b/rapina/src/jobs/worker.rs
@@ -32,9 +32,10 @@ use std::sync::Arc;
 use std::time::Duration;
 
 use sea_orm::{ConnectionTrait, DatabaseConnection, DbBackend, FromQueryResult, Statement, Value};
+use tracing::Instrument;
 
 use crate::jobs::retry::{apply_failure, apply_success};
-use crate::jobs::{JobDescriptor, JobRow, RetryPolicy};
+use crate::jobs::{JobDescriptor, JobResult, JobRow, RetryPolicy};
 use crate::state::AppState;
 
 /// Configuration for the in-process background job worker.
@@ -109,8 +110,13 @@ impl JobConfig {
     /// ```rust,ignore
     /// JobConfig::default().queues(["default", "emails", "heavy"])
     /// ```
+    ///
+    /// check if queues are empty
+    /// if someone calls .queues([]), the build_claim_stmt generates `IN ()` which is invalid
     pub fn queues(mut self, queues: impl IntoIterator<Item = impl Into<String>>) -> Self {
-        self.queues = queues.into_iter().map(Into::into).collect();
+        let q: Vec<String> = queues.into_iter().map(Into::into).collect();
+        assert!(!q.is_empty(), "queues must not be empty");
+        self.queues = q;
         self
     }
 
@@ -252,14 +258,16 @@ impl Worker {
             trace_id = job.trace_id.as_deref().unwrap_or(""),
         );
 
-        let result = {
-            let _guard = span.enter();
-            (descriptor.handle)(job.payload.clone(), self.state.clone()).await
-        };
-
-        // The retry policy type defaults to exponential; max_retries comes from
-        // the job row so each handler's `#[job(max_retries = N)]` is respected.
-        let policy = RetryPolicy::exponential(job.max_retries, Duration::from_secs(1));
+        let result: JobResult = (descriptor.handle)(job.payload.clone(), self.state.clone())
+            .instrument(span)
+            .await;
+        // Policy type and base delay come from the descriptor (set by `#[job]`
+        // attributes at compile time); max_retries comes from the job row.
+        let policy = build_policy(
+            descriptor.retry_policy,
+            job.max_retries,
+            descriptor.retry_delay_secs,
+        );
 
         match result {
             Ok(()) => {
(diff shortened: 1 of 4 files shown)


rapina/src/jobs/worker.rs
```diff
@@ -0,0 +1,473 @@
...
+        // correlated with the HTTP request that enqueued the job.
+        let span = tracing::info_span!(
+            "job",
+            job_type = %job.job_type,
+            job_id   = %job.id,
+            trace_id = job.trace_id.as_deref().unwrap_or(""),
+        );
+
+        let result = {
+            let _guard = span.enter();
```
arferreira: `span.enter()` holds a sync guard across an `.await` point, which is a known footgun in async code. While this task is suspended waiting for the handler, any other task that runs on the same thread will appear inside this span in the logs. Is there a reason not to use `.instrument()` here?

```suggestion
        let result = (descriptor.handle)(job.payload.clone(), self.state.clone())
            .instrument(span)
            .await;
```

rapina/src/jobs/worker.rs
```diff
@@ -0,0 +1,473 @@
...
+            (descriptor.handle)(job.payload.clone(), self.state.clone()).await
+        };
+
+        // The retry policy type defaults to exponential; max_retries comes from
+        // the job row so each handler's `#[job(max_retries = N)]` is respected.
+        let policy = RetryPolicy::exponential(job.max_retries, Duration::from_secs(1));
```
arferreira: The worker always constructs `RetryPolicy::Exponential` with a hardcoded 1s base regardless of what the `#[job]` macro configured. `RetryPolicy::Fixed` and `RetryPolicy::None` have no path to the worker. Were `Fixed` and `None` intended to be configurable per job, or is `Exponential` the only supported policy for now? If the latter, should we hold off on exposing the other variants until the storage side is in place?

rapina/src/jobs/worker.rs
```diff
@@ -0,0 +1,473 @@
...
+    }
+
+    /// Overrides the list of queues to subscribe to.
+    ///
+    /// # Example
+    ///
+    /// ```rust,ignore
+    /// JobConfig::default().queues(["default", "emails", "heavy"])
+    /// ```
+    pub fn queues(mut self, queues: impl IntoIterator<Item = impl Into<String>>) -> Self {
```
arferreira, line 116: `build_claim_stmt` generates `IN ($1, $2, ...)` dynamically. If someone calls `.queues([])` the result is `IN ()` which Postgres will reject at runtime. Would it make sense to guard here?

```suggestion
    pub fn queues(mut self, queues: impl IntoIterator<Item = impl Into<String>>) -> Self {
        let q: Vec<String> = queues.into_iter().map(Into::into).collect();
        assert!(!q.is_empty(), "queues list must not be empty");
        self.queues = q;
        self
    }
```

rapina/src/jobs/worker.rs
ShiraiEd: Added into #[job], so it can be read on the macro, the default is set to exponential with 1s base_delay

arferreira (approved):
LGTM

PR #432, opened 2026-03-31
Add cronjob scheduler feature

### LINKED ISSUE #420: [Feature]: Cron scheduler for non-durable, non-transactional background jobs
### Problem Statement

Currently, Rapina lacks a built-in mechanism to easily schedule and manage background tasks that run at regular intervals. This requires Rapina users to implement implement their own solution for periodic jobs, e.g. periodic polling, data updates, or automated cleanups.

The scope for this feature are **non-durable, non-transactional background jobs**. Durable, transactional background jobs are already covered in Rapina by https://github.com/rapina-rs/rapina/issues/337 

### Proposed Solution

Introduce a cron scheduler module in Rapina, allowing developers to define and spawn background jobs with a simple, declarative API.

This scheduler should support common cron expressions, job cancellation (to cover them part of the graceful shutdown?) and safe error handling

### Alternatives Considered

Developers can manually spawn tasks with custom timers or use third-party crates, but these require boilerplate, lack uniform observability, and fragment the developer experience compared to a built-in solution.

Predictability (conventions, DX), Auditability (contracts, errors, observability)

### Contribution

juv:
## Summary

Adds a cronjob scheduler feature

## Related Issues

Closes #420 

## Checklist

- [ ] `cargo fmt` passes
- [ ] Tests pass

### DIFF
--- a/rapina/src/cron_scheduler.rs
+++ b/rapina/src/cron_scheduler.rs
@@ -159,7 +159,9 @@ impl CronScheduler {
                 .await
                 .expect("Failed to shutdown Rapina background jobs");
         }
-        tracing::info!("Cron scheduler stopped");
+        tracing::info!(
+            "Cron scheduler stopped. Blocking tasks may still be in-flight and will keep the process alive until complete."
+        );
     }
 
     /// Returns the number of scheduled jobs, which have not yet been drained by starting the cron scheduler


rapina/src/cron_scheduler.rs
juv, line 1: TODO: add docs/ page

rapina/src/cron_scheduler.rs
```diff
@@ -0,0 +1,284 @@
+use std::sync::Arc;
+use std::sync::atomic::Ordering;
+use tokio_cron_scheduler::{Job, JobScheduler};
+use tokio_util::sync::CancellationToken;
+
+/// A background job scheduler that wraps `tokio_cron_scheduler::JobScheduler`.
+///
+/// It provides integrated graceful shutdown capabilities by utilizing a
```
juv: This sentence can be removed maybe? It feels a bit unnecessary now after reviewing myself once more

rapina/src/cron_scheduler.rs
```diff
@@ -0,0 +1,284 @@
...
+                }
+                result = task() => {
+                    if let Err(e) = result {
+                        tracing::error!("Error while running Rapina background job: {:?}", e);
+                    }
+                }
+                }
+            })
+        })
+        .expect("Failed to add Rapina background job");
```
juv:
```suggestion
        .expect("Failed to create Rapina background job template");
```

rapina/src/cron_scheduler.rs
```diff
@@ -0,0 +1,284 @@
...
+    /// Initiates a graceful shutdown of the scheduler and its running tasks.
+    ///
+    /// This method triggers the `CancellationToken` to notify all currently running tasks to stop.
+    /// It then shuts down the underlying `JobScheduler`.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the underlying `JobScheduler` fails to shut down properly.
+    pub async fn shutdown(&mut self) {
+        // instantly trigger all tokio::select! branches that watch the token cancellation, stopping the next execution to
```
juv:
```suggestion
        // instantly trigger all tokio::select! branches that watch the token cancellation, stopping the execution immediately in case no blocking code is being executed
```

rapina/src/cron_scheduler.rs
```diff
@@ -0,0 +1,284 @@
...
+        self.cancellation_token.cancel();
+
+        if let Some(mut scheduler) = self.scheduler.take() {
+            scheduler
+                .shutdown()
+                .await
+                .expect("Failed to shutdown Rapina background jobs");
+        }
+        tracing::info!("Shutdown signal received, stopping cron scheduler",);
```
juv:
```suggestion
        tracing::info!("Shutdown signal received, stopping cron scheduler");
```

rapina/Cargo.toml
```diff
@@ -111,7 +115,7 @@ name = "multipart"
 required-features = ["multipart"]
 
 [features]
-default = ["compression"]
+default = ["compression", "cron-scheduler"]
```
juv: `cron-scheduler` should not be in default feature list

rapina/src/cron_scheduler.rs
```diff
@@ -0,0 +1,284 @@
+use std::sync::Arc;
+use std::sync::atomic::Ordering;
+use tokio_cron_scheduler::{Job, JobScheduler};
+use tokio_util::sync::CancellationToken;
+
+/// A background job scheduler that wraps `tokio_cron_scheduler::JobScheduler`.
+///
+/// It provides integrated graceful shutdown capabilities by utilizing a
+/// `CancellationToken` to signal running tasks to stop.
```
juv:
```suggestion
```

docs/content/docs/core-concepts/cron-scheduler.md
```diff
@@ -0,0 +1,178 @@
...
+description = "Run recurring background tasks on a cron schedule"
+weight = 10
+date = 2026-04-01
++++
+
+The cron scheduler lets you run recurring tasks inside your Rapina application on a cron-based schedule, e.g. health checks, cache warm-ups, periodic cleanup, report generation. Anything that needs to happen on a timer without an external trigger.
+
+Rapina makes scheduling cron jobs easy and integrates it into the application lifecycle. Scheduled jobs start automatically when the server boots and shut down gracefully on SIGINT/SIGTERM. Running tasks are cancelled cooperatively.
+
+## Cron Scheduler vs Background Jobs
```
juv, line 12: maybe the section `Cron Scheduler vs Background Jobs`  of the docs should also be added to the `background-jobs.md` file?

rapina/Cargo.toml
```diff
@@ -91,6 +91,10 @@ hyper-tungstenite = { version = "0.19", optional = true }
 tokio-tungstenite = { version = "0.28", optional = true }
 futures-util = { version = "0.3", optional = true, default-features = false, features = ["sink"] }
 
+# CronJob Scheduler (optional)
```
juv:
```suggestion
# Cron Scheduler (optional)
```

docs/content/docs/core-concepts/cron-scheduler.md
arferreira: Perhaps we could just mention, here I think this makes more sense, but in case someone got there, would be okay having something referencing that we as well provide this.

rapina/src/cron_scheduler.rs
```diff
@@ -0,0 +1,289 @@
+use std::sync::Arc;
```
arferreira, line 5: Could you add a module-level doc comment at the top? Not a blocker for the PR but it keeps the pattern consistent with the rest of the codebase. 
Something like: 
```suggestion
//! Background task scheduler for recurring cron-based jobs.
//!
//! For durable, persistent work use [`crate::jobs`] instead. This scheduler
//! is in-memory only and does not survive process restarts.
use std::sync::Arc;
```

rapina/src/cron_scheduler.rs
```diff
@@ -0,0 +1,289 @@
...
+
+    /// Returns whether the list of scheduled jobs, which have not yet been drained by starting the cron scheduler, is empty
+    pub fn is_empty(&self) -> bool {
+        self.jobs.is_empty()
+    }
+
+    /// Returns whether the cron scheduler has been started
+    pub fn is_started(&self) -> bool {
+        if let Some(scheduler) = &self.scheduler {
+            scheduler.inited.load(Ordering::Relaxed)
```
arferreira: Is `scheduler.inited` an internal field of `tokio_cron_scheduler`? A patch bump can rename or remove it and this breaks silently.
Can you track this with your own field instead?

arferreira: Maybe a `started: bool` would work!

rapina/src/cron_scheduler.rs
```diff
@@ -0,0 +1,289 @@
...
+                }
+                result = task() => {
+                    if let Err(e) = result {
+                        tracing::error!("Error while running Rapina background job: {:?}", e);
+                    }
+                }
+                }
+            })
+        })
+        .expect("Failed to create Rapina background job template");
```
arferreira: Well this signature says a `io::Result<()>`, but panics with `expect()`, either drop that `Result` and document the panic or propagate the error, having both is confusing.

rapina/src/cron_scheduler.rs
```diff
@@ -0,0 +1,289 @@
...
+    ///
+    /// For graceful shutdown semantics to work as expected, make sure to not run blocking code as part of `task`.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the job cannot be created or if it fails to be added to the underlying scheduler.
+    pub fn schedule<F, Fut>(&mut self, cron_schedule: String, task: F) -> std::io::Result<()>
+    where
+        F: Fn() -> Fut + Send + Sync + 'static,
+        Fut: Future<Output = std::io::Result<()>> + Send + 'static,
```
arferreira: `std::io::Result` is narrow for a web framework context. A task doing DB calls or HTTP requests won't produce `io::Error`. Would `Box<dyn std::error::Error + Send + Sync>` work here?

rapina/src/app.rs
```diff
@@ -125,6 +128,8 @@ impl Rapina {
             rfc7807_base_uri: "about:blank".to_string(),
             #[cfg(feature = "database")]
             jobs_config: None,
+            #[cfg(feature = "cron-scheduler")]
+            cron_scheduler: Some(CronScheduler::new()),
```
arferreira: This allocates a scheduler on every `Rapina::new()` even when no `.cron()` calls follow. The `start_cronjob_scheduler` path then creates an empty `JobScheduler` and registers a shutdown hook for nothing. Default to `None` and initialize lazily on the first `.cron()` call instead.

docs/content/docs/core-concepts/cron-scheduler.md
```diff
@@ -0,0 +1,178 @@
...
+| **Use case** | Periodic maintenance, polling, cache refresh | Durable, transactional deferred work: emails, uploads, reports |
+| **Infrastructure** | No extra dependencies                        | Requires PostgreSQL                                            |
+
+## Prerequisites
+
+Enable the `cron-scheduler` feature flag:
+
+```toml
+[dependencies]
+rapina = { version = "0.10", features = ["cron-scheduler"] }
```
arferreira: 😁
```suggestion
rapina = { version = "0.11", features = ["cron-scheduler"] }
```

rapina/src/lib.rs
```diff
@@ -88,6 +88,8 @@ pub mod cache;
 pub mod cache_redis;
 pub mod config;
 pub mod context;
+#[cfg(feature = "cron-scheduler")]
+pub mod cron_scheduler;
```
arferreira: Users never construct `CronScheduler` directly. Keeping it pub locks us into maintaining those types as public API
```suggestion
pub(crate) mod cron_scheduler;
```

rapina/src/cron_scheduler.rs
```diff
@@ -0,0 +1,341 @@
...
+//! For durable, persistent work use [`crate::jobs`] instead. This scheduler
+//! is in-memory only and does not survive process restarts.
+use std::sync::Arc;
+use tokio_cron_scheduler::{Job, JobScheduler};
+use tokio_util::sync::CancellationToken;
+
+#[derive(Debug)]
+pub struct CronSchedulerError {
+    pub cron_schedule: String,
+    pub error: tokio_cron_scheduler::JobSchedulerError,
```
arferreira: And update the `Display` impl to populate it. Leaking `JobSchedulerError` from a dependency ties the public error type to the underlying scheduler library.
```suggestion
    pub error_message: String,
```

rapina/src/cron_scheduler.rs
```diff
@@ -0,0 +1,341 @@
...
+            let task = task.clone();
+            Box::pin(async move {
+                tokio::select! {
+                _ = cronjob_cancellation_token.cancelled() => {
+                    tracing::info!("Shutdown signal received, stopping job scheduler");
+                }
+                result = task() => {
+                    if let Err(e) = result {
+                        tracing::error!("Error while running Rapina background job: {:?}", e);
+                    }
```
arferreira, line 91: Does `e` implements `DIsplay`? Perhaps just `{}` gives the cleaner readable msg?

rapina/src/cron_scheduler.rs
```diff
@@ -0,0 +1,341 @@
...
+        // instantly trigger all tokio::select! branches that watch the token cancellation, stopping the execution immediately in case no blocking code is being executed
+        self.cancellation_token.cancel();
+
+        if let Some(mut scheduler) = self.scheduler.take() {
+            scheduler
+                .shutdown()
+                .await
+                .expect("Failed to shutdown Rapina background jobs");
+        }
+        tracing::info!("Shutdown signal received, stopping cron scheduler");
```
arferreira: This runs after shutdown is already complete, right?
```suggestion
        tracing::info!("Cron scheduler stopped");
```

rapina/src/cron_scheduler.rs
```diff
@@ -0,0 +1,344 @@
...
+        // instantly trigger all tokio::select! branches that watch the token cancellation, stopping the execution immediately in case no blocking code is being executed
+        self.cancellation_token.cancel();
+
+        if let Some(mut scheduler) = self.scheduler.take() {
+            scheduler
+                .shutdown()
+                .await
+                .expect("Failed to shutdown Rapina background jobs");
+        }
+        tracing::info!("Cron scheduler stopped");
```
juv:
```suggestion
        tracing::info!(
            "Cron scheduler stopped. Blocking tasks may still be in-flight and will keep the process alive until complete."
        );
```

Now that I read it again, I have the feeling we need to be more explanatory on this log message.
The cron scheduler indeed stopped. However any cron tasks that are still executing will block the Rapina process until they are finished.

What do you think?

arferreira: This looks better!

PR #457, opened 2026-04-03
feat: add Tower compatibility layer with bidirectional adapters

### LINKED ISSUE #328: Tower compatibility layer
rapina has its own middleware trait, which is intentionally simple. But the tower ecosystem has retry, load balancing, circuit breakers, hedging, rate limiting, timeouts, and concurrency limits that are all battle-tested and production-hardened. tonic, reqwest, and every serious Rust HTTP tool speaks tower. Right now rapina users can't use any of it.

The goal isn't replacing rapina's middleware with tower internally. It's providing adapters in both directions so the ecosystems can interop.

**rapina middleware as tower::Service.** Wrap a rapina handler plus its middleware stack into something that implements `tower::Service<Request<Incoming>>`. This lets rapina handlers sit behind tower's `ServiceBuilder` and use its combinators.

**tower::Service as rapina middleware.** Wrap a `tower::Service` into something that implements rapina's `Middleware` trait. This lets users drop tower layers like `tower::timeout::Timeout` or `tower::limit::ConcurrencyLimit` into rapina's middleware chain.

This should be an optional feature gate (`tower-compat` or similar) so users who don't need it pay nothing.

josemoura212:
## Summary

Adds optional Tower ecosystem interop behind the `tower` feature flag:

- **`TowerLayerMiddleware<L>`** — wraps any `tower::Layer` for use as rapina `Middleware`. Supports layers requiring `Clone` on the inner service (retry, circuit breaker, tower-resilience, etc.) via owned `NextService` without lifetime parameters.
- **`RapinaService`** — wraps rapina's middleware + router stack as a `tower::Service<Request<Incoming>>`, enabling rapina apps to sit behind `ServiceBuilder` and tower combinators.
- **`.layer()`** builder method — convenience shorthand matching the axum/hyper/tonic convention.

### Design decisions

- `NextService` owns its data (`Vec<Arc<dyn Middleware>>`, `Arc<Router>`, `Arc<AppState>`) instead of borrowing with lifetimes. This avoids higher-ranked trait bounds (HRTB) and makes it compatible with any Tower layer.
- `Next<'a>` stores `Arc<Router>` and `Arc<AppState>` instead of references, so `NextService` can be constructed from `Next` without lifetime issues. The cost is an `Arc::clone` per middleware step (atomic increment, negligible).
- Errors from Tower services are logged via `tracing::error!` and converted to generic 500 responses — no internal details leak to clients.

### Tested with

| Crate | Layer | 
|-------|-------|
| `tower` 0.5 | `TimeoutLayer` |
| `tower-resilience` 0.9 | `TimeLimiterLayer`, `BulkheadLayer` |
| `tower-http` 0.6 | `SetResponseHeaderLayer` |

Plus custom hand-written Tower layers (`SecurityHeadersLayer`, `ResponseTimeLayer`).

### Feature gating

```toml
rapina = { version = "0.11.0", features = ["tower"] }
```

When disabled, zero impact — no deps pulled, no code compiled.

Closes #328

## Test plan

- [x] `cargo fmt --check` — clean
- [x] `cargo clippy --features tower` — 0 warnings
- [x] `cargo check --no-default-features` — OK
- [x] `cargo check --all-features` — OK
- [x] `cargo test --features tower` — 372+ tests pass, 0 failures
- [x] External project test with tower, tower-resilience, tower-http layers

### DIFF
--- a/rapina/src/app.rs
+++ b/rapina/src/app.rs
@@ -190,6 +190,32 @@ impl Rapina {
         self
     }
 
+    /// Adds a tower layer as middleware.
+    ///
+    /// This is a convenience method equivalent to
+    /// `.middleware(TowerLayerMiddleware::new(layer))`.
+    ///
+    /// Requires the `tower` feature.
+    ///
+    /// # Example
+    ///
+    /// ```ignore
+    /// use rapina::prelude::*;
+    ///
+    /// Rapina::new()
+    ///     .layer(my_tower_layer)
+    ///     .listen("127.0.0.1:3000")
+    ///     .await
+    /// ```
+    #[cfg(feature = "tower")]
+    pub fn layer<L>(self, layer: L) -> Self
+    where
+        L: tower_layer::Layer<crate::middleware::NextService>,
+        crate::middleware::TowerLayerMiddleware<L>: Middleware,
+    {
+        self.middleware(crate::middleware::TowerLayerMiddleware::new(layer))
+    }
+
     /// Enables CORS for the application.
     ///
     /// Use `CorsConfig::permisive()` for development (it allows all origins),
--- a/rapina/src/middleware/tower.rs
+++ b/rapina/src/middleware/tower.rs
@@ -1,0 +1,324 @@
+//! Tower compatibility adapters for Rapina middleware.
+//!
+//! Provides two adapters for interop between Rapina's middleware system
+//! and the Tower ecosystem:
+//!
+//! - [`TowerLayerMiddleware`] wraps a `tower::Layer` for use as a Rapina
+//!   [`Middleware`](super::Middleware).
+//! - [`RapinaService`] wraps Rapina's middleware + router stack as a
+//!   `tower::Service`.
+
+use std::convert::Infallible;
+use std::fmt;
+use std::future::Future;
+use std::pin::Pin;
+use std::sync::{Arc, OnceLock};
+use std::task::{Context, Poll};
+
+use hyper::body::Incoming;
+use hyper::{Request, Response};
+use tower_layer::Layer;
+use tower_service::Service;
+
+use crate::context::RequestContext;
+use crate::middleware::{BoxFuture, Middleware, MiddlewareStack, Next};
+use crate::response::{BoxBody, IntoResponse};
+use crate::router::Router;
+use crate::state::AppState;
+
+// ─── Direction A: Tower Layer → Rapina Middleware ────────────────────────────
+
+/// A [`tower::Service`] that delegates to the Rapina middleware chain.
+///
+/// Owns all data needed to execute the chain, with no lifetime parameter.
+/// This makes it compatible with any Tower layer — including those requiring
+/// `Clone` (e.g. tower-resilience, retry, circuit breaker).
+#[derive(Clone)]
+pub struct NextService {
+    middlewares: Arc<[Arc<dyn Middleware>]>,
+    router: Arc<Router>,
+    state: Arc<AppState>,
+}
+
+impl Service<Request<Incoming>> for NextService {
+    type Response = Response<BoxBody>;
+    type Error = Infallible;
+    type Future = Pin<Box<dyn Future<Output = Result<Response<BoxBody>, Infallible>> + Send>>;
+
+    fn poll_ready(&mut self, _cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+        Poll::Ready(Ok(()))
+    }
+
+    fn call(&mut self, req: Request<Incoming>) -> Self::Future {
+        let middlewares = self.middlewares.clone();
+        let router = self.router.clone();
+        let state = self.state.clone();
+
+        Box::pin(async move {
+            let ctx = req
+                .extensions()
+                .get::<RequestContext>()
+                .cloned()
+                .unwrap_or_default();
+            let next = Next::new(&middlewares, router, state, &ctx);
+            Ok(next.run(req).await)
+        })
+    }
+}
+
+/// Wraps a [`tower::Layer`] so it can be added to a Rapina middleware stack.
+///
+/// The tower service is built once on the first request and cached. Each
+/// subsequent request clones the cached service — since tower layers store
+/// shared state behind `Arc` internally, all requests share the same state.
+/// This means stateful layers like rate limiters, circuit breakers, and
+/// bulkheads work correctly across concurrent requests.
+///
+/// # Body type
+///
+/// Tower layers that preserve the response body type (`Response<BoxBody>`)
+/// work directly. Layers that change the body type (e.g. tower-http
+/// compression) are not compatible without an additional body adapter.
+///
+/// # Example
+///
+/// ```ignore
+/// use rapina::middleware::TowerLayerMiddleware;
+///
+/// // Any tower Layer that accepts NextService works:
+/// Rapina::new()
+///     .middleware(TowerLayerMiddleware::new(my_tower_layer))
+///     .listen("127.0.0.1:3000")
+///     .await
+/// ```
+pub struct TowerLayerMiddleware<L: Layer<NextService>> {
+    layer: L,
+    service: OnceLock<L::Service>,
+}
+
+impl<L: Layer<NextService>> TowerLayerMiddleware<L> {
+    pub fn new(layer: L) -> Self {
+        Self {
+            layer,
+            service: OnceLock::new(),
+        }
+    }
+}
+
+impl<L> fmt::Debug for TowerLayerMiddleware<L>
+where
+    L: Layer<NextService> + fmt::Debug,
+{
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("TowerLayerMiddleware")
+            .field("layer", &self.layer)
+            .finish()
+    }
+}
+
+impl<L> Middleware for TowerLayerMiddleware<L>
+where
+    L: Layer<NextService> + Send + Sync + 'static,
+    <L as Layer<NextService>>::Service:
+        Service<Request<Incoming>, Response = Response<BoxBody>> + Clone + Send + Sync + 'static,
+    <<L as Layer<NextService>>::Service as Service<Request<Incoming>>>::Error:
+        fmt::Display + Send + 'static,
+    <<L as Layer<NextService>>::Service as Service<Request<Incoming>>>::Future: Send + 'static,
+{
+    fn handle<'a>(
+        &'a self,
+        mut req: Request<Incoming>,
+        ctx: &'a RequestContext,
+        next: Next<'a>,
+    ) -> BoxFuture<'a, Response<BoxBody>> {
+        Box::pin(async move {
+            req.extensions_mut().insert(ctx.clone());
+            let template = self.service.get_or_init(|| {
+                let next_svc = NextService {
+                    middlewares: next.middlewares.iter().cloned().collect(),
+                    router: next.router.clone(),
+                    state: next.state.clone(),
+                };
+                self.layer.layer(next_svc)
+            });
+            let mut svc = template.clone();
+
+            if let Err(e) = std::future::poll_fn(|cx| svc.poll_ready(cx)).await {
+                tracing::error!("tower service not ready: {}", e);
+                return crate::error::Error::service_unavailable("service unavailable")
+                    .into_response();
+            }
+
+            match svc.call(req).await {
+                Ok(response) => response,
+                Err(e) => {
+                    tracing::error!("tower service error: {}", e);
+                    crate::error::Error::internal("internal server error").into_response()
+                }
+            }
+        })
+    }
+}
+
+// ─── Direction B: Rapina Stack → Tower Service ──────────────────────────────
+
+/// Wraps Rapina's middleware + router stack as a [`tower::Service`].
+///
+/// This allows embedding a fully-configured Rapina application inside
+/// tower-based infrastructure or using tower testing utilities.
+///
+/// The service is always ready (`poll_ready` returns `Poll::Ready(Ok(()))`)
+/// and never errors (`Error = Infallible`), matching Rapina's infallible
+/// response model.
+///
+/// # Example
+///
+/// ```ignore
+/// use rapina::middleware::RapinaService;
+///
+/// let service = RapinaService::new(router, state, middlewares);
+/// // `service` implements tower::Service<Request<Incoming>>
+/// ```
+#[derive(Clone)]
+pub struct RapinaService {
+    router: Arc<Router>,
+    state: Arc<AppState>,
+    middlewares: Arc<MiddlewareStack>,
+}
+
+impl RapinaService {
+    pub fn new(mut router: Router, state: AppState, middlewares: MiddlewareStack) -> Self {
+        router.freeze();
+        Self {
+            router: Arc::new(router),
+            state: Arc::new(state),
+            middlewares: Arc::new(middlewares),
+        }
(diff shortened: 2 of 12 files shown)


arferreira:
Hey @josemoura212 you don't need to request copilot review, we personally review all PRs

rapina/src/middleware/tower.rs
```diff
@@ -0,0 +1,287 @@
...
+                middlewares: next.middlewares.to_vec(),
+                router: next.router.clone(),
+                state: next.state.clone(),
+                ctx: ctx.clone(),
+            };
+            let mut svc = self.layer.layer(next_svc);
+
+            if let Err(e) = std::future::poll_fn(|cx| svc.poll_ready(cx)).await {
+                tracing::error!("tower service not ready: {}", e);
+                return crate::error::Error::internal("service unavailable").into_response();
```
ShiraiEd: This says `"service unavailable"` but returns an `Error 500`,
the error and the message are not matching, try changing to 
```rust
crate::error::Error::ServiceUnavailable
```
instead

rapina/src/middleware/tower.rs
```diff
@@ -0,0 +1,288 @@
...
+//! - [`TowerLayerMiddleware`] wraps a `tower::Layer` for use as a Rapina
+//!   [`Middleware`](super::Middleware).
+//! - [`RapinaService`] wraps Rapina's middleware + router stack as a
+//!   `tower::Service`.
+
+use std::convert::Infallible;
+use std::fmt;
+use std::future::Future;
+use std::pin::Pin;
+use std::sync::Arc;
```
ShiraiEd: import `OnceLock` to use in TowerLayerMiddleware

rapina/src/middleware/tower.rs
```diff
@@ -0,0 +1,288 @@
...
+/// Owns all data needed to execute the chain, with no lifetime parameter.
+/// This makes it compatible with any Tower layer — including those requiring
+/// `Clone` (e.g. tower-resilience, retry, circuit breaker).
+#[derive(Clone)]
+pub(crate) struct NextService {
+    middlewares: Vec<Arc<dyn Middleware>>,
+    router: Arc<Router>,
+    state: Arc<AppState>,
+    ctx: RequestContext,
+}
```
ShiraiEd: `Vec` → `Arc<[T]>`
Vec` owns its data and allocating a new one costs a heap allocation + copying every pointer. `Arc<[T]>` is a reference-counted slice, cloning it is a single atomic increment regardless of how many middlewares there are.

Before: every request paid this cost. After: paid once on first request, then free.


Removing `ctx`

`ctx` contains the trace ID and start time different for every request. As long as `NextService` holds it, the struct **must** be rebuilt per request because you can't reuse a service that has last request's trace ID baked into it.

Remove `ctx` and `NextService` becomes identical for every request at the same chain position, same remaining middlewares, same router, same state. Now it's safe to build once and reuse.

`ctx` doesn't disappear, it just moves to `req.extensions()` where it already was anyway.

rapina/src/middleware/tower.rs
```diff
@@ -0,0 +1,288 @@
...
+
+    fn poll_ready(&mut self, _cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+        Poll::Ready(Ok(()))
+    }
+
+    fn call(&mut self, req: Request<Incoming>) -> Self::Future {
+        let middlewares = self.middlewares.clone();
+        let router = self.router.clone();
+        let state = self.state.clone();
+        let ctx = self.ctx.clone();
```
ShiraiEd: remove this `ctx`

rapina/src/middleware/tower.rs
```diff
@@ -0,0 +1,288 @@
...
+
+        Box::pin(async move {
```
ShiraiEd, line 57: add the ctx here 
```rust
 let ctx = req
      .extensions()
      .get::<RequestContext>()
      .cloned()
      .unwrap_or_default();
```

rapina/src/middleware/tower.rs
```diff
@@ -0,0 +1,288 @@
...
+/// ```ignore
+/// use rapina::middleware::TowerLayerMiddleware;
+///
+/// // Any tower Layer that accepts NextService works:
+/// Rapina::new()
+///     .middleware(TowerLayerMiddleware::new(my_tower_layer))
+///     .listen("127.0.0.1:3000")
+///     .await
+/// ```
+pub struct TowerLayerMiddleware<L> {
```
ShiraiEd: use `OnceLock` here
```rust
where
    L: Layer<NextService>,
{
    layer: L,
    service: OnceLock<L::Service>,
```

rapina/src/middleware/tower.rs
```diff
@@ -0,0 +1,288 @@
...
+    layer: L,
+}
+
+impl<L> TowerLayerMiddleware<L> {
```
ShiraiEd: and `NextLayer` here
```rust
impl<L> TowerLayerMiddleware<L>
where
    L: Layer<NextService>,
```

rapina/src/middleware/tower.rs
```diff
@@ -0,0 +1,288 @@
...
+    pub fn new(layer: L) -> Self {
+        Self { layer }
```
ShiraiEd:
```rust
Self {
            layer,
            service: OnceLock::new(),
        }
```

rapina/src/middleware/tower.rs
```diff
@@ -0,0 +1,288 @@
...
+            .field("layer", &self.layer)
+            .finish()
+    }
+}
+
+impl<L> Middleware for TowerLayerMiddleware<L>
+where
+    L: Layer<NextService> + Send + Sync + 'static,
+    <L as Layer<NextService>>::Service:
+        Service<Request<Incoming>, Response = Response<BoxBody>> + Send + 'static,
```
ShiraiEd: Service<Request<Incoming>, Response = Response<BoxBody>> + Clone + Send + Sync + 'static,

rapina/src/middleware/tower.rs
```diff
@@ -0,0 +1,288 @@
...
+    ) -> BoxFuture<'a, Response<BoxBody>> {
+        Box::pin(async move {
+            req.extensions_mut().insert(ctx.clone());
+            let next_svc = NextService {
+                middlewares: next.middlewares.to_vec(),
+                router: next.router.clone(),
+                state: next.state.clone(),
+                ctx: ctx.clone(),
+            };
+            let mut svc = self.layer.layer(next_svc);
```
ShiraiEd:
```rust
let template = self.service.get_or_init(|| {
                let next_svc = NextService {
                    middlewares: next.middlewares.iter().cloned().collect(),
                    router: next.router.clone(),
                    state: next.state.clone(),
                };
                self.layer.layer(next_svc)
            });
            let mut svc = template.clone();
```

rapina/src/middleware/tower.rs
```diff
@@ -0,0 +1,288 @@
...
+    layer: L,
+}
+
+impl<L> TowerLayerMiddleware<L> {
+    pub fn new(layer: L) -> Self {
+        Self { layer }
+    }
+}
+
+impl<L: fmt::Debug> fmt::Debug for TowerLayerMiddleware<L> {
```
ShiraiEd: where
    L: Layer<NextService> + fmt::Debug,

rapina/src/middleware/tower.rs
```diff
@@ -0,0 +1,288 @@
...
+                middlewares: next.middlewares.to_vec(),
+                router: next.router.clone(),
+                state: next.state.clone(),
+                ctx: ctx.clone(),
+            };
+            let mut svc = self.layer.layer(next_svc);
+
+            if let Err(e) = std::future::poll_fn(|cx| svc.poll_ready(cx)).await {
+                tracing::error!("tower service not ready: {}", e);
+                return crate::error::Error::service_unavailable("service unavailable")
```
ShiraiEd, line 148: Sorry I have just re-checked, the service_unavailable error still doens't exist in the rapina/src/error.rs

use
`// error.rs:180
crate::error::Error::new(503, "SERVICE_UNAVAILABLE", "service unavailable")
`
for now,

my bad man

ShiraiEd (changes requested):
Hello Jose, good work so far!

With this implementation, for stateful layers like `rate limiter`,each new request creates a new service, so in the end if the dev sets the limit for like 10 requests, each one gonna create its own counter and the limit will never be reached,

so because of this we will have to change somethings on the code,

like the `NextService` struct, the `call` fn, the `TowerLayerMiddleware` struct and its impls.

Also,  `test_rapina_service_routes_requests` doesn't actually use
`RapinaService`, it goes through the normal app builder and duplicates
existing routing tests. Worth replacing with a test that adds `tower` as
a dev-dependency and uses a real stateful layer like `ConcurrencyLimitLayer`
to verify the shared state actually works across concurrent requests.

rapina/src/middleware/tower.rs
```diff
@@ -0,0 +1,306 @@
...
+            Ok(next.run(req).await)
+        })
+    }
+}
+
+/// Wraps a [`tower::Layer`] so it can be added to a Rapina middleware stack.
+///
+/// On each request the layer is applied to a fresh [`NextService`] that
+/// delegates to the remaining Rapina middleware chain. The resulting tower
+/// service is then called to process the request.
```
ShiraiEd: This comment is outdated,
change it into something like 
```
/// Wraps a [`tower::Layer`] so it can be added to a Rapina middleware stack.
///
/// The tower service is built once on the first request and cached. Each
/// subsequent request clones the cached service — since tower layers store
/// shared state behind `Arc` internally, all requests share the same state.
/// This means stateful layers like rate limiters, circuit breakers, and
/// bulkheads work correctly across concurrent requests.
```

rapina/src/middleware/tower.rs
```diff
@@ -0,0 +1,306 @@
...
+
+// ─── Direction A: Tower Layer → Rapina Middleware ────────────────────────────
+
+/// A [`tower::Service`] that delegates to the Rapina middleware chain.
+///
+/// Owns all data needed to execute the chain, with no lifetime parameter.
+/// This makes it compatible with any Tower layer — including those requiring
+/// `Clone` (e.g. tower-resilience, retry, circuit breaker).
+#[derive(Clone)]
+pub struct NextService {
```
ShiraiEd, line 37: gate this behind `pub (crate)` so users can't see/use directly

rapina/src/middleware/tower.rs
```diff
@@ -0,0 +1,306 @@
...
+
+        let app = Rapina::new()
+            .with_introspection(false)
+            .layer(ConcurrencyLimitLayer::new(2))
+            .router(Router::new().route(http::Method::GET, "/ping", |_, _, _| async { "pong" }));
+
+        let client = TestClient::new(app).await;
+
+        let r1 = client.get("/ping").send().await;
+        let r2 = client.get("/ping").send().await;
```
ShiraiEd: test_concurrency_limit_shared_state` is sequential, not concurrent

```rust
let r1 = client.get("/ping").send().await;  // finishes first
let r2 = client.get("/ping").send().await;  // then this starts
```

The limit is never under pressure. Both requests pass whether state is shared or not. To actually test the limit, requests need to overlap using `tokio::join!` with a slow handler so they're in-flight at the same time.

ShiraiEd (changes requested):
Good,we are almost finished
just some little adjusts on the tower.rs

also , the tower tests are only unit tests inside `tower.rs`. 
because this is a new feature that interacts with the existing middleware structure It would be good to add integration tests in `tests/middleware.rs` alongside the existing middleware tests.

A good integration test would:
- Register a tower layer alongside a regular rapina middleware and verify both run and ordering is preserved
- Test `RapinaService` directly by constructing it with `RapinaService::new()` and calling it with a real request

`test_concurrency_limit_shared_state` in tower.rs sends requests sequentially so the concurrency limit is never under pressure — both requests pass whether state is shared or not. To actually test shared state, requests need to overlap using `tokio::join!` with a slow handler so they are in-flight at the same time.

rapina/src/middleware/tower.rs
arferreira: I guess the path to fix this without a larger refactor is:

Remove `NextService` from the re-export in `middleware/mod`:

```rust
pub use tower::{RapinaService, TowerLayerMiddleware};
```

And remove the redundant bound from `.layer()` in `app`:

```rust
pub fn layer<L>(self, layer: L) -> Self
       where
           crate::middleware::TowerLayerMiddleware<L>: Middleware,
       {
```

The `TowerLayerMiddleware<L>: Middleware` impl already requires `L: Layer<NextService>` transitively, so the explicit bound is redundant. With this change, users never need to name NextService. It lives in the private `tower` module and is not reachable from outside the crate.

rapina/src/app.rs
```diff
@@ -188,6 +183,32 @@ impl Rapina {
...
+    ///
+    /// Rapina::new()
+    ///     .layer(my_tower_layer)
+    ///     .listen("127.0.0.1:3000")
+    ///     .await
+    /// ```
+    #[cfg(feature = "tower")]
+    pub fn layer<L>(self, layer: L) -> Self
+    where
+        L: tower_layer::Layer<crate::middleware::NextService>,
```
ShiraiEd, line 213: Delete this line, as referenced here: https://github.com/rapina-rs/rapina/pull/457/changes#r3039228453
Do the final `cargo fmt --all`, `cargo clippy` and `cargo test`,`cargo test --features tower`
resolve the conflicts in Cargo.lock
Squash all the commits into 1 with `git rebase -i HEAD~(number of commits, probably 8 by them)`

and to me we are good to merge

ShiraiEd (changes requested):

rapina/src/app.rs
ShiraiEd: Hey @josemoura212,
how is this going?

josemoura212: Hey @ShiraiEd, sorry for the delay — been swamped with work. Pushing the fixes now!

josemoura212: I looked into the two remaining items:

 1. NextService visibility — it's already pub only inside the private mod tower, and re-exported as  pub(crate) in mod.rs. Making the struct itself pub(crate) triggers private_bounds warnings because TowerLayerMiddleware (which is pub) references it in its generic bound. Since the module is private,  NextService is already invisible to downstream users.

2. Removing the Layer<NextService> bound from .layer() — this doesn't compile. The struct definition TowerLayerMiddleware<L: Layer<NextService>> uses L::Service as a field type, so the bound is          
structurally required, not redundant. Removing it from the struct would require a larger refactor.
                                                                                                     
So the code is already in the correct state. Happy to do the cargo fmt --all / cargo clippy / cargo test pass and squash if you agree these are fine as-is!

rapina/src/middleware/tower.rs
ShiraiEd: it doesn't compile because the struct itself requires `L: Layer<NextService>` to use `L::Service` as a field type

rapina/src/app.rs
ShiraiEd: Understood man 👍,
You can proceed with the checks and the squash pls

ShiraiEd: This pr sure was a long one 😅

ShiraiEd:
Sorry,
but the Cargo.lock has conflicts,
could you resolve that then squash it and push again?

ShiraiEd (approved):
LGTM

arferreira (approved):
Thanks for the contribution, LGTM.

PR #470, opened 2026-04-06
Add unit tests for timeout and body_limit middleware handle methods

### LINKED ISSUE #213: Add unit tests for timeout and body_limit middleware handle methods
The timeout and body_limit middleware have tests for their constructors but the actual `handle` methods are untested.

**Timeout (`rapina/src/middleware/timeout.rs`):**
- Verify a fast handler completes normally
- Verify a slow handler returns a timeout error

**Body limit (`rapina/src/middleware/body_limit.rs`):**
- Verify a request under the limit passes through
- Verify a request over the limit is rejected
- Verify a request with no Content-Length header behavior

Look at `rapina/src/middleware/rate_limit.rs` tests for patterns on how to construct requests and test middleware.

Ismaellima4:
## Summary

What does this PR do?

Add unit tests for middleware constructors and configuration:
- TimeoutMiddleware: test new(), default(), milliseconds, zero duration
- BodyLimitMiddleware: test new(), default(), custom size, zero size

Add integration tests for middleware handle methods:
- Timeout middleware rejects slow handlers with 500
- Body limit middleware rejects large bodies with 400
- Body limit middleware allows requests without Content-Length

## Related Issues

Closes #213 

## Checklist

- [x] `cargo fmt` passes
- [x] Tests pass

### DIFF
--- a/rapina/src/middleware/body_limit.rs
+++ b/rapina/src/middleware/body_limit.rs
@@ -61,32 +61,3 @@ impl Middleware for BodyLimitMiddleware {
         })
     }
 }
-
-#[cfg(test)]
-mod tests {
-    use super::*;
-
-    #[test]
-    fn test_body_limit_middleware_new() {
-        let mw = BodyLimitMiddleware::new(2048);
-        assert_eq!(mw.max_size, 2048);
-    }
-
-    #[test]
-    fn test_body_limit_middleware_default() {
-        let mw = BodyLimitMiddleware::default();
-        assert_eq!(mw.max_size, 1024 * 1024); // 1MB default
-    }
-
-    #[test]
-    fn test_body_limit_custom_size() {
-        let mw = BodyLimitMiddleware::new(5 * 1024 * 1024); // 5MB
-        assert_eq!(mw.max_size, 5 * 1024 * 1024);
-    }
-
-    #[test]
-    fn test_body_limit_zero_size() {
-        let mw = BodyLimitMiddleware::new(0);
-        assert_eq!(mw.max_size, 0);
-    }
-}
--- a/rapina/src/middleware/body_limit.rs
+++ b/rapina/src/middleware/body_limit.rs
@@ -61,3 +61,32 @@ impl Middleware for BodyLimitMiddleware {
         })
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_body_limit_middleware_new() {
+        let mw = BodyLimitMiddleware::new(2048);
+        assert_eq!(mw.max_size, 2048);
+    }
+
+    #[test]
+    fn test_body_limit_middleware_default() {
+        let mw = BodyLimitMiddleware::default();
+        assert_eq!(mw.max_size, 1024 * 1024); // 1MB default
+    }
+
+    #[test]
+    fn test_body_limit_custom_size() {
+        let mw = BodyLimitMiddleware::new(5 * 1024 * 1024); // 5MB
+        assert_eq!(mw.max_size, 5 * 1024 * 1024);
+    }
+
+    #[test]
+    fn test_body_limit_zero_size() {
+        let mw = BodyLimitMiddleware::new(0);
+        assert_eq!(mw.max_size, 0);
+    }
+}
--- a/rapina/src/middleware/timeout.rs
+++ b/rapina/src/middleware/timeout.rs
@@ -12,7 +12,7 @@ use super::{BoxFuture, Middleware, Next};
 /// Middleware that enforces a maximum duration for each request.
 ///
 /// If a handler does not respond within the configured duration the request is
-/// cancelled and a `500 Internal Server Error` is returned to the client.
+/// cancelled and a `408 Request Timeout` is returned to the client.
 /// Defaults to 30 seconds.
 ///
 /// # Example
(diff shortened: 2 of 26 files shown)


rapina/src/middleware/body_limit.rs
```diff
@@ -61,3 +61,32 @@ impl Middleware for BodyLimitMiddleware {
...
+        let mw = BodyLimitMiddleware::new(5 * 1024 * 1024); // 5MB
+        assert_eq!(mw.max_size, 5 * 1024 * 1024);
+    }
+
+    #[test]
+    fn test_body_limit_zero_size() {
+        let mw = BodyLimitMiddleware::new(0);
+        assert_eq!(mw.max_size, 0);
+    }
+}
```
arferreira: These four tests assert that struct fields hold the values passed to the constructor. That's testing Rust assignment, not middleware behavior. A test that exercises `handle()` directly would be more meaningful here. I'd drop these and rely on the integration tests in tests/middleware.rs to cover correctness.

rapina/src/middleware/timeout.rs
```diff
@@ -53,3 +53,33 @@ impl Middleware for TimeoutMiddleware {
         })
     }
 }
+
```
arferreira: Same concern as the `BodyLimitMiddleware` tests. `assert_eq!(mw.duration, Duration::from_secs(60))` after `new(Duration::from_secs(60))` cannot reveal a bug. The integration test below is what actually validates behavior.

rapina/tests/middleware.rs
```diff
@@ -557,3 +557,67 @@ async fn test_trace_id_middleware_preserves_incoming_trace_id() {
...
+            Router::new().route(http::Method::POST, "/upload", |req, _, _| async move {
+                use http_body_util::BodyExt;
+                let body = req.into_body().collect().await.unwrap().to_bytes();
+                format!("Received {} bytes", body.len())
+            }),
+        );
+
+    let client = TestClient::new(app).await;
+    let large_body: &'static str = "x".repeat(200).leak(); // 200 bytes, over the 100 limit
+    let response = client.post("/upload").body(large_body).send().await;
```
arferreira: `String::leak()` is intentional memory leakage to coerce a `String` into `&'static str`. No need for it here, just pass the `String` directly.
```suggestion
    let large_body = "x".repeat(200);
    let response = client.post("/upload").body(large_body).send().await;
```

arferreira (changes requested):
Thanks for the contribution! The integration tests are a solid addition and exactly what was missing. Just left a few things to address before merging.

rapina/tests/middleware.rs
```diff
@@ -557,3 +557,67 @@ async fn test_trace_id_middleware_preserves_incoming_trace_id() {
...
+            Router::new().route(http::Method::GET, "/slow", |_, _, _| async {
+                tokio::time::sleep(Duration::from_millis(200)).await;
+                "should not reach"
+            }),
+        );
+
+    let client = TestClient::new(app).await;
+    let response = client.get("/slow").send().await;
+
+    assert_eq!(response.status(), StatusCode::INTERNAL_SERVER_ERROR);
```
arferreira: A timeout is a client-visible condition and HTTP 408 Request Timeout exists for exactly this case. Is there a reason this returns 500 instead?
Either fix the middleware or leave a comment explaining the deliberate choice, but asserting 500 without context will confuse anyone reading this test later.

rapina/tests/middleware.rs
```diff
@@ -557,3 +557,67 @@ async fn test_trace_id_middleware_preserves_incoming_trace_id() {
...
+        );
+
+    let client = TestClient::new(app).await;
+    // Send a body larger than the limit but without explicit Content-Length
+    // The middleware should pass it through since it can't determine the size
+    let response = client
+        .post("/upload")
+        .body("small body under limit")
+        .send()
+        .await;
```
arferreira: The body here is "small body under limit" at 22 bytes, which is under the 100-byte cap. This test passes whether or not the Content-Length behavior is correct since the request would be allowed even if Content-Length were present and checked. To actually verify what the test name promises, send a body over 100 bytes. The only reason it should pass then is the absence of Content-Length.
```suggestion
    // 200 bytes: if Content-Length were present the middleware would reject this.
    // Without it the middleware has no size information and must pass through.
    let response = client
        .post("/upload")
        .body("x".repeat(200))
        .send()
        .await;
```

arferreira (changes requested):
Good progress on the fixes. Two things left: the timeout returning 500 needs either a fix or an explanation, and the no-content-length test body needs to be over the limit to actually exercise the behavior you described.

rapina/tests/middleware.rs
Ismaellima4: The current implementation returns a 500 status code. I'll update this in the middleware.

https://github.com/rapina-rs/rapina/blob/85a96cf7043aa9c28d1510dfbe551b1423f6cdf8/rapina/src/middleware/timeout.rs#L48-L53

arferreira (approved):
Good work addressing everything. One minor thing: `test_body_limit_middleware_allows_no_content_length` is misleading since hyper auto-injects Content-Length and the test actually asserts BAD_REQUEST. Not a blocker overall. We can clean that up in a follow-up.

PR #473, opened 2026-04-08
Add --database subcommand to rapina new command 

### LINKED ISSUE #371: [Feature]:  add `--database` subcommand to `rapina new` command
### Problem Statement

Currently, when initializing a new project with the **Rapina** CLI, users must manually edit the `Cargo.toml` file to enable specific database features (eg, `sqlite`, `postgres`, `mysql`) and add the corresponding dependencies.

### Proposed Solution

Introduce a `--db` (or `--database`) optional flag to the `new` commands. This allows the user to define their persistence layer at the moment of creation.

### Key Functionalities
1.  **Automatic Dependency Injection:** The CLI will automatically add the necessary crates to `Cargo.toml` with the correct driver features enabled.
2.  **Input Validation:** The CLI will only accept supported values (e.g., `sqlite`, `postgres`, `mysql`) and provide clear error messages for unsupported ones.
---

## Technical Example (Usage)

To initialize a new project with **SQLite** support:

```bash
rapina new my_app --db=sqlite
```
## Expected `Cargo.toml` output snippet:

```toml
[package]
name = "my_app"
version = "0.1.0"
edition = "2024"

[dependencies]
rapina = { version = "0.10.0", features = ["sqlite"] }
tokio = { version = "1", features = ["full"] }
serde = { version = "1", features = ["derive"] }
serde_json = "1"
hyper = "1"
```

### Contribution

Ismaellima4:
## Summary

What does this PR do?

This PR introduces comprehensive database configuration support for the `rapina new` command, enabling users to scaffold projects with or without database integration based on the new `--db` flag. It refactors the template generation system to conditionally inject database-related code, centralizes environment variable generation, and improves CLI validation.

**Key Changes:**
- **CLI & Validation:**
  - Added `--db <sqlite|postgres|mysql>` flag with strict value parsing.
  - Enforced `--db` requirement for `--template crud` using Clap's `requires_if` validation.
  - Moved template & requirement validation to run **before** any filesystem operations, preventing partial directory creation on failure.

- **Conditional Template Generation:**
  - Database code (`.with_database()`, `load_dotenv()`, `DatabaseConfig` imports) is now only generated when `--db` is explicitly provided.

- **Environment Configuration:**
  - Replaced `.env.example` with direct `.env` generation containing clear placeholder values and English replacement warnings (`⚠️ Replace the values below...`).
  - Delegated `.env` creation responsibility to each individual template (`rest_api`, `auth`, `crud`) for better separation of concerns.

- **Architecture & Safety:**
  - Refactored CRUD template to require `&DatabaseType` explicitly, as CRUD scaffolds inherently need a database.
  - Ensured type-safe optional handling across all template generators.

## Related Issues

Closes #371 

## Checklist

- [x] `cargo fmt` passes
- [x] Tests pass

### DIFF
--- a/rapina-cli/src/commands/templates/mod.rs
+++ b/rapina-cli/src/commands/templates/mod.rs
@@ -109,7 +109,7 @@ pub fn generate_db_import(db_type: Option<&DatabaseType>) -> &'static str {
 /// Uses `DatabaseConfig::from_env()` for all database types since `.env` is auto-generated.
 /// Returns an empty string if no database is configured.
 pub fn generate_database_config(db_type: Option<&DatabaseType>) -> &'static str {
-    db_type.map_or("", |_| "let db_config = DatabaseConfig::from_env()?;\n")
+    db_type.map_or("", |_| "let db_config = DatabaseConfig::from_env().expect(\"Failed to configure database\");\n")
 }
 
 /// Generate the `.with_database(db_config)` builder line.
@@ -150,11 +150,11 @@ pub fn generate_env_content(db_type: Option<&DatabaseType>, extra_vars: Option<&
 /// Generate the `.gitignore` extras for a given database type.
 /// Always includes `.env`. Adds `*.db` for SQLite.
 pub fn generate_gitignore_extras(db_type: Option<&DatabaseType>) -> Vec<&'static str> {
-    match db_type {
-        Some(DatabaseType::Sqlite) => vec!["*.db", ".env"],
-        Some(DatabaseType::Postgres) | Some(DatabaseType::Mysql) => vec![".env"],
-        None => vec![],
+    let mut extra = vec![".env"];
+    if let Some(DatabaseType::Sqlite) = db_type {
+        extra.push("*.db");
     }
+    extra
 }
 
 #[cfg(test)]
@@ -204,7 +204,9 @@ mod tests {
     fn test_generate_database_config_with_db() {
         let config = generate_database_config(Some(&DatabaseType::Postgres));
         assert!(config.contains("let db_config ="));
-        assert!(config.contains("DatabaseConfig::from_env()"));
+        assert!(
+            config.contains("DatabaseConfig::from_env().expect(\"Failed to configure database\");")
+        );
     }
 
     #[test]
@@ -264,9 +266,17 @@ mod tests {
     }
 
     #[test]
+    fn test_generate_gitignore_extras_mysql() {
+        let extras = generate_gitignore_extras(Some(&DatabaseType::Mysql));
+        assert!(extras.contains(&".env"));
+        assert!(!extras.contains(&"*.db"));
+    }
+
+    #[test]
     fn test_generate_gitignore_extras_none() {
         let extras = generate_gitignore_extras(None);
-        assert!(extras.is_empty());
+        assert!(extras.contains(&".env"));
+        assert!(!extras.contains(&"*.db"));
     }
 
     #[test]
--- a/rapina-cli/src/commands/templates/mod.rs
+++ b/rapina-cli/src/commands/templates/mod.rs
@@ -6,6 +6,40 @@ use colored::Colorize;
 use std::fs;
 use std::path::Path;
 
+/// Database types supported for project generation.
+#[derive(Clone, PartialEq)]
+pub enum DatabaseType {
+    Sqlite,
+    Postgres,
+    Mysql,
+}
+
+impl std::str::FromStr for DatabaseType {
+    type Err = String;
+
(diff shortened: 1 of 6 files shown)


rapina-cli/src/commands/templates/mod.rs
```diff
@@ -44,3 +77,207 @@ pub fn generate_gitignore(extras: &[&str]) -> String {
...
+    )
+}
+
+/// Generate the `.gitignore` extras for a given database type.
+/// Always includes `.env`. Adds `*.db` for SQLite.
+pub fn generate_gitignore_extras(db_type: Option<&DatabaseType>) -> Vec<&'static str> {
+    match db_type {
+        Some(DatabaseType::Sqlite) => vec!["*.db", ".env"],
+        Some(DatabaseType::Postgres) | Some(DatabaseType::Mysql) => vec![".env"],
+        None => vec![],
```
arferreira: The doc comment says "Always includes .env" but `None` arm returns an empty vec, also `auth` template always writes a `.env` file regardless of `db_type`, so scaffolding `--template auth` without `--db` produces a project where .env exists but is not gitignored. One git add . and the JWT secret is committed.
```suggestion
        None => vec![".env"],
```

rapina-cli/src/main.rs
```diff
@@ -23,9 +23,13 @@ enum Commands {
...
         name: String,
-        /// Starter template (crud, auth). Defaults to a REST API scaffold when omitted.
-        #[arg(long)]
+        /// Starter template (rest-api, crud, auth). Defaults to a REST API scaffold when omitted.
+        /// Requires `--db` when using `crud` template.
+        #[arg(long, value_parser = ["rest-api", "crud", "auth"], requires_if("crud", "db"))]
         template: Option<String>,
+        /// Database type (sqlite, postgres, mysql). Required when using `--template crud`.
+        #[arg(long, value_name = "DB")]
+        db: Option<String>,
```
arferreira: `DatabaseType` already implements `FromStr`. Clap picks that up automatically so this field can be typed directly, which removes the manual parse block and the process::exit(1) in the match arm below.
```suggestion
        db: Option<DatabaseType>,
```
You would also need to import `DatabaseType` at the top of the Commands enum or use the full path, but the parse block in main goes away entirely.

rapina-cli/src/commands/templates/rest_api.rs
```diff
@@ -13,50 +22,66 @@ pub fn generate(name: &str, project_path: &Path, src_path: &Path) -> Result<(),
...
-
+async fn main() -> std::io::Result<()> {{
+    {load_dotenv_line}
+    {db_config}
     Rapina::new()
         .with_tracing(TracingConfig::new())
         .middleware(RequestLogMiddleware::new())
         .with_health_check(true)
-        .router(router)
+{with_database_line}        .discover()
```
arferreira: Okay, I like that change for using our new feature auto discovery, but to be consistent with purpose of this PR, I would suggest for you to bring that change in a follow-up.

rapina-cli/src/commands/templates/mod.rs
```diff
@@ -44,3 +77,207 @@ pub fn generate_gitignore(extras: &[&str]) -> String {
...
+pub fn generate_db_import(db_type: Option<&DatabaseType>) -> &'static str {
+    db_type.map_or("", |_| "use rapina::database::DatabaseConfig;\n")
+}
+
+/// Generate the `let db_config = ...` line for database configuration.
+/// Uses `DatabaseConfig::from_env()` for all database types since `.env` is auto-generated.
+/// Returns an empty string if no database is configured.
+pub fn generate_database_config(db_type: Option<&DatabaseType>) -> &'static str {
+    db_type.map_or("", |_| "let db_config = DatabaseConfig::from_env()?;\n")
+}
```
arferreira, line 113: This emits `DatabaseConfig::from_env()?`. The generated code lands in a main that returns `io::Result<()>`. Whether `DatabaseConfig`'s error type implements `Into<io::Error>` needs verification, if it does not, the scaffolded project will not compile. `crud.rs` uses `.expect()` for the same call, which sidesteps this. Worth aligning and confirming the generated code actually builds.

arferreira (changes requested):
Hey, good feature, nice work so far, just left two things to fix before merging.

rapina-cli/src/commands/templates/mod.rs
Ismaellima4: I tested all possible combinations before opening the PR, and the code using `DatabaseConfig::from_env()?` was compiling fine. However, I’ll switch it to `.expect()` to keep the codebase consistent.

arferreira (approved):
LGTM.

PR #491, opened 2026-04-13
Remove docs/ from Check version consistency in ci.yml

### LINKED ISSUE #490: Update CI check to exclude the docs
Everytime we launch a new release note in the docs, the CI check blocks it due to the outdated rapina versions cited.

Update the workflow so it excludes it from the check

ShiraiEd:
## Summary

Remove "docs/" from the "Check version consistency" in the ci.tml so the new release notes doesn't get blocked by
the outdated versions documented in there

## Related Issues

Closes #490

## Checklist

- [x] `cargo fmt` passes
- [x] Tests pass

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,7 +37,7 @@ jobs:
         run: |
           VERSION=$(grep '^version' rapina/Cargo.toml | head -1 | sed 's/.*"\(.*\)"/\1/')
           MINOR="${VERSION%.*}"
-          BAD=$(grep -rn 'rapina = { version = "' docs/ rapina/examples/ \
+          BAD=$(grep -rn 'rapina = { version = "' docs/ rapina/examples/ --exclude-dir="blog" \
             --include="*.md" --include="*.toml" \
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -37,7 +37,7 @@ jobs:
         run: |
           VERSION=$(grep '^version' rapina/Cargo.toml | head -1 | sed 's/.*"\(.*\)"/\1/')
           MINOR="${VERSION%.*}"
-          BAD=$(grep -rn 'rapina = { version = "' docs/ rapina/examples/ \
+          BAD=$(grep -rn 'rapina = { version = "' rapina/examples/ \
```
arferreira: I think this is too broad, I mean removing docs entirely means guides and tutorials as well. The problem is only with blog posts. 
```suggestion
          BAD=$(grep -rn 'rapina = { version = "' docs/ rapina/examples/ --exclude-dir="blog" \
```

arferreira (commented):
I guess we shouldn't drop `docs/` altogether.

arferreira (approved):

PR #492, opened 2026-04-13
Ignore RUSTSEC-2026-0097 rand unsoundness advisory

arferreira:
The rand 0.8.5 unsoundness advisory (RUSTSEC-2026-0097, published April 9) is failing the security audit on main and blocking PRs. The advisory requires a very specific chain: custom logger + thread_rng + reseeding during logging, which rapina does not trigger.

Adds the ignore to both ci.yml and audit.yml to unblock CI.


ShiraiEd (approved):
LGTM

PR #498, opened 2026-04-15
Feat: add enable/disable wrappers methods

### LINKED ISSUE #476: [Suggestion]:  Replace `with_<feature>(bool)` pattern with explicit `enable_<feature>()` and `disable_<feature>()` methods
### Suggestion

Rename builder methods that take a `bool` parameter to use explicit `enable_*()` / `disable_*()` methods, removing the ambiguity of passing `true`/`false`.

### Context

Methods like `with_introspection(false)`, `with_health_check(true)`, and `with_metrics(true)` are confusing because:

1. **No semantic difference** — calling `.with_introspection(true)` is the same as saying "enable introspection", but the name doesn't convey intent.
2. **Default is already `false`** — for most features, the default is `false`. Passing `true` means "I want this", not "configure this to true".
3. **Can't fluently disable** — in debug builds, introspection is enabled by default (`cfg!(debug_assertions)`). To disable it in tests, you'd need `.with_introspection(false)` which reads awkwardly.
4. **Better discoverability** — `enable_introspection()` and `disable_introspection()` are self-documenting and autocomplete-friendly.

### Proposed API

| Before | After |
|---|---|
| `.with_introspection(true)` | `.enable_introspection()` |
| `.with_introspection(false)` | `.disable_introspection()` |
| `.with_health_check(true)` | `.enable_health_check()` |
| `.with_metrics(true)` | `.enable_metrics()` |

### Scope

API Design

### Before

```rust
Rapina::new()
    .router(router)
    .with_introspection(true)
    .with_health_check(true)
    .with_metrics(true)
    .listen("127.0.0.1:3000")
    .await
```

### After

```rust
Rapina::new()
    .router(router)
    .enable_introspection()
    .enable_health_check()
    .enable_metrics()
    .listen("127.0.0.1:3000")
    .await
```

### Disabling in tests (debug builds)

```rust
// In tests, disable introspection to avoid the /__rapina/routes endpoint
Rapina::new()
    .router(router)
    .disable_introspection()
```

ShiraiEd:
## Summary

Adds new enable/disable Rapina app methods that wraps around the .with_* methods

## Related Issues

Closes #476

## Checklist

- [x] `cargo fmt` passes
- [x] Tests pass

### DIFF
--- a/docs/content/docs/core-concepts/state.md
+++ b/docs/content/docs/core-concepts/state.md
@@ -184,11 +184,22 @@ The default timeout is 30 seconds. After the timeout, remaining connections are 
 
 ## Health Checks
 
-Enable built-in health endpoints with `.with_health_check(true)`:
+Enable built-in health endpoints with `.enable_health_check()`:
 
 ```rust
 Rapina::new()
-    .with_health_check(true)
+    .enable_health_check()
+    .listen("127.0.0.1:3000")
+    .await
+```
+
+When the value comes from config or environment, use `.with_health_check(bool)` to keep the builder chain intact:
+
+```rust
+let cfg = Config::from_env();
+
+Rapina::new()
+    .with_health_check(cfg.health_check_enabled)
     .listen("127.0.0.1:3000")
     .await
 ```
@@ -217,11 +228,16 @@ readinessProbe:
 
 The liveness probe **never** checks external dependencies — a DB outage should pull the pod from the load balancer (readiness failure), not restart it (liveness failure).
 
-Register custom checks for Redis, external APIs, or any dependency:
+Custom checks
+
+`.enable_health_check()` registers the HTTP endpoints. `.add_health_check()` registers a function that runs inside those endpoints — they do different things:
+
+- **`.enable_health_check()`** (or `.with_health_check(bool)` for dynamic config) — turns on the `/__rapina/health` routes. Without it, those paths return 404.
+- **`.add_health_check("name", fn)`** — adds a dependency check (Redis, Stripe, etc.) that runs on every `/ready` request. Optional — the endpoint works fine without any custom checks.
 
 ```rust
 Rapina::new()
-    .with_health_check(true)
+    .enable_health_check()                          // or: .with_health_check(cfg.health_check_enabled)
     .add_health_check("redis", || async {
         redis_ping().await.is_ok()
     })
--- a/rapina/src/app.rs
+++ b/rapina/src/app.rs
@@ -385,15 +385,35 @@ impl Rapina {
 
     /// Enables or disables the introspection endpoint.
     ///
-    /// When enabled, a `GET /.__rapina/routes` endpoint is registered
+    /// When enabled, a `GET /__rapina/routes` endpoint is registered
     /// that returns all routes as JSON.
     ///
     /// Introspection is enabled by default in debug builds.
+    /// Use this when the value is dynamic (e.g. read from config/env).
+    /// For literal call sites, prefer [`enable_introspection`](Self::enable_introspection)
+    /// or [`disable_introspection`](Self::disable_introspection).
     pub fn with_introspection(mut self, enabled: bool) -> Self {
         self.introspection = enabled;
         self
     }
 
+    /// Enables the introspection endpoint.
+    ///
+    /// Convenience wrapper for `.with_introspection(true)`.
+    pub fn enable_introspection(self) -> Self {
+        self.with_introspection(true)
+    }
+
+    /// Disables the introspection endpoint.
+    ///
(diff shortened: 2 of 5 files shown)


ShiraiEd:
Is the way that this is displayed in the docs good?

rapina/src/app.rs
```diff
@@ -484,7 +519,7 @@ impl Rapina {
     /// The function is called on every `GET /__rapina/health` request.
     /// Return `true` if healthy, `false` if not.
     ///
-    /// Requires `.with_health_check(true)` to be set.
+    /// Requires `.enable_health_check()` to be set.
     pub fn add_health_check<F, Fut>(mut self, name: &'static str, f: F) -> Self
```
richasdev, line 584: risk: no guard against registering health checks when `self.health_check == false`. Silent no-op if the caller forgetsn`.enable_health_check()`

 consider logging a warning or panicking at startup when checks are registered but the feature is disabled

rapina/src/app.rs
```diff
@@ -1084,6 +1147,18 @@ mod tests {
         assert!(!app.metrics);
     }
 
+    #[test]
+    fn test_rapina_enable_metrics() {
+        let app = Rapina::new().enable_metrics();
+        assert!(app.metrics);
+    }
+
+    #[test]
```
richasdev: `enable_health_check` and `disable_health_check` have no tests. Every other wrapper pair (`introspection`, `metrics`) got coverage, this one was missed


 add `test_rapina_enable_health_check` and `test_rapina_disable_health_check`

uemuradevexe (changes requested):

rapina/src/app.rs
```diff
@@ -1072,6 +1148,42 @@ mod tests {
...
+        let app = Rapina::new().enable_health_check();
+        assert!(app.health_check);
+    }
+
+    #[test]
+    fn test_rapina_disable_health_check() {
+        let app = Rapina::new().disable_health_check();
+        assert!(!app.health_check);
+    }
+
```
arferreira: These tests are testing one-liner wrappers. Also, for example `enable_introspection()` one, `test_rapina_with_introspection_enabled` already covers that path. Same pattern for all six. Drop them, they add maintenance cost without catching anything the existing tests don't already prove.

rapina/src/app.rs
```diff
@@ -481,10 +528,16 @@ impl Rapina {
 
     /// Registers a custom health check function.
     ///
-    /// The function is called on every `GET /__rapina/health` request.
-    /// Return `true` if healthy, `false` if not.
+    /// Registers a named async health check function.
```
arferreira, line 574: I guess you can drop this and start direct the paragraph with "The function is called on every `GET /__rapina/health/ready` request. That second line repeats the summary.

rapina/src/app.rs
```diff
@@ -385,28 +385,75 @@ impl Rapina {
...
+    }
+
     /// Enables or disables the built-in health check endpoints.
     ///
     /// When enabled, registers readiness and liveness health check endpoints:
     /// - `GET /__rapina/health` — alias for readiness, for simple setups and load balancers
     /// - `GET /__rapina/health/live` — liveness probe, always returns `200 OK`
     /// - `GET /__rapina/health/ready` — readiness probe, runs DB and custom checks
     ///
-    /// Health check is disabled by default.
```
arferreira, line 445: The "disabled by default" note was useful to readers scanning the doc. Keep it and add the new guidance after it, not instead of it.

rapina/src/app.rs
```diff
@@ -565,12 +618,28 @@ impl Rapina {
     /// When enabled, a `GET /metrics` endpoint is registered
     /// that returns all metrics to Prometheus.
     ///
-    /// Metrics is disabled by default unless you call `with_metrics(true)`.
```
arferreira, line 609: I guess this is same as `with_health_check` docstring, this note is useful.

docs/content/docs/core-concepts/state.md
```diff
@@ -217,11 +228,16 @@ readinessProbe:
 
 The liveness probe **never** checks external dependencies — a DB outage should pull the pod from the load balancer (readiness failure), not restart it (liveness failure).
 
-Register custom checks for Redis, external APIs, or any dependency:
+### Custom checks
```
arferreira: This section is already under `## Health Check`, that `###` is adding unnecessary hierarchy.

arferreira (changes requested):
Good work so far, just some nits to address before we merge.

arferreira (approved):
Nice, thanks!

uemuradevexe (approved):

PR #499, opened 2026-04-15
feat: emit sensible defaults and timestamp columns for add resource

### LINKED ISSUE #496: rapina add resource should emit sensible defaults and timestamp columns
The migration generated by `rapina add resource X field:type ...` emits every column as `.type().not_null()` with no default value and no timestamp columns. That forces manual edits on virtually every generated migration before it's usable.

Concretely, running `rapina add resource User email:string password:string active:boolean` produces:

```rust
.col(ColumnDef::new(Users::Email).string().not_null())
.col(ColumnDef::new(Users::Password).string().not_null())
.col(ColumnDef::new(Users::Active).boolean().not_null())
```

Two problems with this output:

A `not_null` boolean with no default means every insert has to set it explicitly. A boolean column should default to `false` unless the user says otherwise, which matches what Rails, Django, and most Active Record-style tooling do.

There's no `created_at` or `updated_at`. SeaORM's `ActiveModelBehavior` and `before_save` hooks assume those columns exist. Omitting them by default is a trap that surfaces later when someone tries to use `.save()` and gets runtime errors, or writes a hook and has to go back and add a new migration. Rails adds `timestamps` by default for this exact reason.

Proposed behavior:

```rust
.col(ColumnDef::new(Users::Email).string().not_null())
.col(ColumnDef::new(Users::Password).string().not_null())
.col(
    ColumnDef::new(Users::Active)
        .boolean()
        .not_null()
        .default(Expr::value(false)),
)
.col(
    ColumnDef::new(Users::CreatedAt)
        .timestamp_with_time_zone()
        .not_null()
        .default(Expr::current_timestamp()),
)
.col(
    ColumnDef::new(Users::UpdatedAt)
        .timestamp_with_time_zone()
        .not_null()
        .default(Expr::current_timestamp()),
)
```

Rules:

- `boolean` columns default to `false`
- `created_at` and `updated_at` are injected on every generated migration with `timestamp_with_time_zone` and `default(Expr::current_timestamp())`
- A `--no-timestamps` flag opts out for the rare case someone really doesn't want them

Other types (`string`, `integer`, etc.) stay unchanged. There's no universally sensible default for those.

Implementation notes: the column mapping lives in `rapina-cli/src/commands/add.rs` around the `parse_field` match. Timestamp injection belongs in `generate_migration` in `rapina-cli/src/commands/codegen.rs`, right after the PK column is emitted and before the user-supplied fields loop. The `DeriveIden` enum also needs the `CreatedAt` and `UpdatedAt` variants added.

uemuradevexe:
## Summary

 - `bool`/`boolean` fields now emit `.default(Expr::value(false))` in                                                                                   
    the generated migration — avoids requiring every insert to set the
    field explicitly                                                                                                                                     
  - `created_at` and `updated_at` (`TIMESTAMPTZ NOT NULL DEFAULT CURRENT_TIMESTAMP`)                                                                                                are injected automatically into every migration                                                                                  
    and the entity `schema!` block, matching Rails/Django conventions                                                                                  
    and satisfying SeaORM `ActiveModelBehavior`                                                                                                          
  - `--no-timestamps` flag opts out; entity receives `#[timestamps(none)]`                                                                             
    to keep migration and entity in sync
## Related Issues

Closes #496 

## Checklist

- [x] `cargo fmt` passes
- [x] Tests pass

### DIFF
--- a/rapina-cli/src/commands/add.rs
+++ b/rapina-cli/src/commands/add.rs
@@ -36,7 +36,11 @@ fn parse_field(input: &str) -> Result<FieldInfo, String> {
         "i64" | "bigint" => ("i64", "i64", ".big_integer().not_null()"),
         "f32" | "float" => ("f32", "f32", ".float().not_null()"),
         "f64" | "double" => ("f64", "f64", ".double().not_null()"),
-        "bool" | "boolean" => ("bool", "bool", ".boolean().not_null()"),
+        "bool" | "boolean" => (
+            "bool",
+            "bool",
+            ".boolean().not_null().default(Expr::value(false))",
+        ),
         "uuid" => ("Uuid", "Uuid", ".uuid().not_null()"),
         "datetime" | "timestamptz" => (
             "DateTimeUtc",
@@ -107,7 +111,7 @@ fn print_next_steps(pascal: &str) {
     println!();
 }
 
-pub fn resource(name: &str, field_args: &[String]) -> Result<(), String> {
+pub fn resource(name: &str, field_args: &[String], with_timestamps: bool) -> Result<(), String> {
     validate_resource_name(name)?;
     codegen::verify_rapina_project()?;
 
@@ -134,9 +138,11 @@ pub fn resource(name: &str, field_args: &[String]) -> Result<(), String> {
 
     let pk_type = "i32"; // New resources default to i32 PK
 
+    let timestamps = if with_timestamps { None } else { Some("none") };
+
     codegen::create_feature_module(singular, plural, pascal, &fields, pk_type, false)?;
-    codegen::update_entity_file(pascal, &fields, None, None, false)?;
-    codegen::create_migration_file(plural, pascal_plural, &fields, pk_type)?;
+    codegen::update_entity_file(pascal, &fields, timestamps, None, false)?;
+    codegen::create_migration_file(plural, pascal_plural, &fields, pk_type, with_timestamps)?;
 
     if let Err(e) = codegen::wire_main_rs(&[plural.as_str()], Path::new(".")) {
         eprintln!("  {} Could not auto-wire main.rs: {}", "!".yellow(), e);
@@ -263,7 +269,7 @@ mod tests {
                 name: "active".to_string(),
                 rust_type: "bool".to_string(),
                 schema_type: "bool".to_string(),
-                column_method: ".boolean().not_null()".to_string(),
+                column_method: ".boolean().not_null().default(Expr::value(false))".to_string(),
                 nullable: false,
             },
         ];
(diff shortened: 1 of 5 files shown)


rapina-cli/src/commands/import.rs
```diff
@@ -577,7 +577,7 @@ fn generate_for_table(
     };
 
     codegen::update_entity_file(&pascal, &fields, timestamps, primary_key.as_deref(), force)?;
-    codegen::create_migration_file(plural, &pascal_plural, &fields, pk_type)?;
+    codegen::create_migration_file(plural, &pascal_plural, &fields, pk_type, false)?;
```
ShiraiEd, line 581: Would be good to comment that is  always set to `false` because the db already owns the schema,
in case users get confused about the why imported tables never get timestamp columns

ShiraiEd (changes requested):
Nice one, just a few changes

3 of the 5 new tests are already covered:
- `test_generate_migration_no_timestamps` -> this was covered in:
```rust
`test_generate_migration` recived ``assert!(!content.contains("CreatedAt"));
assert!(!content.contains("UpdatedAt"));``
```

- test_schema_block_no_timestamps_attr_when_with_timestamps` + `test_schema_block_timestamps_none_when_no_timestamps` -> they are already covered in `test_generate_schema_block_with_timestamps`, with:
```rust
let block = generate_schema_block("Post", &fields, None, None);
assert!(!block.contains("#[timestamps")); // same as test_schema_block_no_timestamps_attr_when_with_timestamps

let block = generate_schema_block("Post", &fields, Some("none"), None);
assert!(block.contains("#[timestamps(none)]")); // same as test_schema_block_timestamps_none_when_no_timestamps
```

NITTS
-  **Silent `false` in `import.rs`**, Always opts out of timestamps when importing from DB. Probably correct (the DB already owns the schema), but there's no comment and no flag to opt in, a future user might wonder why imported tables never get timestamp columns.
- The variable name `timestamps` doesn't signal it's an attribute string for the entity. `timestamps_attr` or `schema_timestamps_attr` would make the intent clearer.

rapina-cli/src/commands/add.rs
```diff
@@ -134,9 +138,11 @@ pub fn resource(name: &str, field_args: &[String]) -> Result<(), String> {
 
     let pk_type = "i32"; // New resources default to i32 PK
 
+    let timestamps = if with_timestamps { None } else { Some("none") };
```
arferreira: A reader seeing `if with_timestamps { None }` has to understand that `None` means "use default" and "default includes timestamps" in the `generate_schema_block` API.
```suggestion
    // None = no #[timestamps] attr → schema! macro adds both (default).
    // Some("none") = #[timestamps(none)] → macro skips them.
    let timestamps = if with_timestamps { None } else { Some("none") };
```

arferreira (commented):
Looks great overall, just that mapping (my comment suggestion) that could cause confusion.

ShiraiEd (approved):
LGTM

PR #510, opened 2026-04-18
feat(url-shortener): use State<AppConfig> in create handler

### LINKED ISSUE #509: [Good First Issue]:  url-shortener: use State<AppConfig> in handlers
### Summary

`AppConfig` is already registered via `.state(config)` in `lib.rs`, but no handler ever extracts it, so the pattern is invisible. At least one handler should accept `config: State<AppConfig>` and use a config value to drive behavior.

### Why This Matters

Registering state without ever using it in a handler is misleading, it looks like boilerplate rather than a feature. Someone reading the example learns that `.state()` exists but not how to consume it, which is the part that matters for real apps.

### Relevant Files

- `rapina/examples/url-shortener/url-shortner/src/lib.rs`
- `rapina/examples/url-shortener/url-shortner/src/urls/handlers.rs`
- `rapina/examples/url-shortener/url-shortner/src/urls/service.rs`

### Acceptance Criteria

- [ ] At least one handler (e.g., `create_url`) accepts `config: State<AppConfig>` in its signature
- [ ] The config value is used meaningfully — for example, `AppConfig` gains a `default_expiry_days: u32` field and `create_url` uses it as the fallback when `expires_at` is not provided in the request, replacing the current hardcoded `24 * 365 * 3600` in `service.rs
- [ ] The example compiles and the `POST /api/v1/shorten` endpoint still works
- [ ] No other handlers are broken
- [ ] .env.example` is updated to include `DEFAULT_EXPIRY_DAYS=365

### Estimated Effort

S — ~1 hour

### Hints & Resources

- Look at how similar functionality is implemented as a reference
- Follow existing patterns in the codebase
- `service::create` currently hardcodes the expiry in `service.rs`. A good change is to pass `config.default_expiry_days` from the handler to the service function instead.
- `State<T>` can be used alongside `Db` and other extractors in the same handler — order in the function signature doesn't matter.
- AppConfig` needs `#[env = "DEFAULT_EXPIRY_DAYS"]` and `#[default = "365"]` added for the new field.

### Maintainer Checklist

- [x] Issue is self-contained and scoped for a first-time contributor
- [x] Relevant files and context are provided above
- [x] Acceptance criteria are clear and verifiable

mvanhorn:
## Summary

- Adds `config: State<AppConfig>` extractor to `create_url` handler
- Uses `config.host` and `config.port` to construct a full `short_url` in the response
- Adds `short_url` field to `CreateUrlResponse`
- Demonstrates the `State<T>` extractor pattern for the url-shortener example

The `AppConfig` was already registered via `.state(config)` in `build_app()` but no handler consumed it, making the pattern invisible to developers reading the example.

Closes #509

## Test plan

- [ ] `POST /api/v1/shorten` now returns `short_url` with the configured host and port
- [ ] Existing `short_code` and `long_url` fields unchanged


ShiraiEd (changes requested):
Hi, thanks for the contribution! The `State<AppConfig>` extractor and `short_url` in the response are nice additions.

However, the core asked from #509 is still missing:

- Add `default_expiry_days: u32` to `AppConfig` in `lib.rs` (`#[env = "DEFAULT_EXPIRY_DAYS"]`, `#[default = "365"]`)
- Pass it from `create_url` to `service::create` and use it as the fallback when `expires_at` is not provided in the request by replacing the hardcoded `24 * 365 * 3600` in `service.rs`
- Update `.env.example` with `DEFAULT_EXPIRY_DAYS=365

ShiraiEd:
Hi @mvanhorn, how are the changes going so far?

rapina/examples/url-shortener/url-shortner/src/urls/handlers.rs
```diff
@@ -39,11 +40,13 @@ pub async fn redirect(db: Db, code: Path<String>) -> Result<http::Response<BoxBo
 #[public]
 #[post("/", group = "/api/v1/shorten")]
 #[errors(UrlsError)]
-pub async fn create_url(db: Db, body: Validated<Json<CreateUrlRequest>>) -> Result<Json<CreateUrlResponse>> {
+pub async fn create_url(db: Db, config: State<AppConfig>, body: Validated<Json<CreateUrlRequest>>) -> Result<Json<CreateUrlResponse>> {
     let input = body.into_inner().into_inner();
     let result = service::create(db.conn(), input).await?;
+    let short_url = format!("http://{}:{}/api/v1/shorten/{}", config.host, config.port, result.short_code);
```
arferreira, line 47: If someone configures a host with HTTPS or puts this behind a reverse proxy, the URL is wrong. For an example app this is probably fine, but a comment like `// simplification: real apps would use a base_url config` field would set expectations.

arferreira (dismissed):
Thanks for the contribution! Just left a comment suggestion, which is not a blocker for merge.

mvanhorn:
@ShiraiEd all three changes are in f148497:

- `AppConfig` gains `default_expiry_days: u32` with `#[env = "DEFAULT_EXPIRY_DAYS"]` and `#[default = "365"]` (lib.rs)
- `service::create` now takes `default_expiry_days` and uses `u64::from(default_expiry_days) * SECONDS_PER_DAY` as the fallback for `expires_at`, replacing the hardcoded `24 * 365 * 3600` (service.rs)
- `.env.example` documents `DEFAULT_EXPIRY_DAYS=365`

Also addressed @arferreira's inline note in b2019e8 by adding a `// simplification` comment on the `short_url` format string. Verified: `cargo check` passes on the `url-shortner` crate.

mvanhorn:
@ShiraiEd Addressed in f148497:

- Added `default_expiry_days: u32` on `AppConfig` with `#[env = "DEFAULT_EXPIRY_DAYS"]` and `#[default = "365"]`.
- `service::create` now takes `default_expiry_days` and uses it (via a `SECONDS_PER_DAY` constant) instead of the hardcoded `24 * 365 * 3600` seconds.
- `handlers::create_url` threads `config.default_expiry_days` through.
- `.env.example` documents `DEFAULT_EXPIRY_DAYS=365`.

Let me know if the shape matches what you had in mind.

rapina/examples/url-shortener/url-shortner/src/urls/handlers.rs
mvanhorn: Added the `// simplification` note in b2019e8 - flags that real apps would need a `base_url` config to handle HTTPS and reverse proxies.

ShiraiEd:
Great work !
Everything was addressed well


Just hang in there for a moment, we will update the `cargo deny` checks to unblock that Ci check
so it can be merged

ShiraiEd (approved):
LGTM

mvanhorn:
Thanks for the merge, @arferreira. State<AppConfig> reads cleaner here.

PR #513, opened 2026-04-20
feat(uuid): enable serde feature for Uuid serialization

### LINKED ISSUE #493: [Suggestion]:  Add `serde` feature to `uuid` dependency
### Suggestion

Add the `serde` feature to the `uuid` dependency in `rapina/Cargo.toml`:

```toml
# Current
uuid = { version = "1", features = ["v4"] }

# Proposed
uuid = { version = "1", features = ["v4", "serde"] }
```

### Context

Currently, the `rapina` crate re-exports `uuid` , making it available for users of the framework. However, without the `serde` feature enabled, `uuid::Uuid` cannot be serialized/deserialized, which prevents users from:

1. **Using `Uuid` in path parameters** - When defining route handlers with `#[get("/:id")]` where `id: Uuid`, the deserialization from string to `Uuid` requires serde support
2. **Using `Uuid` in query parameters** - Similar issue with query string parsing
3. **Using `Uuid` in request/response bodies** - Cannot serialize/deserialize UUIDs in JSON payloads without the `serde` feature
4. **OpenAPI schema generation** - The `schemars` dependency already has `uuid1` feature enabled, but full integration requires serde support on the uuid side

### Scope

API Design

### Examples

**Current limitation (doesn't work without `serde` feature):**

```rust
use rapina::prelude::*;
use rapina::uuid::Uuid;

#[get("/:id")]
async fn get_user(Path(id): Path<Uuid>) -> String {
    format!("User ID: {}", id)
}

// This fails to compile because Uuid can't be deserialized from path string
```

uemuradevexe:
## Summary                              
  Added `"serde"` to the `uuid` feature flags in
  `Cargo.toml` — enables `Serialize`/`Deserialize`         
  for `uuid::Uuid`, which is required for users to use
  `Uuid` in `Path<T>`, `Query<T>`,                         
  and JSON body/response extractors via the re-exported
  `rapina::uuid`.                                          
                                                          
  Added 4 unit tests (`Path<Uuid>` valid/invalid,          
  `Query<Uuid>` valid/invalid) and 2 integration          
  tests (`Uuid` in JSON body and JSON response) — covers   
  all 3 use cases from the issue.                          
                                                 
  ## Related Issues                                        
  Closes #493               
                                                           
  ## Checklist                                            
  - [x] `cargo fmt` passes                       
  - [x] Tests pass

### DIFF
--- a/rapina/Cargo.toml
+++ b/rapina/Cargo.toml
@@ -40,7 +40,7 @@ validator = { version = "0.20.0", features = ["derive"] }
 tracing = "0.1"
 tracing-subscriber = { version = "0.3", features = ["json", "env-filter"] }
 
-uuid = { version = "1", features = ["v4"] }
+uuid = { version = "1", features = ["v4", "serde"] }
 httpdate = "1"
 # FIXME: move behind a `snapshot` feature flag
 regex = "1"
--- a/rapina/src/extract/mod.rs
+++ b/rapina/src/extract/mod.rs
@@ -1118,6 +1118,38 @@ mod tests {
         assert_eq!(err.status(), 400);
     }
 
+    #[tokio::test]
+    async fn test_query_extractor_uuid() {
+        #[derive(serde::Deserialize, PartialEq, Debug)]
+        struct Params {
+            id: uuid::Uuid,
+        }
+
+        let id = uuid::Uuid::new_v4();
+        let (parts, _) = TestRequest::get(&format!("/users?id={id}")).into_parts();
+        let result =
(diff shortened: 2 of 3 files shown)


ShiraiEd (commented):
Solid work!

Your tests cover:
- Path parameters
- Query parameters
- JSON body/response

Non-Blocking-Comment

The OpenAPI tests would be good, but too much for this pr,
if you think so it would be good to open an issue covering those.

also make sure to check the formating with `cargo fmt --all` before commiting

ShiraiEd (approved):
LGTM

PR #517, opened 2026-04-20
Fix in-flight request metric leak

### LINKED ISSUE #516: [Bug]: Permanent metric leak for in-flight gauge on cancelled requests
### Description

When a client unexpectedly disconnects before the server finishes processing their request, the underlying async runtime (Tokio/hyper) drops the executing request handler future. Because `MetricsMiddleware` increments the `http_requests_in_flight` gauge before calling `next.run(req).await` and manually decrements it afterwards, the decrement logic is never reached if the future is dropped at the `.await` yield point.

This causes a **permanent** metric leak where the `http_requests_in_flight` gauge continually climbs and never resets, leading to false alarms and potentially breaking auto-scaling like K8s Horizontal Pod Autoscaling (HPA) in production environments.

Common triggers for dropped connections include:

- User Navigation: A user clicks a button that triggers an API call, but immediately clicks a link to navigate away, causing the browser to abort the pending request
- Mobile Networks: A mobile client loses signal or switches from Wi-Fi to cellular mid-request, dropping the connection
- Load Balancer Timeouts: If a reverse proxy or load balancer (like NGINX or HAProxy) has a 30-second timeout and the handler logic takes 31 seconds to compute, the proxy closes the connection

### Steps to Reproduce

1. Add `.with_metrics(true)` to the `Rapina` instance
2. Create an endpoint that takes long to finish, for example by sleeping:
```
#[get("/sleep")]
async fn sleep() -> Json<String> {
    tracing::info!("sleeping");
    tokio::time::sleep(std::time::Duration::from_secs(10)).await;
    Json("Hello from Rapina!".to_string())
}
```
3. Execute `curl -X GET http://127.0.0.1:3000/metrics` and verify that `http_requests_in_flight` is 0
4. Send a request to that long-running endpoint using a client 
5. Abort the request on the client side before the sleep time elapses (10s in the example above)
6. Execute `curl -X GET http://127.0.0.1:3000/metrics` and note that `http_requests_in_flight` is now permanently increasing when you do steps 3) to 5) repeatedly

### Expected Behavior

The `http_requests_in_flight` gauge should accurately reflect the number of active requests. If a request is aborted mid-flight by the client, the framework should guarantee that the gauge is decremented back to `0`

### Code Example

```rust

```

### Environment

- Rapina version: v0.11.0

### Error Output / Logs

```shell

```

juv:
## Summary

Fixes the `http_requests_in_flight` metric leak in case the client closes the connection to the Rapina webserver before the endpoint handler finished processing.

## Related Issues

Closes #516

## Checklist

- [x] `cargo fmt` passes
- [x] Tests pass

### DIFF
--- a/rapina/src/metrics/middleware.rs
+++ b/rapina/src/metrics/middleware.rs
@@ -1,11 +1,11 @@
 use std::time::Instant;
 
-use hyper::body::Incoming;
-use hyper::{Request, Response};
-
 use crate::context::RequestContext;
 use crate::middleware::{BoxFuture, Middleware, Next};
 use crate::response::BoxBody;
+use hyper::body::Incoming;
+use hyper::{Request, Response};
+use prometheus::IntGauge;
 
 use super::prometheus::MetricsRegistry;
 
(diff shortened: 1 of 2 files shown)


arferreira (approved):
Solid fix! Thanks.

PR #519, opened 2026-04-20
Refactor/unified type system and expanded types

Ismaellima4:
## Summary

What does this PR do?

  * Type System Unification: Renames and centralizes the internal type representation in rapina-cli to NormalizedType, eliminating redundancies between Rust-specific and
     Schema-specific type enums.
   * New Native Types: Adds full support for Time and Bytes (mapped to Vec<u8>) across the entire workflow, including code generation, DTOs, and the schema! macro.
   * Enhanced Database Introspection:
       * SQLite: The import command now recognizes Custom column types (such as UUID, DATETIME, BINARY) by intelligently parsing the inner type name.
       * Postgres/MySQL: Fixed pattern matching for sea-schema variants like Time(_) and Blob(_), which previously caused import warnings.
   * Improved DX (Developer Experience): Significantly expands the list of supported type aliases (e.g., blob, binary, timestamp, numeric, money, jsonb), allowing the CLI
     to accept common database naming conventions without manual correction.
   * System Consistency: Ensures that any type generated by rapina import or rapina add resource is immediately recognized by the schema! macro, preventing "Unknown type"
     compiler errors.

## Related Issues

Closes #

## Checklist

- [x] `cargo fmt` passes
- [x] Tests pass


rapina-cli/src/commands/mod.rs
```diff
@@ -18,6 +18,247 @@ pub mod test;
...
+            NormalizedType::Json => ".json()",
+            NormalizedType::Bytes => ".binary()",
+            NormalizedType::Time => ".time()",
+            #[cfg(feature = "import")]
+            NormalizedType::Unmappable(_) => "",
+        };
+
+        let column_method = ColumnMethod(base.to_string());
+
+        if nullable || base.is_empty() {
```
richasdev: Nullable columns silently lose the `.null()` suffix in the generated method chain. `ColumnMethod::new` with `nullable=true` returns `.text()` instead of `.text().null()`. SeaORM does not default to nullable the generated migration will implicitly apply `NOT NULL`, causing schema drift. The test `test_normalized_to_field_info_nullable` was updated to assert the wrong behavior instead of catching the regression.

rapina-cli/src/commands/codegen.rs
```diff
@@ -502,14 +506,16 @@ pub(crate) fn generate_migration(
 
     let readable_name = format!("create {}", plural);
 
-    let col = match pk_type.to_lowercase().as_str() {
-        "uuid" => format!("ColumnDef::new({pascal_plural}::Id).uuid().not_null().primary_key()"),
-        "i32" | "integer" => {
+    let col = match pk_type {
```
richasdev: The match on `pk_type` only covers `Uuid`, `I32`, and `I64`. The original code had a `_` wildcard arm with an `Alias::new` fallback. The PR replaced `&str` with `&NormalizedType` without covering the remaining 12+ variants. Rust rejects non-exhaustive matches on concrete enums  any `NormalizedType` outside the three covered variants breaks the build.

uemuradevexe (changes requested):

rapina-cli/src/commands/import.rs
richasdev: `let tables` type cannot be inferred by `rt.block_on`. Fails to compile with `--features import`:                                                      
                                                                                                                                                         
  error[E0282]: type annotations needed   
                                                                                                                                                         
  Fix:                                                                                                                                                   
 `let tables: Result<Vec<IntrospectedTable>, String> = rt.block_on(async {...}`

uemuradevexe (changes requested):

rapina-cli/src/commands/import.rs
```diff
@@ -131,52 +131,23 @@ fn map_sqlite_type(col_type: &sea_schema::sea_query::ColumnType) -> NormalizedTy
...
-        rust_type,
-        schema_type,
-        column_method: format!("{}{}", column_base, null_suffix),
-        nullable: is_nullable,
-    })
+    Some(FieldInfo::new(
+        col_name.to_string(),
+        normalized_type,
+        is_nullable,
+        false,
```
richasdev: `is_primary_key` is hardcoded to `false` here. Columns named `id` imported from the DB will generate a migration without `.primary_key()` / `.auto_increment()`, causing schema drift.

**Fix:** pass `col_name == "id"` as the 4th argument:

```rust
Some(FieldInfo::new(
    col_name.to_string(),
    normalized_type,
    is_nullable,
    col_name == "id",
))
```

uemuradevexe (changes requested):

rapina-cli/src/commands/mod.rs
```diff
@@ -18,6 +18,275 @@ pub mod test;
...
+        }
+    }
+}
+
+#[derive(Debug, Clone)]
+pub(crate) struct FieldInfo {
+    pub name: String,
+    pub ident: String,
+    pub normalized_type: NormalizedType,
+    _column_method: String,
```
arferreira: Hmmm dead field... always written as `String::new()` and never read. Either use it or delete it.

rapina-cli/src/commands/mod.rs
```diff
@@ -18,6 +18,275 @@ pub mod test;
...
+        normalized_type: NormalizedType,
+        nullable: bool,
+        is_primary_key: bool,
+    ) -> Self {
+        let ident = codegen::to_pascal_case(&name);
+        Self {
+            name,
+            ident,
+            normalized_type,
+            _column_method: String::new(),
```
arferreira: I commented about this dead code above.

arferreira: Also `FieldInfo::from_str` validates the name via `ValidationContext::Field.validate(name)?` but `new` skips it. Two construction paths, one invariant. An imported column named `123` or `self` gets past `new` and breaks later in codegen with a worse error. Move the validation into a shared helper both paths call.

rapina-cli/src/commands/mod.rs
```diff
@@ -18,6 +18,275 @@ pub mod test;
...
+            NormalizedType::DateTimeUtc => "DateTime".to_string(),
+            NormalizedType::DateTime => "NaiveDateTime".to_string(),
+            NormalizedType::Date => "Date".to_string(),
+            NormalizedType::Decimal => "Decimal".to_string(),
+            NormalizedType::Json => "Json".to_string(),
+            NormalizedType::Bytes => "Vec<u8>".to_string(),
+            NormalizedType::Time => "Time".to_string(),
+            #[cfg(feature = "import")]
+            NormalizedType::Unmappable(s) => s.clone(),
+        };
```
arferreira: This only depends on `normalized_type` and `nullable`, not on field-level state. Move to `NormalizedType::schema_type_name(&self, nullable: bool)` or a free function. Same comment doesn't apply to `generate_column` since that one uses `ident` and `is_primary_key`, so it correctly stays on `FieldInfo`.

rapina-cli/src/commands/mod.rs
```diff
@@ -18,6 +18,275 @@ pub mod test;
...
+            NormalizedType::Decimal => "Decimal",
+            NormalizedType::Json => "Json",
+            NormalizedType::Bytes => "Vec<u8>",
+            NormalizedType::Time => "Time",
+            #[cfg(feature = "import")]
+            NormalizedType::Unmappable(s) => s,
+        };
+        write!(f, "{}", s)
+    }
+}
```
arferreira, line 111: Display returns "DateTimeUtc" for the `DateTimeUtc` variant, but `FieldInfo::schema_type_name` returns "DateTime" for the same variant. Same input, two strings, the only signal which is which is the method name. Either rename `Display` to something like `rust_type_name()` to make the distinction visible, or add a doc comment on both methods. Right now while reading couldn't predict the output.

rapina-cli/src/commands/mod.rs
```diff
@@ -18,6 +18,275 @@ pub mod test;
...
+
+#[derive(Debug, Clone)]
+pub(crate) struct FieldInfo {
+    pub name: String,
+    pub ident: String,
+    pub normalized_type: NormalizedType,
+    _column_method: String,
+    pub nullable: bool,
+    pub is_primary_key: bool,
```
arferreira, line 138: Sorry, two sources of truth for primary-key membership. `generate_schema_block` already takes a separate `pk: Vec<String>` for composite keys, migration emission now reads `field.is_primary_key`. They can disagree. I mean, not introduced by this PR but cemented by it. Please, would you mind filing a follow-up?

rapina-cli/src/commands/mod.rs
```diff
@@ -18,6 +18,275 @@ pub mod test;
...
+    pub(crate) fn validate(&self, name: &str) -> Result<(), String> {
+        let ctx_prefix = self.as_str();
+
+        if name.is_empty() {
+            return Err(format!("{} cannot be empty", ctx_prefix));
+        }
+
+        if name.chars().next().unwrap().is_ascii_digit() {
+            return Err(format!("{} cannot start with a digit", ctx_prefix));
+        }
```
arferreira:
```suggestion
        if name.starts_with(|c: char| c.is_ascii_digit()) {
            return Err(format!("{} cannot start with a digit", ctx_prefix));
        }
```

rapina-cli/src/commands/mod.rs
```diff
@@ -18,6 +18,275 @@ pub mod test;
...
+    DateTimeUtc,
+    DateTime,
+    Date,
+    Decimal,
+    Json,
+    Bytes,
+    Time,
+    #[cfg(feature = "import")]
+    #[allow(dead_code)]
+    Unmappable(String),
```
arferreira: This variant only exists with the feature on, so every match arm on `NormalizedType` has to remember to cfg-gate the `Unmappable` branch. You're actually doing it correctly. The next person to add a method will forget. Cheaper option IMO: keep the variant unconditional, gate only the import code that produces it.

rapina-cli/src/commands/mod.rs
```diff
@@ -18,6 +18,275 @@ pub mod test;
 
 #[cfg(feature = "import-openapi")]
 pub mod import_openapi;
+pub(crate) use colored::Colorize;
```
arferreira: Re exporting an external trait through the module hides the dependency. Each consumer should `use colored::Colorize` directly.

rapina-cli/src/commands/mod.rs
arferreira, line 16: I missed a more robust file level comment here!

arferreira (changes requested):
Hey good work so far, thanks! Just left some improvements/nits that would be great of having before merge.

arferreira (approved):
LGTM, thanks! Sorry, one more thing, that composite PK source of truth thing I flagged is still open, can you file a follow-up issue so we don't lose it?

uemuradevexe (commented):
LGTM!

uemuradevexe (approved):

PR #524, opened 2026-04-22
fix/migrate command execution

### LINKED ISSUE #520: `rapina migrate` can't actually run migrations
`rapina migrate new <name>` scaffolds a file and that's it. There's no way to apply it from the CLI.

I got stuck trying to run a SeaORM migration in a Rapina app. The options I found were:

1. Call `run_migrations::<Migrator>()` at app startup. Runs on every boot, no rollback, no status, and on multi-replica deploys it races.
2. Hand-write `src/bin/migrate.rs` with a tiny `main` that calls into `sea_orm_migration`. Works, but it's the exact boilerplate `sea-orm-cli` exists to avoid, and now every Rapina project has to reinvent it.

Neither is what I want from a framework that ships a CLI.

What should work:

```
rapina migrate up
rapina migrate down [--steps N]
rapina migrate status
rapina migrate fresh
rapina migrate reset
```

The reason this isn't already there is the same reason `sea-orm-cli` uses a separate crate: `rapina-cli` is a standalone binary, it can't link against the user's `Migrator`. So the CLI can't just call `rapina::migration::run_pending` directly.

Way I'd do it:

- Add `rapina::migration::run_cli::<M: MigratorTrait>()` in `rapina`. Parses args, reads `DATABASE_URL`, dispatches.
- `rapina new` scaffolds `src/bin/rapina_migrate.rs` with `fn main() { rapina::migration::run_cli::<crate::migrations::Migrator>() }`.
- `rapina migrate up` etc. shell out to `cargo run --bin rapina_migrate -- up`.

Existing projects won't have the bin target. For those, fail with a clear message pointing at `rapina migrate init` (or whatever we call it) that drops the file in.

One thing I'm not sure about: do we keep `run_migrations::<Migrator>()` at startup as a convenience, or drop it now that there's a real CLI path? I'd keep it, off by default, with a note in the docs that it's fine for single-node and a footgun otherwise. Curious what others think.

Relates to the broader "solve the trust problem" framing: running migrations from the web process is one of those things that works until it doesn't, and there's no reason the framework should make it the default path.

uemuradevexe:
## Summary

the CLI cannot link against the user's `Migrator` type, so each subcommand shells out to `cargo run -q --binrapina_migrate -- <cmd>`. The binary lives at `src/bin/rapina_migrate.rs` (scaffolded by `rapina migrate init` or `rapina new --db`) and calls `rapina::migration::run_cli::<Migrator>()`, which loads `.env`, reads `DATABASE_URL`, parses args, and dispatches to sea-orm-migration.

  - `rapina migrate up` — apply all pending migrations
  - `rapina migrate down [--steps N]` — roll back (default: 1 step)
  - `rapina migrate status` — list applied and pending migrations
  - `rapina migrate fresh` → `Migrator::fresh()` — drop all tables + re-run
  - `rapina migrate reset` → `Migrator::refresh()` — down all + up all
  - `rapina migrate init` — scaffolds `src/bin/rapina_migrate.rs` for existing projects
  - Clear error if the binary is missing, pointing to `rapina migrate init`

## Related Issues

Closes #520 

## Checklist

- [x] `cargo fmt` passes
- [x] Tests pass


ShiraiEd (changes requested):
Solid approach, a few things worth addressing

- **`Path::new(".")` assumption (`check_migrate_bin` / `run_migrate_cmd`)**

Both functions resolve the binary path relative to the current working directory. If the user runs `rapina migrate up` from a subdirectory (e.g. `src/`), it silently falls through to the "binary not found" error, which points them to `rapina migrate init`, but that's not the actual problem. A better UX would be to detect the project root (e.g. walk up looking for `Cargo.toml`) and give a more specific error if CWD is wrong.

- **Hardcoded `#[path = "../migrations/mod.rs"]` in scaffolded binary**

This assumes `src/bin/` and `src/migrations/` are siblings, which holds for the default layout but breaks the moment someone customizes their structure. Worth either documenting this constraint clearly or making the path configurable via `rapina migrate init`.

- **Silent recompile on every `rapina migrate up`**

`cargo run -q` will recompile if sources changed. On slow machines this can take several seconds with no output (the `-q` suppresses it). Consider printing something like `Compiling migrate binary...` before the `cargo run`, or dropping `-q` so users see what's happening instead of a frozen terminal.

- **Docs don't cover the new commands**

The `migrations.md` page only covers `rapina migrate new` and `.run_migrations()` at startup, none of the new commands are documented. The following should be added:

- `rapina migrate init` what it does and when to use it (existing projects)
- `rapina migrate up/down/status/fresh/reset` 
- What `src/bin/rapina_migrate.rs` is and why it exists
- A note that `.run_migrations()` is a footgun on multi-replica deploys and that `rapina migrate up` is the recommended path for production

- **Tests**

- `run_migrate_cmd` it's the main function but it doen't have any tests,
- `parse_steps` unexpected argument branch is untested**, the `other => Err(...)` path in `parse_steps` is never hit by any test
- `test_dispatch_down` only tests 1 step, the `--steps N` path with N > 1 is never exercised at the trait dispatch level

rapina/tests/migration_test.rs
```diff
@@ -218,6 +331,27 @@ mod migrator_trait_tests {
...
+    #[tokio::test]
+    async fn test_dispatch_down_multiple_steps() {
+        // Apply migrations then roll back N > 1 steps.
+        // With a single migration the result is the same as steps=1, but
+        // this exercises the Some(N) path at the trait-dispatch level.
+        let conn = Database::connect("sqlite::memory:").await.unwrap();
+        Migrator::up(&conn, None).await.unwrap();
+        // SeaORM clamps steps to the number of applied migrations, so this should succeed.
+        Migrator::down(&conn, Some(5)).await.unwrap();
+    }
```
ShiraiEd: This is redundant, `test_dispatch_down_two_steps()` is covering N > 1.

with only 1 migration applied, `down(Some(5))` still only rolls back 1. The `Some(N)` path is hit in code but the behavior is identical to `Some(1)`,
so it proves nothing about multi-step rollback, it works just like `test_dispatch_down()` above

ShiraiEd (changes requested):
Just deleted that test that I comment and it will be good to me

ShiraiEd (approved):
LGTM

PR #528, opened 2026-04-23
feat: add llms.txt endpoint and CLI export command

### LINKED ISSUE #522: Serve `/llms.txt` from the running app
**Background, for anyone picking this up cold**

[llms.txt](https://llmstxt.org/) is a proposed convention. Think of it as a `robots.txt` but for LLMs. A single Markdown file, served at a well-known path, that gives an agent a compact, reliable summary of what a project is and how to work with it. The spec itself is short and worth reading: [llmstxt.org](https://llmstxt.org/). The proposal came from [Jeremy Howard in September 2024](https://www.answer.ai/posts/2024-09-03-llmstxt.html) and has picked up adoption in the docs-site space through 2025-2026.

The motivation is practical. LLMs trained before a project existed have no idea it's there. LLMs trained before your last breaking change are working from stale information. When an agent needs to reason about your API, it either scrapes HTML (slow, noisy) or it hallucinates. `llms.txt` gives it a Markdown file it can read in one fetch.

Two flavors exist in the spec:

- `llms.txt`: an index. Short description plus links to detailed pages.
- `llms-full.txt`: everything inlined in one file.

**Who ships this today**

Mostly docs sites, not applications:

- [Hono](https://hono.dev/llms.txt), JS web framework, serves it on their docs.
- [Anthropic docs](https://docs.anthropic.com/llms.txt).
- [Cloudflare](https://developers.cloudflare.com/llms.txt).
- [Svelte](https://svelte.dev/llms.txt).

An [unofficial directory](https://llmstxt.site/) tracks adopters.

Notably missing: web frameworks serving it from the running app. A docs-site `llms.txt` describes the framework in general; it can't tell an agent what routes *your* app exposes, what schemas *your* handlers accept, what error codes *your* code emits. That's the gap this issue is about.

**Why this belongs in Rapina specifically**

For a framework, the running app is the source of truth. Routes, schemas, and error codes exist in memory, reachable through introspection. Rapina already exposes `/__rapina/openapi.json` (`rapina/src/app.rs:797`) which is machine-readable, but OpenAPI is verbose JSON designed for tooling, not agents. An agent reads JSON Schema and spends tokens on the shape of the spec itself. Markdown at `/llms.txt` is what agents consume natively.

`RouteInfo` already captures everything we need: method, path, request and response JSON Schemas, content types, and error variants (`rapina/src/introspection/route_info.rs:22-44`). This issue is a Markdown serializer over the existing registry plus an endpoint to serve it.

**What we build**

Three pieces:

- `rapina::introspection::llms::to_llms_txt(&RouteRegistry) -> String`. No new deps.
- `GET /__rapina/llms.txt` returning `text/plain; charset=utf-8`. Generated once at startup into an `Arc<String>`, served on every request. Public by default.
- `rapina llms export [--output file]`, parallel to `rapina openapi export`.

Sketch of the output. One file, everything inlined. The index + detail split the spec allows is a docs-site concern; a running app should serve one canonical document.

```
# <project_name>

> <description from Cargo.toml>

Built with Rapina v<version>. Auth required unless marked public.

## Routes

### POST /v1/users
Create a new user. Requires JWT.

Request (application/json):
  { "email": "string", "name": "string" }

Response 201 (application/json):
  { "id": "number", "email": "string", "name": "string" }

Errors:
  - 409 CONFLICT: email is already registered
  - 422 VALIDATION_ERROR: request body failed validation

### GET /health [public]

Response 200: { "status": "ok" }

## Error envelope

  { "error": { "code": "string", "message": "string" }, "trace_id": "string" }
```

**Decisions I'd lock in**

- Path is `/__rapina/llms.txt`, not `/llms.txt`. Root-path mounting squats on user route space. Every other framework-served endpoint lives under `/__rapina/`. Anyone who wants it at root can re-expose with one line.
- On by default in dev, toggleable in `App::builder()` like the OpenAPI endpoint. Private APIs that don't want route metadata leaking turn it off entirely.
- Public. The value is being agent-accessible. Gating per-request makes no sense.
- JSON Schema rendered verbatim for v1. Compaction is cleaner but lossy on nested types; can add later.

**Scope: routes only.** Extractor conventions, error envelope as a general concept, and framework rules belong in the static AGENTS.md fragments from #521, not in a dynamic runtime endpoint. Keeping them separate means this endpoint stays trivially cacheable and the AGENTS.md fragments stay git-reviewable.

AGENTS.md should tell agents to fetch `/__rapina/llms.txt` when `rapina dev` is running. Closes the loop: agent starts the dev server, reads the live endpoint, knows every route without guessing. Tracking that piece under #521, not here.

**One gap worth naming**

`RouteInfo.handler_name` gives us `create_user` and nothing richer. Output will look thin for non-trivial routes until we have real handler descriptions. Either rustdoc extraction or a macro attribute, either way its own scope and its own issue. v1 ships thin, which is still better than nothing.

**v1**

- Generator in `rapina::introspection::llms`
- `/__rapina/llms.txt` endpoint
- `rapina llms export` CLI
- Docs page
- One test covering output shape

Later, separate issues: handler descriptions, `rapina routes --format llms` shorthand, compact schema mode.

**Further reading**

- [llms.txt spec](https://llmstxt.org/)
- [Answer.AI post introducing the proposal](https://www.answer.ai/posts/2024-09-03-llmstxt.html)
- [llmstxt.site directory of adopters](https://llmstxt.site/)
- Existing related issue in Rapina: #521 (AGENTS.md + bundled docs)

uemuradevexe:
## Summary

- Adds `GET /__rapina/llms.txt` endpoint serving a Markdown document (llms.txt convention) with all user-defined routes, request/response schemas, and error codes
- Enabled by default in debug builds (`cfg!(debug_assertions)`), off in release; controllable via `enable_llms_txt()` / `disable_llms_txt()` / `with_llms_txt(bool)` builder methods
- Adds `rapina llms export` CLI command to fetch and write the document from a running server
- Adds docs page at `/docs/core-concepts/llms-txt/` and CLI reference entry
- 6 unit tests + 6 integration tests
## Related Issues

Closes #522

## Checklist

- [x] `cargo fmt` passes
- [x] Tests pass


arferreira:
Gonna take a look on this tomorrow!

rapina/src/introspection/llms.rs
```diff
@@ -0,0 +1,160 @@
...
+    let mut out = String::new();
+
+    // Filter out internal /__rapina/* routes so the document only describes
+    // user-defined API surface.
+    let user_routes: Vec<&RouteInfo> = routes
+        .iter()
+        .filter(|r| !r.path.starts_with("/__rapina"))
+        .collect();
+
+    writeln!(out, "# API").unwrap();
```
arferreira: Okay, this heading is hardcoded to `# API`, but `Rapina` already takes `openapi_title` (default `"API"`, configurable), let's say we want to customize our OpenAPI title, this gets a mismatch: our OpenAPI spec says one thing, and llms.txt says another, Worth changing it.

rapina/src/introspection/llms.rs
```diff
@@ -0,0 +1,160 @@
...
+                }
+            }
+        }
+
+        if let Some(schema) = &route.response_schema {
+            writeln!(out, "\nResponse:").unwrap();
+            let pretty = serde_json::to_string_pretty(schema).unwrap_or_default();
+            for line in pretty.lines() {
+                writeln!(out, "  {}", line).unwrap();
+            }
```
arferreira: JSON Schema is written with a 2 space prefix indent. CommonMark needs 4 space indents for indented code blocks, so any Markdown renderer treats this as flowing prose, not code. For agents this is totally fine but for us previewing it's ugly lol, and for any agent that strips Markdown this becomes a wall of text.

arferreira: We could just wrap in fenced blocks instead.

rapina/src/introspection/llms.rs
```diff
@@ -0,0 +1,160 @@
...
+//! `/__rapina/llms.txt` so AI agents can discover all routes without
+//! scraping HTML or guessing.
+
+use std::fmt::Write as _;
+
+use crate::introspection::RouteInfo;
+
+/// Render the route registry as an llms.txt Markdown document.
+///
+/// The output is a single self-contained file (no index/detail split —
```
arferreira: Let's keep the doc focused on what the function does, this kind of concern is part only our discussion.

rapina/src/introspection/llms.rs
```diff
@@ -0,0 +1,160 @@
...
+/// that is a docs-site concern). JSON Schemas are rendered verbatim.
+pub fn to_llms_txt(routes: &[RouteInfo]) -> String {
+    let mut out = String::new();
+
+    // Filter out internal /__rapina/* routes so the document only describes
+    // user-defined API surface.
+    let user_routes: Vec<&RouteInfo> = routes
+        .iter()
+        .filter(|r| !r.path.starts_with("/__rapina"))
```
arferreira: Well, this is the second place that filters internal routes by prefix. `build_openapi_spec` does the same. Two places, same rule. Extract that `RouteInfo::is_internal()` or `RouteRegistry::user_routes()` and call it from both. Otherwise the next time we add an internal prefix, we'll forget one.

rapina/src/introspection/llms.rs
```diff
@@ -0,0 +1,160 @@
...
+            "internal routes must be filtered"
+        );
+    }
+
+    #[test]
+    fn test_to_llms_txt_empty_routes() {
+        let output = to_llms_txt(&[]);
+        assert!(output.contains("## Routes"));
+        assert!(!output.contains("###"));
+    }
```
arferreira: I think all of these assertions deserve a snapshot, so you can run a bless to make this better, a single snapshot of the full output against a fixture replaces all substrings checks, catches and so on.

rapina/src/app.rs
```diff
@@ -886,6 +913,15 @@ impl Rapina {
                 .get_named("/__rapina/routes", "list_routes", list_routes);
         }
 
+        if self.llms_txt {
+            let routes = self.router.routes();
```
arferreira, line 920: Worth a one line comment here:
```suggestion
            // Snapshot routes BEFORE registering /__rapina/llms.txt so the document
            // doesn't reference itself. The /__rapina/* filter in to_llms_txt is
            // defense-in-depth.
            let routes = self.router.routes();
```

rapina/src/introspection/llms.rs
```diff
@@ -0,0 +1,160 @@
...
+
+    writeln!(out, "# API").unwrap();
+    writeln!(out).unwrap();
+    writeln!(
+        out,
+        "Built with [Rapina](https://rapina.rs) v{}.",
+        env!("CARGO_PKG_VERSION")
+    )
+    .unwrap();
+    writeln!(out).unwrap();
```
arferreira: So many unwraps lol, `fmt::Write` is infallible, so they are safe, but they are a lot and this is in fact noisy, from the top of my head I'd suggest using a small macro or even a helper....
Something like this:
```rust
macro_rules! w { ($($t:tt)*) => { writeln!(out, $($t)*).unwrap() }; }
```

arferreira (changes requested):
Hey sheena, nice! We have some nits before merge.
Also two things worth filing as separate issues, not for this PR:
- `rapina openapi export` and now `rapina llms export` both shell out to curl. Adds an undocumented runtime dependency, drops error fidelity (HTTP status codes collapse to "curl exited non-zero"), no timeout control. The codebase should use a real HTTP client (`ureq` for blocking, `reqwest` if we need async) consistently across CLI fetchers. This wasn't introduced by this PR, but the pattern is worth fixing.

- `rapina doctor` doesn't check llms.txt configuration. With it on by default in debug and off in release, a user moving from rapina dev to a release build will see `/__rapina/llms.txt` start returning 404. A doctor warning would help.

arferreira (approved):
LGTM

PR #535, opened 2026-04-25
feat: ship version-matched bundled docs and AGENTS.md for AI agents

### LINKED ISSUE #521: Ship version-matched bundled docs for AI agents, with a thin `AGENTS.md` pointer
Agents working on Rapina projects today rely on training data that's out of date the moment we publish a release. Every breaking change in `rapina`, `sea_orm`, `axum`, or `tokio` is a footgun an LLM will happily walk into. The fix isn't a bigger AGENTS.md, it's putting the current, version-matched docs in a place the agent can always read.

Vercel ran this eval for Next.js 16.2. Their result: giving agents access to bundled documentation achieved a 100% pass rate on their internal evals, outperforming skills-based retrieval which maxed out at 79% ([source](https://vercel.com/blog/agents-md-outperforms-skills-in-our-agent-evals)). The lesson isn't "ship an AGENTS.md", it's "ship docs the agent can't miss." The AGENTS.md is three lines pointing at the bundled docs. That's the whole move.

For reference, [Next.js 16.2](https://nextjs.org/blog/next-16-2-ai) and [Phoenix 1.8](https://www.phoenixframework.org/blog/phoenix-1-8-released) both ship generator-side AGENTS.md today.

This issue is about doing the Rust-specific version of that, plus the concrete Rapina rules that peer frameworks don't have to write.

**What we ship**

1. **Version-matched docs bundled with the project.** `rapina new` drops a `.rapina-docs/` directory at the project root, populated from the exact `rapina = "x.y.z"` in `Cargo.toml`. Offline, deterministic, agent-readable.
2. **`AGENTS.md`** composed from feature-flagged fragments under `rapina-cli/templates/agents/`:
   - `core.md` — always on
   - `extractors.md` — always on
   - `errors.md` — always on (envelope shape, stable error codes)
   - `testing.md` — always on (test app pattern, assertion idioms)
   - `migrations.md` — when `--db` is set
   - `websocket.md` — when `websocket` feature is enabled
   - `jobs.md` — when jobs are scaffolded
3. **`CLAUDE.md`** = a one-line `@AGENTS.md`. Follow Next.js's pattern. If the spec forks later, Rapina follows, no prediction needed.

**Marker, version stamp, and hash**

The rules block is wrapped in:

```
<!-- BEGIN:rapina-agent-rules v0.8.2 sha256:abc123... -->
...
<!-- END:rapina-agent-rules -->
```

Pattern adapted from [Next.js's `BEGIN:nextjs-agent-rules` markers](https://nextjs.org/docs/app/guides/ai-agents).

The hash covers the rendered rules content between the markers, excluding the marker lines themselves. Version stamp and hash are independent: a version bump with identical content won't invalidate the hash, and a user edit inside the block won't masquerade as a version mismatch.

**Drift detection**

`rapina doctor` compares the installed Rapina version + fragments against the on-disk block. Three cases:

- Version changed, fragments changed, block hash matches what Rapina last wrote: clean refresh via `rapina doctor --fix-agents`.
- Block hash matches current bundled fragments: up to date, no action.
- Block hash matches neither (user edited inside the markers): **refuse to auto-fix**. Print a diff, tell the user to move custom rules outside the markers, offer `--force` for people who know what they're doing.

This is v1, not future work. It's ~a day once the hash stamp is in place.

**Rapina-specific rules (imperative, not descriptive)**

Peer frameworks cover tooling basics. Rapina's opinionated surface means agents need these:

- `State<T>` must be the first handler argument, before `Path`, `Json`, `Validated`
- Required: `#[public]` on any route that should not require JWT. Forgetting it returns 401 in prod.
- Prefer `Validated<T>` over `Json<T>`. `Validated<T>` returns 422 with field-level errors.
- Return typed errors with `trace_id`. Never return plain strings or `String`-bodied responses from handlers.
- Run `rapina doctor` before concluding the app is misconfigured
- Use `rapina migrate up`, not `sea-orm-cli`

`testing.md` covers:

- Use the Rapina test harness, not raw axum test helpers
- Assert on the error `code` field, not the human message
- Use `#[tokio::test]` for async handler tests
- Reset DB state between tests (pattern to be spelled out once #TBD lands)

Plus a "don't do X" section. Phoenix's [1.8 AGENTS.md fragments](https://github.com/phoenixframework/phoenix/commit/50ffaa5aa1c60503f01cd2107edd43f22435f9e7) use the same pattern and it's disproportionately effective:

- Don't import from `axum::` directly. Use `rapina::` re-exports.
- Don't hand-roll JSON error responses. Use the error envelope.
- Don't use `sqlx::query!`. The ORM layer exists for a reason.
- Don't `.unwrap()` in handlers. Use `?` with typed errors.
- Don't invoke `sea-orm-cli`. Use `rapina migrate`.

**Open design question: where do the bundled docs live?**

Next.js gets away with `node_modules/next/dist/docs/` because `npm install` reconstitutes it on every clone and every CI run. Rust doesn't have that guarantee — `target/` is gitignored and often empty on fresh clones. Three options:

- **A.** Ship docs inside the `rapina` crate, resolve the path at runtime via `cargo metadata`. Works like Next.js. Requires `cargo fetch` to have run.
- **B.** Drop `.rapina-docs/` into the project, commit it by default. Every clone and every CI run has the docs. Repo is larger.
- **C.** Gitignore `.rapina-docs/`, regenerate on demand via `rapina doctor` or a build script. Clean repos, but relies on users running the regen.

Proposed default: **B**. The whole point is reliability of agent context, and gitignoring the thing that provides that reliability is self-defeating. Users who want C can gitignore it themselves and add a `rapina doctor` step to CI. Open to pushback.

**Flags**

- `--no-agents-md` — skip AGENTS.md + CLAUDE.md generation (parity with Next.js)
- `--no-bundled-docs` — skip `.rapina-docs/`
- `--agents-md-only` — generate AGENTS.md + CLAUDE.md but skip bundled docs (hand-curated docs, or point at your own)

**Rollout**

v1:

- Fragment system + AGENTS.md composition
- `.rapina-docs/` population on `rapina new`
- Version + hash stamped rules block
- `rapina doctor` drift detection with the three-case UX above, and `--fix-agents` / `--force`
- CLAUDE.md pointer file
- Three flags above

Existing projects run `rapina doctor --fix-agents` to drop the files in.

ShiraiEd:
## Summary

Composes `AGENTS.md` from feature-flagged fragments (`core`, `extractors`, `errors`, `testing`, plus `migrations`/`websocket`/`jobs` when active), stamped with a version and SHA256 hash so drift can be detected reliably
- Drops `.rapina-docs/` into the project on `rapina new` — individual fragment files committed to the repo so agents always have current docs without relying on stale training data
- Adds `CLAUDE.md` as a pointer file (`@AGENTS.md`) so Claude Code picks up the rules automatically
- Adds `rapina doctor --fix-agents [--force]` to onboard existing projects and refresh stale blocks, refusing to overwrite user edits inside the markers without `--force`
- Adds `--no-agents-md`, `--no-bundled-docs`, `--agents-md-only` flags to `rapina new` for projects that manage their own docs

## Related Issues

Closes #521 

## Checklist

- [x] `cargo fmt` passes
- [x] Tests pass


rapina-cli/src/commands/agents.rs
```diff
@@ -0,0 +1,642 @@
...
+/// The BEGIN marker embeds the current CLI version and a SHA256 hash of `content`.
+/// The hash covers exactly the bytes between the markers, so drift detection can distinguish
+/// between a clean version bump (hash still matches) and a user edit (hash no longer matches).
+pub fn wrap_with_markers(content: &str) -> String {
+    let version = env!("CARGO_PKG_VERSION");
+    let hash = sha256_hex(content);
+    format!(
+        "<!-- BEGIN:rapina-agent-rules v{version} sha256:{hash} -->\n{content}\n<!-- END:rapina-agent-rules -->\n"
+    )
+}
```
richasdev, line 89: `sha256_hex` formats each byte individually with a collect. The `sha2` output type implements `LowerHex`, so this works:

```rust
pub fn sha256_hex(s: &str) -> String {
    let mut hasher = Sha256::new();
    hasher.update(s.as_bytes());
    format!("{:x}", hasher.finalize())
}
```

Same output, no iterator.

rapina-cli/src/commands/agents.rs
```diff
@@ -0,0 +1,642 @@
...
+pub fn check_drift(base: &Path) -> DriftStatus {
+    let source = match std::fs::read_to_string(base.join("AGENTS.md")) {
+        Ok(s) => s,
+        Err(_) => return DriftStatus::Missing,
+    };
+
+    let block = match parse_agents_block(&source) {
+        Some(b) => b,
+        None => return DriftStatus::NoBlock,
+    };
```
richasdev, line 262: `parse_agents_block` returns `None` for both "no block" and "BEGIN found but END missing". Both hit this arm and produce `DriftStatus::NoBlock`, so a corrupt file (half-written marker) shows the same message as a file with no block at all.

Suggest checking for the begin prefix before returning `NoBlock`:

```rust
let block = match parse_agents_block(&source) {
    Some(b) => b,
    None => {
        if source.contains("<!-- BEGIN:rapina-agent-rules ") {
            // BEGIN without END: corrupt or truncated file
            eprintln!("AGENTS.md has an unclosed marker. Run: rapina doctor --fix-agents --force");
        }
        return DriftStatus::NoBlock;
    }
};
```

uemuradevexe (changes requested):

rapina-cli/src/commands/agents.rs
```diff
@@ -0,0 +1,645 @@
...
+}
+
+/// Return the content to write to `CLAUDE.md`.
+///
+/// Includes the `@AGENTS.md` include directive (Claude Code CLI syntax) plus a brief
+/// human-readable header so the file is understandable outside Claude Code.
+pub fn generate_claude_md() -> &'static str {
+    "# Claude Rules\n\
+     \n\
+     This project uses [Rapina](https://rapina.dev), a Rust web framework.\n\
```
arferreira: Hmm wrong URL.
```suggestion
     This project uses [Rapina](https://userapina.com), a Rust web framework.\n\
```

rapina-cli/src/commands/agents.rs
```diff
@@ -0,0 +1,645 @@
...
+/// Rewrite the `rapina-agent-rules` block in `AGENTS.md` with current bundled content.
+///
+/// Preserves any text outside the markers (e.g. custom rules the user added above or below).
+/// Refuses to overwrite user edits inside the markers unless `force` is `true` — detected by
+/// comparing `sha256(on_disk_body)` against `stored_hash` in the marker.
+///
+/// Also creates `CLAUDE.md` (if absent) and refreshes `.rapina-docs/` to match the new block.
+pub fn fix_agents(base: &Path, force: bool) -> Result<(), String> {
+    let agents_path = base.join("AGENTS.md");
+    let source = std::fs::read_to_string(&agents_path).unwrap_or_default();
```
arferreira: `unwrap_or_default` masks every IO error as empty file, worth covering this with a match, "File doesn't exist" is a real fresh creation case. Other IO errors should fail loudly, not silently overwrite.

rapina-cli/src/commands/agents.rs
```diff
@@ -0,0 +1,645 @@
...
+    // Fall back to empty flags when Cargo.toml is absent or has no rapina dependency
+    // (e.g. running `rapina doctor --fix-agents` outside a project). The generated
+    // AGENTS.md will include only the always-on fragments; conditional ones are omitted.
+    let flags = super::verify_rapina_project()
+        .map(|cargo| detect_flags(&cargo))
+        .unwrap_or(AgentsFlags {
+            with_db: false,
+            with_websocket: false,
+            with_jobs: false,
+        });
```
arferreira: Lets require a `Cargo.toml` to exist when `fix_agents` runs in a real directory and only relax for tests.

rapina-cli/src/commands/agents.rs
```diff
@@ -0,0 +1,645 @@
...
+        .unwrap_or_default();
+
+    AgentsFlags {
+        with_db: features
+            .iter()
+            .any(|f| f == "sqlite" || f == "postgres" || f == "mysql"),
+        with_websocket: features.iter().any(|f| f == "websocket"),
+        with_jobs: features.iter().any(|f| f == "jobs"),
+    }
+}
```
arferreira, line 48: Reads `dependencies.rapina.features directly`. Doesn't handle `[workspace.dependencies]` inheritance, a workspace member with `rapina = { workspace = true, features  = ["postgres"] }` will silently report `with_db: false`. Worth a follow-up issue. Most users won't hit this, I will cuz usually I'm running monorepos.

rapina-cli/src/commands/agents.rs
```diff
@@ -0,0 +1,645 @@
...
+        let result = std::fs::read_to_string(dir.path().join("AGENTS.md")).unwrap();
+        assert!(
+            result.starts_with("# Custom header\n"),
+            "header lost: {result}"
+        );
+        assert!(
+            result.contains("## Custom footer\n"),
+            "footer lost: {result}"
+        );
+    }
```
arferreira: The test name says "Stale" but it actually exercises fresh fix_agents on a file that already has a current block.

rapina-cli/src/commands/agents.rs
```diff
@@ -0,0 +1,645 @@
...
+
+    // Create CLAUDE.md only if absent — never overwrite user customisations
+    let claude_path = base.join("CLAUDE.md");
+    if !claude_path.exists() {
+        std::fs::write(&claude_path, generate_claude_md())
+            .map_err(|e| format!("Failed to write CLAUDE.md: {}", e))?;
+    }
+
+    // Refresh .rapina-docs/ to match the regenerated block
+    generate_rapina_docs(base, &flags)?;
```
arferreira, line 399: This regenerates `.rapina-docs/` on every call, even on `UpToDate`, worth a follow-up for improvement either skip when content matches for example.

arferreira (changes requested):
This is going very well, I left some suggestions also questions we could address before merge.

uemuradevexe (commented):
Three issues flagged via inline comments below. See individual comments for details.

rapina-cli/src/commands/agents.rs
```diff
@@ -0,0 +1,721 @@
...
+
+// ── Marker parsing ────────────────────────────────────────────────────────────
+
+/// The parsed contents of a `rapina-agent-rules` marker block in `AGENTS.md`.
+pub struct ParsedBlock {
+    /// CLI version recorded when the block was last written (e.g. `"0.11.0"`).
+    /// Retained in the struct so callers can inspect it for debugging, but drift
+    /// detection uses only the hash — version bumps alone don't trigger Stale.
+    #[allow(dead_code)]
+    pub stored_version: String,
```
richasdev, line 148: Dead field with `#[allow(dead_code)]`. The `Stale` path is exactly where this would be useful showing "was v{stored_version}, now v{current}" gives the user context. Either surface it there or remove the field.

rapina-cli/src/commands/agents.rs
```diff
@@ -0,0 +1,721 @@
...
+    let flags = match std::fs::read_to_string(&cargo_path) {
+        // Cargo.toml found — detect which optional fragments to include.
+        // If the file can't be parsed, fall back to always-on fragments only.
+        Ok(content) => toml::from_str::<toml::Value>(&content)
+            .map(|parsed| detect_flags(&parsed))
+            .unwrap_or(AgentsFlags {
+                with_db: false,
+                with_websocket: false,
+                with_jobs: false,
+            }),
```
richasdev, line 345: Silent fallback: if `Cargo.toml` exists but is malformed, this generates `AGENTS.md` without optional fragments (no `migrations.md`, etc.) with no warning emitted.

Suggestion:
```rust
.unwrap_or_else(|e| {
    eprintln!("Warning: could not parse Cargo.toml ({e}), generating base fragments only");
    AgentsFlags { with_db: false, with_websocket: false, with_jobs: false }
})
```

rapina-cli/src/commands/agents.rs
```diff
@@ -0,0 +1,721 @@
...
+    Ok(())
+}
+
+/// Produce a simple line-level diff between `old` and `new` for human display.
+///
+/// Lines present only in `old` are prefixed with `- `; lines present only in `new`
+/// with `+ `. Lines present in both (accounting for duplicates) are omitted. This is not a
+/// true LCS diff — it uses count-based multiset difference, sufficient for `AGENTS.md` drift display.
+pub fn simple_diff(old: &str, new: &str) -> String {
+    use std::collections::HashMap;
```
richasdev, line 410: All removals are printed first (old-line order), then all additions. For a multi-section `AGENTS.md` this looks like "everything deleted and replaced" rather than a localized change. This is the one place where the user needs to understand *what* they edited before deciding to move content outside the markers.

Doesn't need a full LCS even a header like `(grouped: removals first, then additions)` would help set expectations.

uemuradevexe (changes requested):

rapina-cli/src/commands/agents.rs
arferreira: Sorry, you marked as resolved, but link's still wrong.

arferreira (changes requested):
One thing slipped: URL at agents file is still wrong.

arferreira (approved):
LGTM

richasdev (approved):
LGTM

PR #536, opened 2026-04-26
Support streaming and SSE responses

### LINKED ISSUE #326: Support streaming and SSE responses

arferreira:
Closes #326

Replaces the `BoxBody = Full<Bytes>` alias with a real boxed `http_body::Body` trait object so streaming and buffered responses share one response type. Adds `StreamResponse` for raw chunked bytes and `SseResponse` plus `SseEvent` for Server-Sent Events. Keep-alive is implemented as a custom `Body` that owns a `tokio::time::Sleep`, no `tokio-stream` dep, no separate task. Same shape axum's `KeepAlive` uses. Compression middleware now per-chunk-encodes streaming bodies with a persistent `flate2` encoder and unconditionally skips `text/event-stream`, matching Starlette's commit a9a8dab from Feb 2025 (gzip across SSE event boundaries breaks framing at proxies). Cache middleware checks `body.size_hint().exact().is_none()` and skips streaming responses since they can't be captured as a single `Bytes` blob. Breaking change for users constructing `Response<BoxBody>` manually with `Full::new(...)`. Replace with `rapina::response::full(...)`. `BoxBody::default()` becomes `rapina::response::empty()`.

### DIFF
--- a/rapina/Cargo.toml
+++ b/rapina/Cargo.toml
@@ -27,6 +27,8 @@ hyper-util = { version = "0.1.19", features = [
 http = "1.4.0"
 http-body-util = "0.1.3"
 bytes = "1.11.1"
+futures-core = { version = "0.3", default-features = false, features = ["std"] }
+pin-project-lite = "0.2"
 
 # Serialization
 serde = { version = "1.0.228", features = ["derive"] }
@@ -103,6 +105,7 @@ tokio-cron-scheduler = { version = "0.15.1", optional = true }
 tokio-util = { version = "0.7.18", optional = true }
 
 [dev-dependencies]
+async-stream = "0.3"
 criterion = { version = "0.5", features = ["html_reports"] }
 insta = "1"
 matchit = "0.9"
--- a/rapina/src/cache.rs
+++ b/rapina/src/cache.rs
@@ -26,13 +26,13 @@ use std::time::{Duration, Instant};
 use bytes::Bytes;
 use dashmap::DashMap;
 use http::{Response, header};
-use http_body_util::{BodyExt, Full};
+use http_body_util::BodyExt;
 use hyper::Request;
-use hyper::body::Incoming;
+use hyper::body::{Body, Incoming};
 
 use crate::context::RequestContext;
 use crate::middleware::{BoxFuture, Middleware, Next};
-use crate::response::BoxBody;
+use crate::response::{BoxBody, empty, full};
 
 /// Internal header injected by the `#[cache(ttl = N)]` macro.
 /// The middleware reads this to determine caching behavior, then strips it.
@@ -257,13 +257,26 @@ impl Middleware for CacheMiddleware {
 
                 // Check if handler wants caching
                 if let Some(ttl) = extract_ttl_header(&response) {
+                    // Streaming bodies cannot be captured as a single Bytes blob.
+                    // A handler asking to cache one is a programming error: log
+                    // a warning so the operator notices the #[cache] is silently
+                    // being ignored, then pass the response through unchanged.
+                    if response.body().size_hint().exact().is_none() {
+                        tracing::warn!(
+                            path = %path,
+                            ttl,
+                            "cache middleware: streaming response cannot be cached, \
+                             passthrough. Remove #[cache] from this handler."
+                        );
+                        return response;
+                    }
                     let (parts, body) = response.into_parts();
                     let body_bytes = match body.collect().await {
                         Ok(collected) => collected.to_bytes(),
                         Err(_) => {
                             return Response::builder()
                                 .status(http::StatusCode::INTERNAL_SERVER_ERROR)
-                                .body(Full::new(Bytes::new()))
+                                .body(empty())
                                 .unwrap();
                         }
                     };
@@ -288,7 +301,7 @@ impl Middleware for CacheMiddleware {
                         .await;
 
                     // Return response without the internal header, with MISS marker
-                    let mut response = Response::from_parts(parts, Full::new(body_bytes));
+                    let mut response = Response::from_parts(parts, full(body_bytes));
                     response.headers_mut().remove(CACHE_TTL_HEADER);
                     response
                         .headers_mut()
@@ -362,7 +375,7 @@ fn build_response_from_cache(cached: CachedResponse, status: &'static str) -> Re
         }
     }
 
-    let mut response = builder.body(Full::new(cached.body)).unwrap();
+    let mut response = builder.body(full(cached.body)).unwrap();
 
     response
         .headers_mut()
--- a/rapina/src/middleware/compression.rs
+++ b/rapina/src/middleware/compression.rs
@@ -1,15 +1,18 @@
 use std::io::Write;
+use std::pin::Pin;
+use std::task::{Context, Poll};
 
 use bytes::Bytes;
 use flate2::Compression;
 use flate2::write::{DeflateEncoder, GzEncoder};
 use http::{HeaderValue, Response, header};
-use http_body_util::{BodyExt, Full};
+use http_body_util::BodyExt;
 use hyper::Request;
-use hyper::body::Incoming;
+use hyper::body::{Body, Frame, Incoming, SizeHint};
+use pin_project_lite::pin_project;
 
 use crate::context::RequestContext;
-use crate::response::{APPLICATION_JSON, BoxBody};
+use crate::response::{APPLICATION_JSON, BodyError, BoxBody, empty, full};
 
 use super::{BoxFuture, Middleware, Next};
 
@@ -153,6 +156,19 @@ impl Middleware for CompressionMiddleware {
 
             let response = next.run(req).await;
 
+            // SSE bodies must never be compressed: gzip across event boundaries
+            // breaks framing at proxies. See Starlette's GZipMiddleware (commit
+            // a9a8dab, Feb 2025) for the same exclusion.
+            let is_event_stream = response
+                .headers()
+                .get(header::CONTENT_TYPE)
+                .and_then(|v| v.to_str().ok())
+                .map(|s| s.starts_with("text/event-stream"))
+                .unwrap_or(false);
+            if is_event_stream {
+                return response;
+            }
+
             let algorithm = match algorithm {
                 Some(alg)
                     if !Self::is_already_encoded(&response)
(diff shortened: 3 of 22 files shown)


rapina/src/cache.rs
```diff
@@ -257,13 +257,19 @@ impl Middleware for CacheMiddleware {
 
                 // Check if handler wants caching
                 if let Some(ttl) = extract_ttl_header(&response) {
+                    // Streaming bodies cannot be captured as a single Bytes blob.
+                    // A handler asking to cache one is a programming error; pass
+                    // the response through unchanged.
+                    if response.body().size_hint().exact().is_none() {
+                        return response;
```
ShiraiEd, line 271: Silent passthrough means users will never know their #[cache]
  annotation is being ignored. A `tracing::warn!` here would surface this immediately:
```rust
  tracing::warn!(
      "cache middleware: streaming response cannot be cached, \
       passthrough. Remove #[cache] from this handler."
  );
```

rapina/tests/streaming.rs
ShiraiEd, line 1: No test for keep-alive frame emission

  Every SSE test calls `.keep_alive(None)` to disable the timer. There is no test that verifies the `:\n\n` frame actually arrives when the user stream is
  idle. This is the most novel logic in the PR and has zero test coverage.

rapina/src/response/stream.rs
```diff
@@ -0,0 +1,416 @@
...
+            &bytes[..],
+            b"event: update\nid: 7\nretry: 1000\ndata: payload\n\n"
+        );
+    }
+
+    #[test]
+    fn test_keep_alive_frame_format() {
+        // The wire format the SSE spec defines for a no-op comment.
+        assert_eq!(SSE_KEEP_ALIVE_FRAME, b":\n\n");
+    }
```
ShiraiEd, line 622: This test only asserts that the constant `SSE_KEEP_ALIVE_FRAME == b":\n\n"`,  it's effectively `assert_eq!(b":\n\n", b":\n\n")`. 
it proves nothing about whether the frame is ever actually sent.
Can it be removed, or is there a reason to keep it?

rapina/src/response/stream.rs
```diff
@@ -0,0 +1,416 @@
...
+    pub fn status(mut self, status: StatusCode) -> Self {
+        self.status = status;
+        self
+    }
+
+    pub fn content_type(mut self, content_type: &'static str) -> Self {
+        self.content_type = HeaderValue::from_static(content_type);
+        self
+    }
+}
```
ShiraiEd, line 119: `content_type()` accepts `&'static str`, but the content type can vary per request (e.g. a user downloading report.pdf vs video.mp4). Since the struct already stores a `HeaderValue` internally, a solution would be for the setter to accept `HeaderValue` directly and let the caller construct it from a static string or a runtime value  depending on their use case.

They would do it at the call site:

  // static — known at compile time
  StreamResponse::new(s).content_type(HeaderValue::from_static("application/pdf"))

  // dynamic — known only at request time
  let mime = mime_from_ext(&ext); // from a match table → &'static str
  StreamResponse::new(s).content_type(HeaderValue::from_static(mime))

  // truly dynamic string
  let mime = format!("application/{}", ext);
  StreamResponse::new(s).content_type(HeaderValue::from_str(&mime).unwrap())

  The user picks whichever construction method fits their case. The setter just receives the already-built HeaderValue and stores it.

What do you think?

rapina/src/response/stream.rs
```diff
@@ -0,0 +1,416 @@
...
+        let stream = unsafe { Pin::new_unchecked(&mut this.stream) };
+
+        match stream.poll_next(cx) {
+            Poll::Ready(Some(Ok(event))) => {
+                if let Some(ka) = this.keep_alive.as_mut() {
+                    ka.reset();
+                }
+                Poll::Ready(Some(Ok(Frame::data(event.encode()))))
+            }
+            Poll::Ready(Some(Err(e))) => Poll::Ready(Some(Err(e))),
```
ShiraiEd: Producer errors are silently propagated and the connection closes on the client side but nothing is logged on the server.

It returns the error up to hyper, which then closes the connection, no log, no trace, nothing printed on the server side.

The error is just silently swallowed from the operator's perspective. The client sees the connection drop, the server sees nothing.

A `tracing::error!` here would help operators diagnose mid-stream failures.

Also, no test covers this path, a test should yield `Err(BodyError)` from the producer and assert the connection closes cleanly rather than hanging or panicking.

rapina/src/response/stream.rs
```diff
@@ -0,0 +1,416 @@
...
+    type Error = BodyError;
+
+    fn poll_frame(
+        self: Pin<&mut Self>,
+        cx: &mut Context<'_>,
+    ) -> Poll<Option<Result<Frame<Bytes>, BodyError>>> {
+        // Safety: stream and keep_alive are never moved out; structural pin
+        // projection on stream so Stream::poll_next can run.
+        let this = unsafe { self.get_unchecked_mut() };
+        let stream = unsafe { Pin::new_unchecked(&mut this.stream) };
```
ShiraiEd: `FrameStream` and `SseBody` both use manual unsafe pin projection. `pin-project-lite` is already in the dependency tree (pulled transitively by `tokio`) so using it would add no new dependencies and eliminate both unsafe blocks at zero cost.

rapina/Cargo.toml
```diff
@@ -103,6 +104,7 @@ tokio-cron-scheduler = { version = "0.15.1", optional = true }
 tokio-util = { version = "0.7.18", optional = true }
 
 [dev-dependencies]
+async-stream = "0.3"
```
ShiraiEd, line 108: A user reading the docs and copying that example would get a compile error because async-stream isn't in their [dependencies]. The docs don't mention they need to add it themselves.

rapina/src/middleware/compression.rs
```diff
@@ -201,6 +234,134 @@ impl Middleware for CompressionMiddleware {
...
+    }
+
+    /// Take whatever output bytes the encoder has buffered so far. Empty if
+    /// the encoder is still accumulating internally.
+    fn drain(&mut self) -> Bytes {
+        let buf = match self {
+            Self::Gzip(e) => e.get_mut(),
+            Self::Deflate(e) => e.get_mut(),
+        };
+        Bytes::from(std::mem::take(buf))
```
ShiraiEd: *NITT*

Maybe this would improve the performance a little 

drain() uses std::mem::take which resets the encoder's internal buffer to zero capacity after every chunk, causing a fresh heap allocation on each chunk:

  chunk 1 → allocate (0 → N), fill, drain → reset to 0 capacity
  chunk 2 → allocate (0 → N), fill, drain → reset to 0 capacity
  chunk 3 → allocate (0 → N), fill, drain → reset to 0 capacity

  Using mem::replace with the existing capacity would reuse the buffer after the first chunk:

  chunk 1 → allocate (0 → N), fill, drain → keeps capacity N
  chunk 2 → reuse capacity, fill, drain → keeps capacity N
  chunk 3 → reuse capacity, fill, drain → keeps capacity N

  For long streams this is one allocation vs one per chunk. Suggested fix:

```rust
  fn drain(&mut self) -> Bytes {
      let buf = match self {
          Self::Gzip(e) => e.get_mut(),
          Self::Deflate(e) => e.get_mut(),
      };
      let cap = buf.capacity();
      Bytes::from(std::mem::replace(buf, Vec::with_capacity(cap)))
  }
```

what do you think

ShiraiEd (commented):
Hi, I have left some comments on the code

I have left comments on the following:

  Unsafe pin projection: pin-project-lite is already in the dependency tree, no reason to use manual unsafe

  Silent cache bypass: streaming responses silently skip #[cache] with no warning to the operator

  content_type() signature: accepts &'static str but should accept HeaderValue directly to support runtime values

  Keep-alive untested: test_keep_alive_frame_format only checks the constant value; all SSE integration tests disable keep-alive via .keep_alive(None)

  Mid-stream errors invisible: Poll::Ready(Some(Err(e))) propagates silently with no server-side log; also no test covering this path

  drain() allocation per chunk: std::mem::take resets buffer capacity on every chunk; mem::replace would reuse it

  async-stream missing from docs: used in all examples but only a dev-dependency; users need to add it themselves

arferreira:
Hey I just addressed those points, also while I was there I realized that it was a bit hard for using so I dropped the generic parameter on `StreamResponse` and `SseResponse` now users can return them directly like:

```rust
async fn handler() -> SseResponse
```
Also some other improvements you will see on review.

rapina/src/response/stream.rs
```diff
@@ -0,0 +1,826 @@
...
+        let frames = FrameStream { inner: self.stream };
+        let body = StreamBody::new(frames).boxed_unsync();
+        let mut builder = Response::builder()
+            .status(self.status)
+            .header(header::CONTENT_TYPE, self.content_type);
+        if let Some(h) = builder.headers_mut() {
+            for (name, value) in self.extra_headers.into_iter().filter_map(|(n, v)| {
+                let n = n?;
+                Some((n, v))
+            }) {
```
ShiraiEd: When 'n = None', filter_map drops the value instead of associating it with the previous 'HeaderName'.

The way HeaderMap works it is by using a Vec with values and indices forming a linked list, and when passing a (None, value) into it, it appends the value into the last stored HeaderName

A simple solution is to use the 'extend()' method instead of the filter because it supports (Option<HeaderName>, HeaderValue) and handles 'None' correctly by keeping track of the last used 'HeaderName'

```rust
h.extend(self.extra_headers);
```

ShiraiEd (commented):
Great, everything was covered

I just realized that I have missed something in stream.rs on the last review, left a comment there

ShiraiEd (approved):
LGTM

Issue #518, opened 2026-04-20
[Suggestion]: Add more sophisticated path normalization in metrics middleware


juv:
### Suggestion

To avoid cardinality explosion, the metrics middleware already normalizes request paths. The current normalization is based on splitting the full request path on `/` characters and then in case all characters of the subpath are numbers (`0..9`), the subpath is rewritten as `:id`.

This works fine in case the endpoint is always called with numbers, e.g.
`http://127.0.0.1:3000/endpoint/1`
`http://127.0.0.1:3000/endpoint/2`
`http://127.0.0.1:3000/endpoint/3`
`http://127.0.0.1:3000/endpoint/67`

The `/metrics` endpoint will show this: 
`http_request_duration_seconds_count{method="GET",path="/endpoint/:id"} 4`

However, in many real world backends, path params do not only include numbers but also uuids and arbitrary strings.

`http://127.0.0.1:3000/endpoint/e58ed763-928c-4155-bee9-fdbaaadc15f6`
`http://127.0.0.1:3000/endpoint/whatever`

The cardinality explosion adds up quickly in a production backend.
In terms of developer experience, it would be best if the request path normalization can re-use the original route definition, i.e. `/endpoint/:name` instead of hardcoded normalization logic. Hardcoded rewrites can lead to confusion when inspecting the `/metrics` output and not being able to find the matching route definitions.

### Context

Anything that currently is not a number will lead to cardinality explosion in the metrics middleware, which adds up quickly:
```
http_request_duration_seconds_bucket{method="GET",path="/endpoint/e58ed763-928c-4155-bee9-fdbaaadc15f3",le="0.005"} 1
http_request_duration_seconds_bucket{method="GET",path="/endpoint/e58ed763-928c-4155-bee9-fdbaaadc15f3",le="0.01"} 1
http_request_duration_seconds_bucket{method="GET",path="/endpoint/e58ed763-928c-4155-bee9-fdbaaadc15f3",le="0.025"} 1
http_request_duration_seconds_bucket{method="GET",path="/endpoint/e58ed763-928c-4155-bee9-fdbaaadc15f3",le="0.05"} 1
http_request_duration_seconds_bucket{method="GET",path="/endpoint/e58ed763-928c-4155-bee9-fdbaaadc15f3",le="0.1"} 1
http_request_duration_seconds_bucket{method="GET",path="/endpoint/e58ed763-928c-4155-bee9-fdbaaadc15f3",le="0.25"} 1
[...]

PR #537, opened 2026-04-27
Invalidate all ancestor collection prefixes on mutation

### LINKED ISSUE #534: bug: cache invalidation prefix strips too many path segments for nested routes
## Description

`build_invalidation_prefix` strips the last path segment to derive the invalidation prefix, but for nested routes this leaves parent collection caches stale after mutations.

## Affected Code

`rapina/src/cache.rs` — `build_invalidation_prefix`

```rust
fn build_invalidation_prefix(path: &str) -> String {
    let base = path
        .rfind('/')
        .filter(|&i| i > 0)
        .map(|i| &path[..i])
        .unwrap_or(path);
    format!("GET:{}", base)
}
```

## Expected vs Actual Behavior

Given a `POST /users` request (successful), the middleware invalidates `GET:/users` — correct.

Given a `POST /users/123/posts` request (successful):
- **Expected:** invalidates `GET:/users/123/posts` **and** `GET:/users` (parent collection)
- **Actual:** invalidates only `GET:/users/123` — the `/users` list cache remains stale

The function only strips one level, so mutations on deeply nested resources never propagate up to ancestor collection caches.

## Reproduction

```rust
// Cache both endpoints
cache.set("GET:/users", ..., ttl).await;
cache.set("GET:/users/123/posts", ..., ttl).await;

// Simulate POST /users/123/posts → triggers invalidation
let prefix = build_invalidation_prefix("/users/123/posts");
// prefix = "GET:/users/123" — does NOT match "GET:/users"
cache.invalidate_prefix(&prefix).await;

// GET:/users is still cached (stale)
assert!(cache.get("GET:/users").await.is_some()); // should be None
```

## Environment

- Rapina: current main
- Affects: `InMemoryCache` and any `CacheBackend` implementation relying on `build_invalidation_prefix`

richasdev:
## Summary

- Replace build_invalidation_prefix with build_invalidation_prefixes
- Collect root collection + nested collections (skip ID segments)
- is_id_segment detects numeric and UUID-shaped path params
- Middleware iterates all prefixes, calls invalidate_prefix for each
## Related Issues

Closes #534 

## Checklist

- [x] `cargo fmt` passes
- [x] Tests pass


rapina/src/cache.rs
```diff
@@ -324,15 +325,64 @@ fn build_cache_key(path: &str, query: &str) -> String {
...
+    let full = format!("GET:{}", path);
+    if !prefixes.contains(&full) {
+        prefixes.push(full);
+    }
+
+    prefixes
+}
+
+/// Heuristic: segment looks like an identifier (numeric or UUID-shaped).
+fn is_id_segment(segment: &str) -> bool {
```
juv: This looks like it’s trying to solve a very similar thing,  which also causes  #518 : our middleware has no access to the handler pattern that was matched for the request. It would allow to stop making inconsistent assumptions about how to split the full path into its segments. 

With a consistent access to the handler pattern we could more easily purge the upper collections imho. And also solve the problem in the Prometheus metrics aggregation code afterwards. We could keep some static reference to the matched handler pattern and reuse it multiple times. What do you think?

rapina/src/cache.rs
```diff
@@ -317,8 +317,9 @@ impl Middleware for CacheMiddleware {
 
             // Auto-invalidate on successful mutations
             if is_mutation(&method) && response.status().is_success() {
-                let prefix = build_invalidation_prefix(&path);
-                self.backend.invalidate_prefix(&prefix).await;
+                for prefix in build_invalidation_prefixes(&path) {
+                    self.backend.invalidate_prefix(&prefix).await;
```
arferreira: Each `invalidate_prefix` here is awaited serially. For the Redis backend, each call runs `SCAN` + `DEL` over the keyspace, so an N prefix path costs N round trips for what is, under the current `starts_with` semantics, one prefix's worth of actual work. If we keep the multi prefix model after that's resolved, this should at minimum be `futures::future::join_all` over the futures rather than a sequential await loop.

rapina/src/cache.rs
```diff
@@ -337,15 +338,64 @@ fn build_cache_key(path: &str, query: &str) -> String {
...
-fn build_invalidation_prefix(path: &str) -> String {
-    // /users/123 -> invalidate GET:/users
-    // /users -> invalidate GET:/users
-    let base = path
-        .rfind('/')
-        .filter(|&i| i > 0)
-        .map(|i| &path[..i])
-        .unwrap_or(path);
-    format!("GET:{}", base)
+fn build_invalidation_prefixes(path: &str) -> Vec<String> {
```
arferreira: I think the walk and the `is_id_segment` heuristic aren't actually load-bearing here. `InMemoryCache::invalidate_prefix` is implemented as `key.starts_with(prefix)`, so the very first prefix this function emits, the root collection, e.g. `GET:/users`  already invalidates everything underneath it via prefix matching, including all the nested prefixes computed afterward in this loop. The function is doing precise work that the backend immediately discards.

rapina/src/cache.rs
```diff
@@ -337,15 +338,64 @@ fn build_cache_key(path: &str, query: &str) -> String {
...
+
+    for (i, segment) in segments.iter().enumerate() {
+        current.push('/');
+        current.push_str(segment);
+
+        let is_id = is_id_segment(segment);
+
+        if i == 0 {
+            // First segment is always a collection root
+            prefixes.push(format!("GET:{}", current));
```
arferreira: This assumption breaks under common API mounting patterns. For `POST /api/v1/users`, `i == 0` is `api`, so we emit `GET:/api`; `v1` and `users` then fall through the `!is_id && prev_was_id` arm with `prev_was_id == false`, so neither emits a prefix. The current `starts_with` semantics mask this (everything under `/api` gets clobbered, including `/api/v1/users`), but if invalidation ever becomes precise, both the `/api/v1` and `/api/v1/users` list caches go stale.

rapina/src/cache.rs
```diff
@@ -337,15 +338,64 @@ fn build_cache_key(path: &str, query: &str) -> String {
...
+            // Nested collection after an ID: /users/123/posts
+            prefixes.push(format!("GET:{}", current));
+        }
+
+        prev_was_id = is_id;
+    }
+
+    // Always include the path itself as a prefix (handles the direct collection case)
+    let full = format!("GET:{}", path);
+    if !prefixes.contains(&full) {
```
arferreira: Moot if the function collapses per the comment on `fn build_invalidation_prefixes`, but as is `prefixes.contains` is O(n) linear search on every call. By construction `current` after the loop equals `path`, so checking `prefixes.last() == Some(&full)` would do the same job in constant time.

rapina/src/cache.rs
arferreira: This treats the segment as a collection name, and the parent prefix is never emitted. The fundamental issue is that path shape heuristics can't distinguish identifiers from collection names without knowing the route pattern, which is exactly @juv earlier pointed.

arferreira (changes requested):
Good work so far! I'd push back on shipping this shape. I left some comments for the matched route pattern to be exposed to the middleware. That's the design Phoenix and FastAPI both ended up at, and it would also close #518 as Juv's earlier commented as well.

arferreira (approved):
LGTM

PR #538, opened 2026-04-28
Add custom Prometheus collector support via add_metric()

### LINKED ISSUE #412: [Feature]: Custom Prometheus metrics
### Problem Statement

The [Prometheus metrics covered by Rapina](https://userapina.com/docs/core-concepts/metrics/) are a great start. 
There are three metrics being exported in the Prometheus endpoint at the time of this writing.

When developing production-ready backends, custom metrics are often a must have.
To accomplish this, I need to implement a second metrics endpoint to export the custom metrics of my backend.

### Proposed Solution

I'd like to mix in my custom Prometheus metrics into the Rapina-provided metrics endpoint.
These metrics should be collected after the standard Rapina metrics and exported in one go.

### Alternatives Considered

_No response_

### Contribution

richasdev:
## Summary

- src/metrics/prometheus.rs — new_with_collectors() constructor registers custom Box<dyn Collector> alongside built-ins; new() delegates to it; 4 unit tests added
- src/app.rs — custom_metrics field + add_metric() builder method; prepare() passes collectors to registry; 3 unit tests added
- src/lib.rs — pub use prometheus; re-export (feature-gated) so users access rapina::prometheus::IntCounter etc.
- tests/metrics.rs — 4 integration tests: single custom metric, labeled metric, multiple metrics, coexistence with built-ins

## Related Issues

Closes #412 

## Checklist

- [x] `cargo fmt` passes
- [x] Tests pass


rapina/src/app.rs
```diff
@@ -689,6 +694,37 @@ impl Rapina {
...
+    /// )
+    /// .unwrap();
+    ///
+    /// Rapina::new()
+    ///     .enable_metrics()
+    ///     .add_metric(Box::new(orders_total.clone()))
+    ///     .listen("127.0.0.1:3000");
+    /// ```
+    #[cfg(feature = "metrics")]
+    pub fn add_metric(mut self, collector: Box<dyn prometheus::core::Collector>) -> Self {
```
juv: Open for discussion: Does the repeated call on .add_metric() feel good enough in terms of DX for production grade backends?

I’m afraid that it’s good for one to a handful of metrics, everything beyond that kind of creates a mess? What do you think?

rapina/src/app.rs
juv, line 1: Docs are missing for this new feature

rapina/src/metrics/prometheus.rs
```diff
@@ -58,6 +63,12 @@ impl MetricsRegistry {
             .register(Box::new(http_requests_in_flight.clone()))
             .expect("failed to register http_requests_in_flight");
 
+        for collector in collectors {
```
juv, line 64: A subtle but important edge case is metric name collisions with the built‑in metrics. 

I think the current implementation will panic on duplicate registration, which is technically correct but should be clearly documented so users know to avoid clashing names and labels at least.

rapina/src/app.rs
juv: It might be possible to reduce Box-noise here for the callers, and let our code Box the parameter before adding it to the registry. It will enforce a static‘ lifetime though - but this might already be implicitly enforced anyways?

juv: Users might expect the existing .discover() functionality to also handle the registration of custom metrics. I personally would also be a fan of that instead of manually registering metrics. Mixing both things (having macros for the endpoint handlers and registering metrics manually) feels a bit inconvenient. What do you think?

juv (changes requested):

rapina/src/app.rs
ShiraiEd: It would be a nice feat having them registered with just a `#[metrics]` for example,

It would be just like the Router::new()

Another nice feat would be a '#[derive(GroupMetrics)]' to group mutiple related metrics on structs,
if we add this it will be better to just make a new issue about this

ShiraiEd (commented):
It is just missing a test to check the metrics names collision

maybe @juv 's .discover() suggestion should be left to another issue

rapina/src/app.rs
juv: I've created https://github.com/rapina-rs/rapina/issues/541 to track it

juv:
> It is just missing a test to check the metrics names collision
> 
> maybe @juv 's .discover() suggestion should be left to another issue

+1 on this

arferreira:
Hey guys, is this completed?

ShiraiEd:
Is just missing the test that I suggested, but other than that it looks good

ShiraiEd (approved):
LGTM

arferreira (approved):
LGTM

juv (approved):
LGTM

PR #540, opened 2026-05-01
Add cursor-based pagination

### LINKED ISSUE #456: [Feature]: Cursor-based Pagination
### Problem Statement

The recent addition of built-in pagination in [PR #267](https://github.com/rapina-rs/rapina/pull/267) introduced the offset/limit pagination. A cursor-based pagination on top of that would improve API efficiency and flexibility and is well-suited for high performance APIs

### Proposed Solution

Introduce a cursor-based pagination system in addition to the existing limit/offset implementation. 

- Provide a new `CursorPaginate` extractor (or similar), allowing endpoints to paginate based on unique record tokens rather than simple offsets.
- Update or extend the `Paginated<T>` response type to handle cursor serialization alongside page results.
- Potential API usage:

```rust
#[get("/users")]
async fn list_users(db: Db, cursor: CursorPaginate) -> Result<Paginated<UserResponse>> {
    Ok(cursor.exec(User::find(), db.conn()).await?.map(UserResponse::from))
}
```
- Support next/previous navigation, make it easy to use for both clients and backend code.
- Documentation and examples for adopting cursor pagination in endpoints.

### Alternatives Considered

Alternative approaches include sticking with the current limit/offset pagination, or implementing custom cursor logic per endpoint rather than as a shared core feature. However, a built-in solution provides consistency and reuse across Rapina APIs.

### Contribution

arferreira:
Closes #456

Adds cursor-based pagination as a sibling to the existing offset/limit primitive. Built on SeaORM's `cursor_by` with opaque URL-safe base64(JSON) tokens, Stripe-style `?after=&before=&limit=` semantics, no count query, no shared response shape. Lands behind the `database` feature alongside `Paginate`.

Handlers stay terse via a `CursorKey` trait that declares the entity's default cursor column once:

```rust
impl CursorKey for user::Model {
    type Column = user::Column;
    type Value = i32;
    const COLUMN: user::Column = user::Column::Id;
    fn cursor_value(&self) -> i32 { self.id }
}

#[get(\"/users\")]
async fn list_users(db: Db, cursor: CursorPaginate<i32>) -> Result<CursorPaginated<user::Model>> {
    cursor.exec(User::find(), db.conn()).await
}
```

`exec_by(select, column, conn, |m| ...)` is available for one-off cursors on a non-default column.

Single-column cursors only in this PR. Composite cursors are a non-breaking follow-up, the wire format is opaque so it can carry tuples later without breaking existing clients.

### DIFF
--- a/docs/content/docs/core-concepts/pagination.md
+++ b/docs/content/docs/core-concepts/pagination.md
@@ -237,3 +237,111 @@ Any request with `?per_page=51` now returns a 422.
 `exec` runs the data fetch and count queries **concurrently** using `tokio::join!`, not sequentially. Two queries hit the database in parallel, so latency is the cost of whichever query is slower, not both combined.
 
 The `Select<E>` is cloned before splitting into fetch and count paginators. SeaORM query builders are cheap to clone (they're just AST nodes, not connections).
+
+## Cursor Pagination
+
+For high-traffic feeds, infinite-scroll lists, and any endpoint where rows can shift while a user is paging through them, cursor pagination is the better fit. Rapina ships `CursorPaginate<V>` and `CursorPaginated<T>` alongside the offset variant.
+
+### When to choose which
+
+| | Offset pagination | Cursor pagination |
+|---|---|---|
+| URL shape | `?page=2&per_page=20` | `?after=<token>&limit=20` |
+| Total count | Yes (`total`, `total_pages`) | No (intentional) |
+| Stable across writes | No, rows can shift between pages | Yes |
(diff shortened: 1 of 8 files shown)


docs/content/docs/core-concepts/pagination.md
juv: Maybe this line can be adjusted a bit to indicate that Rapina also comes with built-in cursor pagination. The reader will eventually end up in the `Cursor Pagination` section while scrolling through so I think it's a nit pick and not strictly necessary.

juv (dismissed):
LGTM

juv (approved):
LGTM

PR #542, opened 2026-05-05
feat: add Header<T> typed header extractor

### LINKED ISSUE #523: Typed header extractor with FastAPI-style ergonomics and OpenAPI flow-through
Rapina's `Headers` extractor (`rapina/src/extract/mod.rs:152`) returns a raw `http::HeaderMap`. Every handler that wants a specific header rewrites the same pattern: `.get("authorization").ok_or_else(...)?.to_str()?...`. JWT auth has `JsonWebToken<T>` (`rapina/src/jwt/extract.rs:120`) so this isn't about auth. It's about everything else: content negotiation, conditional requests, custom protocol headers, rate-limit keys, request IDs.

**What peers ship, ranked by real DX**

1. **FastAPI.** Headers declared as function parameters with defaults and automatic snake-to-kebab conversion. `x_request_id: str = Header()` reads `X-Request-Id`. Optional via `| None = Header(default=None)`. Flows into OpenAPI without extra wiring. Validation via Pydantic.
2. **Nest.js.** Decorator per header, clean per line, no OpenAPI unless you stack another decorator.
3. **Spring.** `@RequestHeader("X-Request-Id") String requestId`. Clean.
4. **axum** (`axum-extra::TypedHeader<T>`). Full type coverage via the `headers` crate, but verbose: `TypedHeader(Authorization(bearer)): TypedHeader<Authorization<Bearer>>`. No OpenAPI integration. Generic 400 on failure.
5. Everyone else (Express, Hono, Gin, Rocket). String access plus manual parsing.

FastAPI wins per-case ergonomics. axum wins type coverage. Nobody combines them. Nobody has stable machine-parseable error codes for header failures. Nobody does OpenAPI flow-through in Rust.

**What Rapina ships**

A typed header extractor that combines FastAPI-style per-param ergonomics with Rust-native type coverage and OpenAPI flow-through. Three pieces:

1. **`Header<T>` extractor.** Declared as a handler parameter. The identifier becomes the header name via snake-to-kebab conversion. Override with `#[header("exact-name")]`.

   ```rust
   use rapina::prelude::*;
   use rapina::extract::headers::{Authorization, Bearer, UserAgent};

   async fn handler(
       x_request_id: Header<Uuid>,
       user_agent: Header<UserAgent>,
       if_none_match: Option<Header<String>>,
       #[header("x-tenant")] tenant: Header<TenantId>,
   ) -> Result<()> { ... }
   ```

   `x_request_id` reads `X-Request-Id`. `user_agent` reads `User-Agent`. Optional via `Option<Header<T>>`, which returns `None` when the header is missing and still 400s when it's malformed.

2. **Automatic OpenAPI parameter generation.** Every `Header<T>` in a handler signature becomes a documented parameter in the OpenAPI spec. Type inferred from `T`, `required: false` only when the param is `Option<Header<T>>`. No second declaration. FastAPI does this in Python; nobody does it in Rust yet.

3. **Stable error codes with field-level details.** `MISSING_HEADER` (400) when required and absent. `INVALID_HEADER` (400) when present but unparseable. Both include `details: { header: "x-request-id", reason: "invalid uuid" }`. Parseable by an agent, useful now and tighter once the error catalog work lands.

Parsing reuses the `headers` crate under the hood so we don't reinvent `Authorization<Bearer>`, `IfMatch`, `Range`, and friends. Common types re-exported under `rapina::extract::headers::*` so users don't add another `Cargo.toml` line.

**Design decisions**

1. **Name: `Header<T>`, not `TypedHeader<T>`.** Shorter, closer to FastAPI's mental model. The `headers` crate has a trait named `Header`, but users only touch that trait when implementing custom headers, and our wrapper is a struct. The collision is visible only in custom-header territory, which is rare.

2. **Identifier to header name.** Snake-to-kebab in the handler macro (`#[get]/#[post]/...`). `x_request_id` maps to `X-Request-Id` unambiguously. The explicit `#[header("exact-name")]` attribute handles non-standard casing and prefixed IDs.

3. **Inner type trait bound.** A small `HeaderValue` trait with `from_str` and `to_str` methods. Blanket impl over `T: headers::Header`, plus convenience impls for `String`, `Uuid`, and integer types. `Header<Uuid>` works without touching the `headers` trait directly.

4. **OpenAPI integration.** The handler macro already emits route metadata into `RouteInfo`. One new field: `header_parameters: Vec<HeaderParam>`. `build_openapi_spec` (`rapina/src/openapi/spec.rs:222`) picks it up and emits each as a `Parameter { location: Header, ... }`.

5. **Error codes.** `MISSING_HEADER` and `INVALID_HEADER`, both 400, with `details: { header, reason }`. Stable strings for now; move to the typed catalog when that lands.

**Out of scope for v1**

Each of these is a straight-line follow-up once v1 ships, better as its own issue:

- `Accept<T>` content negotiation (pick the best supported match)
- Response-side `Header<T>` for setting typed headers on outgoing responses
- `#[derive(FromHeaders)]` for struct-based grouping of many headers
- `Validated<Header<T>>` for per-header validation beyond what `headers` already does

**v1**

- `Header<T>` extractor in `rapina/src/extract/header.rs`
- Snake-to-kebab identifier conversion in the route handler macro
- `#[header("explicit-name")]` attribute for manual override
- `HeaderValue` trait with blanket impl over `headers::Header` plus `String`, `Uuid`, integer convenience impls
- Re-exports of common header types under `rapina::extract::headers::*`
- OpenAPI parameter emission per declared header
- `MISSING_HEADER` and `INVALID_HEADER` error codes
- Docs page
- Tests: required present, required missing → 400, malformed → 400, optional missing → None, explicit name override, OpenAPI output shape

**Why this lands as "best typed-header DX in any framework"**

Versus axum: one wrapper instead of two, identifiers read as header names, OpenAPI integration built in, stable error codes.

Versus FastAPI: compile-time type checking instead of runtime coercion, Rust-ecosystem header types (`Authorization<Bearer>`, `IfMatch`, `Range`) work without writing validators, zero runtime cost for type conversion.

**One piece worth validating before merge**

The macro-level identifier-to-header conversion is the load-bearing piece. A quick spike on the existing handler macro (`rapina-macros`) should confirm the param identifier is reachable at macro-expansion time. If not, fall back to explicit `#[header("...")]` on every param, which is still cleaner than axum but loses the FastAPI-style magic.

richasdev:
## Summary

Adds `Header<T>` extractor for type-safe HTTP header parsing with FastAPI-style ergonomics.
                                                                                                                                        
  - `Header<T>` struct with `into_inner()`, `deref`, and `header_name()`                                                                
  - `HeaderValue` trait with impls for String, Uuid, integers, floats, bool
  - Snake-case param name → kebab-case header name auto-conversion via heck                                                             
  - `#[header("explicit-name")]` attribute for manual override                                                                          
  - `Option<Header<T>>` support: absent=None, malformed=Err                                                                             
  - MISSING_HEADER (400) and INVALID_HEADER (400) with `details: {header, reason}`                                                      
  - Proc-macro codegen: detects Header<T> params, emits inline extraction, strips #[header] attrs                                       
  - OpenAPI flow-through: header_parameters() → RouteDescriptor → RouteInfo → spec Parameter{location: Header}                          
  - Re-exported under rapina::extract::* and prelude

## Related Issues

Closes #523 

## Checklist

- [x] `cargo fmt` passes
- [x] Tests pass


rapina/src/extract/header.rs
```diff
@@ -0,0 +1,352 @@
...
+    }
+}
+
+// ── HeaderValue trait ────────────────────────────────────────────────────────
+
+/// Trait for types that can be parsed from an HTTP header value string.
+///
+/// Implement this for custom types.  Blanket impls are provided for `String`,
+/// `uuid::Uuid`, and all primitive integer types.
+pub trait HeaderValue: Sized {
```
arferreira: This trait name collides with `http::HeaderValue`, which is the fundamental HTTP header value type from the `http` crate. Anyone writing `use rapina::prelude::*;` `use http::HeaderValue;` (or just use `http::*`) gets ambiguous name resolution.

rapina/src/lib.rs
```diff
@@ -141,7 +141,9 @@ pub mod prelude {
     };
     pub use crate::context::RequestContext;
     pub use crate::error::{DocumentedError, Error, ErrorVariant, IntoApiError, Result};
-    pub use crate::extract::{Context, Cookie, Form, Headers, Json, Path, Query, State, Validated};
+    pub use crate::extract::{
+        Context, Cookie, Form, Header, Headers, Json, Path, Query, State, Validated,
```
arferreira, line 145: Two issues stack here. First, the existing `Headers` extractor is also re-exported, so we now have both `Header` and `Headers` (raw `HeaderMap`). The naming pair is fine in isolation, but in any IDE autocomplete and stack traces they're easy to confuse. Worth a docstring on `Headers` calling out the distinction.

rapina-macros/src/lib.rs
```diff
@@ -302,8 +350,36 @@ fn route_macro_core(
...
+                    header_extractions.push(quote! { let #pat = #tmp; });
+                }
+            }
+            quote! {
+                let (__rapina_parts, __rapina_body) = __rapina_req.into_parts();
+                #(#header_extractions)*
+                let __rapina_result #return_type_annotation = (async #inner_block).await;
+                let __rapina_response = rapina::response::IntoResponse::into_response(__rapina_result);
+                #cache_header_injection
+                __rapina_response
```
arferreira, line 382: `__rapina_body` is bound and never used in this, the body is split off only because the surrounding pattern destructures into both names.

rapina-macros/src/lib.rs
```diff
@@ -550,6 +654,133 @@ fn extract_public_attr(attrs: &mut Vec<syn::Attribute>) -> bool {
...
+        let name = if let Some(n) = explicit_name {
+            n
+        } else if let Pat::Ident(pat_ident) = &*pat_type.pat {
+            pat_ident.ident.to_string().to_kebab_case()
+        } else {
+            // Destructure pattern — can't infer name, user must use #[header("name")]
+            panic!(
+                "Header<T> parameter with a destructure pattern must have a #[header(\"name\")] attribute"
+            );
+        };
```
arferreira: This produces a generic "panicked at" message in the user's compile output, with the span pointing at the macro call site rather than the offending parameter. 

The shape we want is:

```rust
return Err(syn::Error::new(
       pat_type.pat.span(),
       "Header<T> parameter with a destructure pattern must have a #[header(\"name\")] attribute",
   ).to_compile_error().into());
```

rapina/src/extract/header.rs
```diff
@@ -0,0 +1,352 @@
...
+}
+
+// ── Macro-generated FromRequestParts helper ──────────────────────────────────
+//
+// The rapina-macros crate calls this when it emits the blanket impl for a
+// handler parameter `foo: Header<T>`.  We expose the two helpers as public API
+// so the generated code compiles without needing to reach into private modules.
+
+/// Convenience re-export used by macro-generated code.
+pub use extract_header as __extract_header;
```
arferreira: These are explicitly described as "convenience re-export used by macro-generated code" but ship as ordinary `pub` items, which means they show up in published rustdoc with their `__`-prefixed names and that confusing description. Worth adding `#[doc(hidden)]`.

rapina/src/extract/header.rs
```diff
@@ -0,0 +1,352 @@
...
+// ── Placeholder FromRequestParts impl (test / manual use) ────────────────────
+//
+// A blanket impl is NOT possible here because `Header<T>` wraps a name that
+// must be supplied by the macro.  We provide a dummy impl gated behind a
+// test-only feature flag so unit tests can call `from_request_parts` directly
+// with a hard-coded name.  Production handler wiring goes through the macro.
+
+#[cfg(test)]
+pub struct NamedHeader<T, const N: usize>(pub Header<T>, std::marker::PhantomData<[(); N]>);
+
```
arferreira: This was never referenced in any test in this file. Dead code shipped behind a cfg flag.

rapina/src/extract/header.rs
```diff
@@ -0,0 +1,352 @@
...
+/// extraction.  The name is set by the proc-macro at compile time.
+#[derive(Debug, Clone)]
+pub struct Header<T> {
+    value: T,
+    name: &'static str,
+}
+
+impl<T> Header<T> {
+    /// Creates a new `Header` wrapper (used internally by the extractor impl).
+    pub fn new(name: &'static str, value: T) -> Self {
```
arferreira, line 52: The `&'static str` constraint on name works for the macro path because the codegen emits string literals, but it makes manual construction from runtime data (testing, custom extractors that wrap this one, dynamic header names from config) impossible without `Box::leak`. Suggest an additional constructor that takes `String` (or owning the name in the struct itself), or document explicitly that runtime construction is not supported.

rapina-macros/src/lib.rs
```diff
@@ -550,6 +654,133 @@ fn extract_public_attr(attrs: &mut Vec<syn::Attribute>) -> bool {
...
+    let last = type_path.path.segments.last()?;
+
+    // Direct Header<T>
+    if last.ident == "Header" {
+        if let syn::PathArguments::AngleBracketed(args) = &last.arguments {
+            if let Some(syn::GenericArgument::Type(inner)) = args.args.first() {
+                return Some((inner.clone(), true));
+            }
+        }
+    }
```
arferreira, line 747: `detect_header_type` matches by identifier only, so any user type named `Header<T>` (e.g. someone's own struct in a different module) will be picked up as Rapina's `Header<T>` and the macro will emit `rapina::extract::extract_header::<...>` against it. Probably surfaces as a compile error today (their type doesn't impl `HeaderValue`), but the diagnostic will point at macro generated code rather than the user's mistake. Worth at minimum a doc comment flagging the limitation, better is to also check the leading path segments when present.

rapina/src/extract/header.rs
```diff
@@ -0,0 +1,352 @@
...
+// `Header<T>` without a name because there is no way to infer the name from
+// `T` alone.  The macro supplies `name` through code generation.
+
+// ── Optional header support ──────────────────────────────────────────────────
+
+/// Extract an optional typed header.
+///
+/// Returns `None` if the header is absent, `Err` only if the header is present
+/// but cannot be parsed.
+pub fn extract_optional_header<T: HeaderValue>(
```
arferreira: `extract_header<T>` returns `Result<T, Error>` (raw inner type) but this returns `Result<Option<Header<T>>, Error>` (already wrapped). The macro then has to wrap the required case via `Header::new` because of this asymmetry. Either both should return `Header<T>` or both should return the raw `T` and let the macro be the only thing that constructs the wrapper. Picking one direction will simplify the codegen path.

arferreira (changes requested):
Hey richas good work! I left some comments worth to be addressed before merge.

arferreira (changes requested):
Looks good apart from two nits: `detect_header_type` doc just describes the bad failure mode. If I have my own `Header<T>` in another module gets a compile error pointing at macro generated code, which is the confusing case. Worth resolving the full path inside `detect_header_type` so we only match `rapina::extract::Header`. Then it's "no match, fall through to normal handling" instead of "compile error from somewhere weird."

The `&'static str` design rationale is reasonable but rules out unit testing extraction logic directly and writing wrapper extractors that delegate to `Header<T>`. A `#[doc(hidden)]` constructor that takes `String` for those cases would keep the macro path unchanged but unblock testing.

arferreira (approved):
LGTM

PR #546, opened 2026-05-06
fix: consolidate PK source of truth in codegen

### LINKED ISSUE #543: [Follow-up] Consolidate Primary-Key Source of Truth in Codegen
### Context
We currently have **two sources of truth** for primary-key membership:

1. `generate_schema_block` receives a separate `pk: Vec<String>` (used for composite keys).
2. Migration generation reads `field.is_primary_key`.

This can lead to divergence when both representations disagree.  
While this was not introduced by the current PR, the PR reinforces the current shape and makes this debt more explicit.

### Problem
Primary-key membership is modeled in two places with different data flows, which increases the risk of:
- schema and migration mismatch
- incorrect PK flags in generated output
- fragile future refactors around composite keys

### Notes
PR reference:  #519

richasdev:
## Summary

 - generate_migration/create_migration_file accept primary_key: Option<&[String]>                                                      
  - import command now forwards introspected PK list to migration codegen                                                               
  - composite PK columns no longer emit .auto_increment() (invalid SQL)
  - add is_composite_pk_member to FieldInfo to suppress auto_increment                                                                  
  - 3 new tests: composite PK, custom-named PK, default id regression

## Related Issues

Closes #543 

## Checklist

- [x] `cargo fmt` passes
- [x] Tests pass


arferreira (changes requested):
Hey nice work! Just one small test gap: `test_generate_migration_custom_named_pk` uses a `uuid` PK which never emits `.auto_increment()` regardless of state, so the new composite vs single branch in `generate_column` isn't actually exercised. Adding a case with `Some(pk)` and a single `i32` PK named other than id for example `user_id` would cover it. 

Also worth noting that the new `primary_key: Option<&[String]>` path through `generate_migration` is currently unreachable from `import` because `filter_and_validate_tables` rejects composite PK and non id named PK tables upstream. That's fine for this PR, but worth a follow-up issue to either unblock the filter when composite PKs are properly supported.

arferreira (approved):
LGTM

PR #547, opened 2026-05-06
[Feature]: Opt-in auto-migrations at startup + Django-style pending warning when disabled

### LINKED ISSUE #539: [Feature]: Opt-in auto-migrations at startup + Django-style pending warning when disabled
### Problem Statement

Rapina apps typically apply schema changes by calling run_migrations::<migrations::Migrator>() at startup. That is easy for local development and single-node setups, but it is a poor default for real deployments: every replica runs migrations on boot (racing and duplicating work), there is no first-class status/down at the process boundary, and “migrate inside the web process” is a pattern that works until it does not.

A concrete footgun is environment mismatch: you point DATABASE_URL at staging or production to debug or patch something, but your local checkout still has migration files you have not merged or released. On startup, auto-run applies whatever is in your tree to their database, so you can unintentionally migrate prod/staging from work-in-progress schema changes. That is hard to notice until something breaks or until you realize the remote DB is ahead of what the rest of the fleet expects.

Meanwhile, scaffolding-only commands (e.g. rapina migrate new) do not apply migrations by themselves; the maintainable fix is a dedicated CLI path (rapina migrate init, rapina migrate up, etc.) so applying schema is an explicit step. Even with that in place, the framework should not treat “always migrate from the HTTP process” as the normal production default.

There is still no configuration-level way to say: “do not auto-apply on boot; instead, tell me what is pending and point me to the supported CLI,” analogous to Django’s dev-server message about unapplied migrations. Operators need a single, high-signal log line (count, names, risk note, explicit rapina migrate up / init guidance) without blocking server startup for the default case.

### Proposed Solution

Config: Add DatabaseConfig::auto_migrate (and env DATABASE_AUTO_MIGRATE) consistent with other DATABASE_* variables. Keep threading a minimal migration policy into state from with_database (not the full URL).

Behavior in run_migrations:

auto_migrate == true: Current behavior — run all pending migrations via existing run_pending; fail if migration fails.
auto_migrate == false: Do not apply (new default). If MigratorTrait::get_pending_migrations (or equivalent) is non-empty, emit one tracing::warn! in Django’s shape: number of pending migrations, their names, warning that behavior may be wrong until applied, and Run rapina migrate up to apply them (docs mention rapina migrate init for projects missing the bin).
Compatibility: If policy is absent (e.g. tests that only register a connection), treat as auto-apply or document a single clear rule to avoid breaking existing tests.

Docs: Update migrations + database/configuration docs: when startup auto-migrate is reasonable (single-node/dev), when to use rapina migrate up, and the trust/deploy framing.

Default: Align with project semver policy — either default false once CLI is stable (explicit opt-in to boot-time migrate) or default true for one release with a deprecation path before flipping.

### Alternatives Considered

Startup-only migrations, no flag: Simplest, but leaves multi-replica and ops best practices unaddressed and gives no Django-style visibility when you intentionally skip boot-time applies.

Remove run_migrations from the builder: Forces everyone through the CLI, but breaks ergonomics for small apps and samples; a toggle preserves the API while making the safe default documented.

Fail hard if pending when auto_migrate is false: Safer for production, but diverges from Django’s warn-and-continue dev UX; can be a follow-up behind a separate strict env flag.

Put the flag only on run_migrations(...) instead of DatabaseConfig: Avoids new env vars but splits configuration across call sites and is harder to discover than DATABASE_*.

Expose full DatabaseConfig in AppState for this: Avoided — increases risk of accidentally extracting secrets in handlers; a tiny MigrationPolicy is enough.

### Contribution

chiucchi:
## Summary

Adds DATABASE_AUTO_MIGRATE and DatabaseConfig::auto_migrate (default false). When run_migrations runs with auto-migrate off, pending migrations are not applied at startup; Rapina logs a Django-style warning (count, migration names, and pointers to rapina migrate up / rapina migrate init). Opt in to the old behavior with true or .auto_migrate(true).

Wires this through with_database by registering a small DatabaseMigrationPolicy in app state (no URL in state). If only a raw DatabaseConnection is registered, policy is treated as off.

Introduces migration::run_startup_migrations and format_pending_migrations_warning. Updates docs, CHANGELOG (breaking default), CRUD template, and todo example.

## Related Issues

Closes #539 

## Checklist

- [x] `cargo fmt` passes
- [x] Tests pass


rapina/src/database.rs
```diff
@@ -182,6 +206,16 @@ impl DatabaseConfig {
...
+fn parse_env_bool(var: &str) -> Option<bool> {
+    std::env::var(var)
+        .ok()
+        .and_then(|v| match v.to_lowercase().as_str() {
+            "1" | "true" | "yes" => Some(true),
+            "0" | "false" | "no" => Some(false),
+            _ => v.parse().ok(),
+        })
+}
+
```
ShiraiEd, line 223: If `DATABASE_AUTO_MIGRATE` is set to an unrecognized value (e.g. "enabled", "on", or a typo), the function returns `None` and `unwrap_or(false)` silently opts
  the user out of auto-migrate with no indication anything went wrong. An operator who sets `DATABASE_AUTO_MIGRATE=enabled` expecting migrations to run will
  be surprised.
```rust
  Suggest adding a tracing::warn! in the _ arm before returning None:
  _ => {
      tracing::warn!(
          "Unrecognized value for {var}: {v:?}. Expected true/false/1/0/yes/no. Defaulting to false."
      );
      None
  }
```

ShiraiEd (commented):
Hi, great work

I left a comment in database.rs `parse_env_bool` function

Also I think `DATABASE_AUTO_MIGRATE` should be added to the `.env.example` in `todo-app`, 
The url-shortner also need an update to address this feat, the same way crud and todo-app were fixed

rapina/src/database.rs
```diff
@@ -182,6 +206,16 @@ impl DatabaseConfig {
     }
 }
 
+fn parse_env_bool(var: &str) -> Option<bool> {
+    std::env::var(var)
+        .ok()
+        .and_then(|v| match v.to_lowercase().as_str() {
+            "1" | "true" | "yes" => Some(true),
+            "0" | "false" | "no" => Some(false),
+            _ => v.parse().ok(),
```
arferreira: The only outcome of this arm is `None`, I mean this fallback can never produce `Some`.

arferreira (changes requested):
Looks great, thanks! Just few nits to address before merge.

arferreira (approved):
LGTM, I see two follow-up issues, when I was looking the examples that `.auto_migrate(true)` makes me think that default is probably wrong, also the `run_migrations()` is misleading when auto-apply is false, it doesn't run migrations, just warns,  we need to see these in the real world.

PR #548, opened 2026-05-07
Generate relationship fields in schema! block from foreign keys

### LINKED ISSUE #244: Generate relationship fields in schema! block from foreign keys

ShiraiEd:
## Summary

- `rapina import database` now translates FK columns into `BelongsTo`/`HasMany` relationship fields in the generated `schema!` block instead of emitting them as plain integer columns
- Raw FK columns (e.g. `user_id`) are excluded from the scalar fields list; the owning side gets a `belongs_to: RelatedModel` (or `Option<RelatedModel>` if nullable) field instead
- The referenced side gets a `items: Vec<OwningModel>` has-many field; duplicate has-many entries from multiple FKs to the same target are deduplicated
- Introduces `RelationshipSpec` / `RelationshipKind` in `codegen.rs` and threads them through `generate_schema_block` and `update_entity_file`

## Related Issues

Closes #244

## Checklist

- [x] `cargo fmt` passes
- [x] Tests pass

### DIFF
--- a/rapina-cli/src/commands/codegen.rs
+++ b/rapina-cli/src/commands/codegen.rs
@@ -4,6 +4,8 @@ use super::{Colorize, FieldInfo, NormalizedType};
 use std::fs;
 use std::path::Path;
 
+pub(crate) use super::relationships::{RelationshipKind, RelationshipSpec};
+
 pub(crate) fn to_pascal_case(s: &str) -> String {
     s.split('_')
         .map(|part| {
@@ -450,8 +452,9 @@ pub(crate) fn generate_schema_block(
     fields: &[FieldInfo],
     timestamps: Option<&str>,
     primary_key: Option<&[String]>,
+    relationships: &[RelationshipSpec],
 ) -> String {
-    let schema_fields: Vec<String> = fields
+    let mut all_field_lines: Vec<String> = fields
         .iter()
         .map(|f| {
             format!(
@@ -461,6 +464,21 @@ pub(crate) fn generate_schema_block(
             )
         })
         .collect();
+
+    for r in relationships {
+        let line = match r.kind {
+            RelationshipKind::BelongsTo { nullable: true } => {
+                format!("        {}: Option<{}>,", r.field_name, r.related_pascal)
+            }
+            RelationshipKind::BelongsTo { nullable: false } => {
+                format!("        {}: {},", r.field_name, r.related_pascal)
+            }
+            RelationshipKind::HasMany => {
+                format!("        {}: Vec<{}>,", r.field_name, r.related_pascal)
+            }
+        };
+        all_field_lines.push(line);
+    }
 
     let mut attrs = String::new();
 
@@ -483,7 +501,7 @@ schema! {{
 "#,
         pascal = pascal,
         attrs = attrs,
-        fields = schema_fields.join("\n"),
+        fields = all_field_lines.join("\n"),
     )
 }
 
@@ -615,6 +633,7 @@ pub(crate) fn update_entity_file(
     timestamps: Option<&str>,
     primary_key: Option<&[String]>,
     force: bool,
+    relationships: &[RelationshipSpec],
 ) -> Result<(), String> {
     update_entity_file_in(
         pascal,
@@ -622,6 +641,7 @@ pub(crate) fn update_entity_file(
         timestamps,
         primary_key,
         force,
+        relationships,
         Path::new("src/entity.rs"),
     )
 }
@@ -632,9 +652,11 @@ fn update_entity_file_in(
     timestamps: Option<&str>,
     primary_key: Option<&[String]>,
     force: bool,
+    relationships: &[RelationshipSpec],
     entity_path: &Path,
 ) -> Result<(), String> {
-    let schema_block = generate_schema_block(pascal, fields, timestamps, primary_key);
+    let schema_block =
+        generate_schema_block(pascal, fields, timestamps, primary_key, relationships);
 
     if entity_path.exists() {
         let mut content = fs::read_to_string(entity_path)
@@ -1262,7 +1284,7 @@ mod tests {
             "title:string".parse().unwrap(),
             "done:bool".parse().unwrap(),
         ];
-        let content = generate_schema_block("Todo", &fields, None, None);
+        let content = generate_schema_block("Todo", &fields, None, None, &[]);
 
         assert!(content.contains("title: String,"));
         assert!(content.contains("done: bool,"));
@@ -1365,16 +1387,16 @@ mod tests {
     fn test_generate_schema_block_with_timestamps() {
         let fields = vec!["title:string".parse().unwrap()];
 
-        let block = generate_schema_block("Post", &fields, None, None);
+        let block = generate_schema_block("Post", &fields, None, None, &[]);
         assert!(block.contains("schema! {"));
         assert!(block.contains("Post {"));
         assert!(block.contains("title: String,"));
         assert!(!block.contains("#[timestamps"));
 
-        let block = generate_schema_block("Post", &fields, Some("none"), None);
+        let block = generate_schema_block("Post", &fields, Some("none"), None, &[]);
         assert!(block.contains("#[timestamps(none)]"));
 
-        let block = generate_schema_block("Post", &fields, Some("created_at"), None);
+        let block = generate_schema_block("Post", &fields, Some("created_at"), None, &[]);
         assert!(block.contains("#[timestamps(created_at)]"));
     }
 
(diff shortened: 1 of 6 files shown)


rapina-cli/src/commands/import.rs
```diff
@@ -460,27 +455,42 @@ fn resolve_relationships(tables: &[IntrospectedTable]) -> HashMap<String, Vec<Re
             let ref_singular = codegen::singularize(&fk.referenced_table);
             let ref_pascal = codegen::to_pascal_case(&ref_singular);
 
+            let nullable = table
+                .columns
+                .iter()
+                .find(|c| &c.name == fk_column)
+                .map(|c| c.is_nullable)
+                .unwrap_or(false);
+
```
richasdev: **Blocker:** If `fk_column` is not found in `table.columns` (schema inconsistency, case-sensitivity mismatch, partial introspection), this silently defaults to `nullable = false` and emits `author: User` instead of `author: Option<User>`. No warning, no error.

Suggested fix: skip the FK with a warning instead of inventing a value:
```rust
let Some(col) = table.columns.iter().find(|c| &c.name == fk_column) else {
    eprintln!(
        "  {} table {:?}: FK column {:?} not found in column list — skipping relationship",
        "warn:".yellow(), table.name, fk_column
    );
    continue;
};
let nullable = col.is_nullable;
```

rapina-cli/src/commands/import.rs
```diff
@@ -579,7 +589,46 @@ fn generate_for_table(
...
+    codegen::update_entity_file(
+        &pascal,
+        &entity_fields,
+        timestamps,
+        primary_key.as_deref(),
+        force,
+        &rel_specs,
+    )?;
     codegen::create_migration_file(plural, &pascal_plural, &fields, false)?;
     codegen::create_feature_module(&singular, plural, &pascal, &fields, &pk_type, force)?;
```
richasdev, line 644: **Blocker:** `create_feature_module` receives the raw `&fields` list (including FK columns), while `update_entity_file` receives the filtered `&entity_fields`. This means `dto.rs` will contain `author_id: i32` while `entity.rs` has `author: User` — **the generated code will not compile** when a handler references `author_id` on the entity type.

The migration correctly uses raw `fields` (it mirrors the DB schema), but the feature module must mirror the entity shape.

Fix: pass `&entity_fields` instead:
```rust
codegen::create_feature_module(&singular, plural, &pascal, &entity_fields, &pk_type, force)?;
```

rapina-cli/src/commands/codegen.rs
```diff
@@ -4,6 +4,21 @@ use super::{Colorize, FieldInfo, NormalizedType};
...
+pub(crate) struct RelationshipSpec {
+    pub field_name: String,
+    pub related_pascal: String,
+    pub nullable: bool,
+    pub kind: RelationshipKind,
+}
+
+// Construction sites live in import.rs, which is #[cfg(feature = "import")].
+#[allow(dead_code)]
+#[derive(Clone, Copy, Debug)]
```
richasdev: The `#[allow(dead_code)]` is understandable (variants are constructed in feature-gated `import.rs`, which is `#[cfg(feature = "import")]`), but it is too broad — it suppresses warnings that would catch unhandled variants in `generate_schema_block`. Consider scoping the types under the same `#[cfg(feature = "import-*")]` guards as their construction sites.

uemuradevexe (changes requested):

rapina-cli/src/commands/codegen.rs
ShiraiEd: Gating both of them would break `rapina add resource` because add.rs is a ungated caller,
if you gate the parameter, the call site in add.rs also needs a #[cfg] block to conditionally pass or omit the
argument, so the cfg noise spreads there too.
I could either gate `relationships: &[RelationshipSpec]` parameter in `generate_schema_block` or make another function that does the same thing but without the `relationships: &[RelationshipSpec]`,
I could also replace the enum with a bool but that would lose context in the code,
I could also make a enum just for RelationshipSpec but that would be a duplicated code.

the simpler solution would be keeping the `allow`
or remove it and duplicate code

richasdev: Keep #[allow(dead_code)]. The enum is not semantically dead, it's conditionally constructed. The comment already explains why.
Duplication or a second function are worse tradeoffs. This is the right call.

ShiraiEd: ok, just gonna comment it so future contributors can understand why

rapina-cli/src/commands/codegen.rs
```diff
@@ -4,6 +4,24 @@ use super::{Colorize, FieldInfo, NormalizedType};
...
+}
+
+// Construction sites live in import.rs, which is #[cfg(feature = "import")].
+// Alternatives considered: gating the types under the same cfg (breaks ungated callers in add.rs),
+// gating the `relationships` parameter (spreads cfg noise to every call site),
+// a second function without relationships (duplicated code), or replacing the enum with bool (loses context).
+#[allow(dead_code)]
+#[derive(Clone, Copy, Debug)]
+pub(crate) enum RelationshipKind {
+    BelongsTo,
```
richasdev: Using `#[allow(dead_code)]` fixes the compilation warning when the `import` feature is disabled, but it is still a smell.

A cleaner alternative would be to move `RelationshipKind` and `RelationshipSpec` into their own module (for example, `commands/relationships.rs`) and gate only the parts that are actually import-specific behind `#[cfg(feature = "import")]`.

That would keep the types available to `add.rs` without needing to suppress warnings in `codegen.rs`.

rapina-cli/src/commands/codegen.rs
```diff
@@ -450,8 +468,9 @@ pub(crate) fn generate_schema_block(
     fields: &[FieldInfo],
     timestamps: Option<&str>,
     primary_key: Option<&[String]>,
+    relationships: &[RelationshipSpec],
 ) -> String {
-    let schema_fields: Vec<String> = fields
```
richasdev, line 454: The `nullable` field in `RelationshipSpec` only has an effect for `BelongsTo`. For `HasMany`, it is always `false`, but it is still exposed as if it were relevant. It would be more expressive to model it as an enum:

```rust
pub(crate) enum RelationshipKind {
    BelongsTo { nullable: bool },
    HasMany,
}
```

This way, `nullable` only exists where it actually makes sense, and the `match` already guarantees that.

rapina-cli/src/commands/import.rs
```diff
@@ -460,27 +455,46 @@ fn resolve_relationships(tables: &[IntrospectedTable]) -> HashMap<String, Vec<Re
...
             // BelongsTo on the FK side
             relationships
                 .entry(table.name.clone())
                 .or_default()
                 .push(RelationshipInfo {
                     field_name: field_name.to_string(),
                     related_pascal: ref_pascal.clone(),
-                    kind: RelationKind::BelongsTo,
+                    kind: codegen::RelationshipKind::BelongsTo,
+                    nullable,
```
richasdev: Here it converts `RelationshipInfo` → `codegen::RelationshipSpec` field by field. Since the types have the same structure after the refactor, it would be worth implementing `From<&RelationshipInfo> for codegen::RelationshipSpec` and using `.iter().map(Into::into).collect()`.

This removes the manual mapping and makes the call site trivial.

rapina-cli/src/commands/import.rs
```diff
@@ -425,17 +425,12 @@ fn filter_and_validate_tables(
...
 #[derive(Debug, Clone)]
-#[allow(dead_code)]
 struct RelationshipInfo {
     field_name: String,
     related_pascal: String,
-    kind: RelationKind,
-}
-
-#[derive(Debug, Clone)]
-enum RelationKind {
```
richasdev, line 436: The `HasMany` deduplication via `.iter().any(...)` works correctly, but it is `O(n²)` in the number of foreign keys per table. For schemas with many self-references (for example, hierarchical threads), this does not scale well.

A `HashSet` keyed by `field_name` before the loop would solve this, but given the context (an import CLI, not a hot path), the current approach is acceptable. It may still be worth adding a comment explaining that this tradeoff was intentional.

richasdev (commented):
Good PR  the feature is correct and well tested. Leaving a few notes below.

richasdev (changes requested):

ShiraiEd:
@richasdev

richasdev (approved):
nice!
LGTM!

ShiraiEd:
@arferreira just need your approval to make sure before merging this

arferreira (approved):
Lit! Thanks for the work.

PR #557, opened 2026-05-08
fix(macros): reject unsigned integer types in schema fields (closes #549)

### LINKED ISSUE #549: bug: unsigned integer types (u64, u32) silently map to i64 in schema codegen
## Description

In `rapina-macros/src/schema/types.rs`, the `ScalarType::from_ident` function maps both `u64` and `u32` to `ScalarType::I64`, which generates a Rust `i64` field with no warning or compile error.

## Affected Code

`rapina-macros/src/schema/types.rs` — `ScalarType::from_ident`

```rust
"i64" | "u64" | "u32" | "bigint" => Some(ScalarType::I64),
```

## Expected vs Actual Behavior

**Expected:** Writing `u64` or `u32` in a `schema!` block either (a) maps to the correct unsigned Rust type, or (b) produces a compile-time error explaining that unsigned types are not supported.

**Actual:** Both types silently generate an `i64` field. A schema field declared as `u64` accepts values up to `u64::MAX` (18446744073709551615), but the generated model field is `i64`, which overflows at `i64::MAX` (9223372036854775807). No warning is emitted.

## Reproduction

```rust
schema! {
    Entity {
        view_count: u64,  // silently becomes i64 in generated model
    }
}
```

Generated output:

```rust
pub struct Model {
    pub view_count: i64,  // expected u64
}
```

Any value above `i64::MAX` inserted into the database will overflow or be rejected at runtime with no compile-time indication of the mismatch.

## Context

Most SQL databases (including Postgres) do not have a native unsigned 64-bit integer type, so some mapping decision is unavoidable. The issue is the silent coercion: a developer writing `u64` reasonably expects either a `u64` field or a compile-time error, not a silent downcast to `i64`.

A compile-time diagnostic (e.g. `compile_error!` or at minimum a `#[deprecated]`-style warning) would surface this at the schema definition site rather than at runtime.

## Acceptance Criteria

- [ ] `u64` and `u32` in `schema!` blocks produce a compile-time error (or a clear `compile_error!` message) explaining the type is not supported
- [ ] Alternatively, if unsigned types are intentionally mapped to signed, the behavior is documented in the type mapping table and a warning is emitted via `proc_macro_warning` or similar
- [ ] Existing schemas using `i64`/`i32` are unaffected

## Environment

- Rapina: current main
- Affected: `rapina-macros/src/schema/types.rs` — `ScalarType::from_ident`

mvanhorn:
## Summary

Resolves #549. `parse_field_type` and `parse_inner_type` in `rapina-macros/src/schema/parse.rs` previously accepted `u8`, `u16`, `u32`, and `u64` and silently mapped them to `ScalarType::I32` / `I64`. Most SQL backends (including Postgres) have no native unsigned integer type, so a `u32` field with a value above `i32::MAX` would silently truncate when widened to a signed column.

## What changes

- `rapina-macros/src/schema/types.rs`: drop `u8/u16/u32/u64` from `ScalarType::from_ident`. Add `ScalarType::is_unsupported_unsigned` for parser-side detection. Document in the `from_ident` doc comment why unsigned types are intentionally rejected here.
- `rapina-macros/src/schema/parse.rs`: in both `parse_field_type` (top-level fields) and `parse_inner_type` (`Option<T>` / `Vec<T>` inner types), call `is_unsupported_unsigned` before scalar fallback and emit a focused `syn::Error` that names the offending type and suggests `i32`/`i64` with app-layer bounds, or `Decimal` for the full unsigned range.
- Tests in `parse.rs`:
  - `test_unsigned_integer_field_rejected` covers `u8/u16/u32/u64` at the top level.
  - `test_unsigned_integer_field_inside_option_rejected` pins the `Option<u64>` path.

## Why this matters

The old behavior was a quiet correctness bug: `u32` values above `i32::MAX` round-tripped through the schema as truncated signed integers without any compile-time or runtime warning. A compile-time rejection is a better contract because it forces callers to make the bound or representation choice explicitly.

Fixes #549


ShiraiEd:
Hello @mvanhorn, thanks for contributing!

The code change looks like it belongs here, but the PR description appears to be from a different project,
Could you update the summary for us?

mvanhorn:
@ShiraiEd good catch, sorry about that. Body was copy-pasted from a different PR by mistake. Updated to describe the actual change (rejecting `u8/u16/u32/u64` in schema fields with a focused compile-time error). The diff and tests were always correct.

ShiraiEd:
Thank you!

The review will be done soon

ShiraiEd (approved):
Solid work!

**LGTM**

richasdev (approved):
nice work @mvanhorn!

LGTM

mvanhorn:
Thanks @ShiraiEd, glad the unsigned-integer rejection landed. Let me know if anything looks off in the macro error messages, happy to iterate.

ShiraiEd:
Will do!
Thanks again for the contribution, @mvanhorn 
that was a great first PR.

mvanhorn:
Thanks @ShiraiEd, appreciate the warm welcome. Happy to circle back if you spot any other macro-side rough edges.

mvanhorn:
Thanks for the merge @arferreira. Rejecting the silent u32 -> i32 mapping at the macro layer is the right call; better to fail loud than corrupt data on the Postgres side.

PR #561, opened 2026-05-10
Resolve workspace-inherited rapina features in detect_flags

### LINKED ISSUE #545: [Good First Issue]: detect_flags silently ignores features declared under [workspace.dependencies] in monorepo setups
## This depends on PR #535.

detect_flags in `rapina-cli/src/commands/agents.rs` determines which optional fragments to include in AGENTS.md by reading `[dependencies].rapina.features` directly from the member's `Cargo.toml.`

In a Cargo workspace, a member can inherit the rapina dependency from the workspace root instead of declaring it locally:
```rust
  # workspace root Cargo.toml
  [workspace.dependencies]
  rapina = { version = "0.11", features = ["postgres"] }

  # member Cargo.toml
  [dependencies]
  rapina.workspace = true   # features live in the root, not here
```
In this case, detect_flags finds no features key in the member's `Cargo.toml` and silently returns `with_db: false` so migrations.md is never included in AGENTS.md, even though the project uses postgres.

https://github.com/rapina-rs/rapina/pull/535#discussion_r3157809012

### Why This Matters

- Monorepo users get an incomplete AGENTS.md with no error or warning because the bug is silent.
- The migrations.md fragment (and potentially websocket.md / jobs.md) are omitted, meaning AI agents get no guidance on database migrations for projects that need it.

### Relevant Files

- rapina-cli/src/commands/agents.rs
    - detect_flags — lines 28–48 (reads only member-level [dependencies])
    - fix_agents — line 328 (call site that passes the member's Cargo.toml)

### Acceptance Criteria

- [ ] Documentation is updated if needed
- [ ] detect_flags correctly reads features from [workspace.dependencies] when the member uses rapina.workspace = true
- [ ] Behavior for non-workspace projects is unchanged
- [ ] A test covers the workspace inheritance case (member with rapina.workspace = true, root with features)
- [ ] If features are declared in both root and member, they are merged and all are detected (e.g. root has ["postgres"], member adds ["websocket"] → with_db: true and with_websocket: true)
- [ ] If the same feature appears in both root and member, it is not double-counted or treated as an error

### Estimated Effort

L — half a day

### Hints & Resources

The fix requires two steps:

  1. After parsing the member's Cargo.toml, check if rapina is declared as { workspace = true }.
  2. If so, walk up the directory tree to find the workspace root (the Cargo.toml that contains a [workspace] table) and read [workspace.dependencies].rapina.features from there.

Member-level features (if any) should be merged with workspace-level features, Cargo treats them as additive.

Relevant Cargo docs: [Inheriting dependencies from a workspace](https://doc.rust-lang.org/cargo/reference/workspaces.html#inheriting-dependencies-from-the-workspace)

### Maintainer Checklist

- [ ] Issue is self-contained and scoped for a first-time contributor
- [ ] Relevant files and context are provided above
- [ ] Acceptance criteria are clear and verifiable

mvanhorn:
## Summary

Closes #545.

`detect_flags` in `rapina-cli/src/commands/agents.rs` reads
`[dependencies].rapina.features` from the member's `Cargo.toml`
directly. In a Cargo workspace where a member declares the rapina
dependency via `rapina.workspace = true` and the actual feature list
lives in the workspace root's `[workspace.dependencies]`, the previous
implementation found no features key on the member side and silently
produced an empty `AgentsFlags`. The generated `AGENTS.md` was missing
`migrations.md` / `websocket.md` / `jobs.md` even when the workspace
clearly used those features — and there was no warning to point a
monorepo user at the cause.

This PR adds a `detect_flags_with_workspace(member_cargo, member_dir)`
variant that resolves the workspace-inherited features and wires it
into the two call sites that have a base path available.

## Related Issues

Closes #545.

## Implementation

| File | Change |
| --- | --- |
| `rapina-cli/src/commands/agents.rs` | New `detect_flags_with_workspace` (and three private helpers); `fix_agents` and the drift detector switched to it; mapping logic factored into `flags_from_features` so the sqlite/postgres/mysql → `with_db` rule has a single source of truth. |

Resolution order in the new helper:

1. Read `dependencies.rapina.features` from the member's `Cargo.toml`
   (existing behaviour, identical bytes).
2. If the member declares `rapina.workspace = true`, walk up from
   `member_dir` looking for the first `Cargo.toml` whose top level
   contains a `[workspace]` table; read
   `workspace.dependencies.rapina.features` from it.
3. Union the two feature sets. A feature appearing in both layers is
   mapped to the same flag once — no double-counting, no error.

The walk-up uses `Path::parent()` and stops at the filesystem root.
A missing or unparseable workspace `Cargo.toml` falls back to the
member's own features rather than panicking — `rapina doctor`-style
commands shouldn't crash on a malformed parent.

## Acceptance criteria from #545

- [x] `detect_flags` correctly reads features from `[workspace.dependencies]` when the member uses `rapina.workspace = true` (test `test_detect_flags_with_workspace_inherits_features_from_root`)
- [x] Behavior for non-workspace projects is unchanged (test `test_detect_flags_with_workspace_ignores_root_when_member_declares_inline`)
- [x] A test covers the workspace inheritance case (member with `rapina.workspace = true`, root with features) — same first test
- [x] Features in both root and member are merged so all flags fire (test `test_detect_flags_with_workspace_merges_member_and_root_features`)
- [x] Same feature in both root and member is not double-counted or treated as an error (test `test_detect_flags_with_workspace_dedupes_overlapping_features`)
- [x] Tests added (5 new cases — see `cargo test detect_flags`)
- [x] `cargo fmt` clean
- [x] `cargo clippy --all-targets -- -D warnings` clean

I also added a regression-style test for the missing-workspace-root
fallback (`test_detect_flags_with_workspace_missing_root_falls_back_to_member`)
to nail down the contract for the panic-vs-degrade question, since the
issue's hint section describes the fix path but not the failure mode.

## Verification

```
cargo build                        # clean
cargo test -p rapina-cli detect_flags
# running 8 tests
# ........
# test result: ok. 8 passed; 0 failed
cargo test --workspace             # all green
cargo fmt --all -- --check         # clean
cargo clippy --quiet --all-targets -- -D warnings  # no output
```

The existing 3 detect_flags tests still pass unchanged — the public
`detect_flags(&toml::Value)` signature is preserved.

## Notes

- `detect_flags` (single-arg) stays public and behaves exactly as
  before for callers that don't have a path. Its docstring now points
  at `detect_flags_with_workspace` for the workspace-inheritance case.
- The drift detector reads `std::env::current_dir()` to feed
  `member_dir`, since `verify_rapina_project` works from the cwd. If
  that read fails (rare — daemonised process, deleted cwd), the call
  falls back to the path-less `detect_flags` so drift detection still
  produces a verdict.
- I did not touch the existing per-feature mapping (`postgres` etc.)
  or change the `AgentsFlags` struct shape — the issue is purely
  about reading the right place.

## Checklist

- [x] `cargo fmt` passes
- [x] Tests pass
- [ ] Documentation updated (if needed) — the docstrings on the new
      helper and the updated `detect_flags` cover the workspace
      behaviour. Happy to add user-facing docs to AGENTS.md or the
      monorepo section of the README if you'd prefer.


rapina-cli/src/commands/agents.rs
```diff
@@ -556,6 +671,189 @@ rapina = "0.11""#,
...
+        )
+        .unwrap();
+
+        let member_dir = workspace.path().join("api");
+        std::fs::create_dir_all(&member_dir).unwrap();
+        // Member adds `websocket` on top of the inherited `postgres`. Cargo
+        // doesn't actually allow this exact shape (you can't both set
+        // `workspace = true` and inline-add features), but the resolver here
+        // is intentionally permissive about the source layout — the only
+        // promise is "every feature mentioned anywhere makes it into flags".
```
ShiraiEd: This is incorrect. Cargo fully supports { workspace = true, features = ["extra"] }. From the Cargo reference:                                            
                                                                                                                                                           
  "The workspace member can also add additional features not listed in the workspace dependency."                                                        
                                                                                                                                                         
  The TOML shape being tested is not a quirky edge case the code is being lenient about — it's the primary real-world pattern for additive workspace       
  features (e.g., serde = { workspace = true, features = ["derive"] }). The comment misleads contributors into thinking this test exercises unusual
  permissiveness when it's actually the most important production case.                                                                                    
                                               
  Fix: Replace the comment with something like:                                                                                                          
```rust  
// Cargo allows a member to add features on top of the workspace-inherited
// set: `{ workspace = true, features = ["extra"] }`. The merge test      
// verifies that both layers of features contribute to the final flags. 
```

ShiraiEd (changes requested):
Nice work,

I left one comment I line 733

also:
`workspace_rapina_features` starts the walk from `member_dir` rather than `member_dir.parent(),` which means it correctly handles the case where the root 
package is also the workspace root (the first `Cargo.toml` it finds has `[workspace]` and is the right file to read from). 
That case isn't covered by a test though, if someone later changes the walk to start from `parent()`, it would silently break. Consider adding a test where `member_dir` is the 
workspace root itself.

mvanhorn:
Thanks for the review!

Addressed both points in 9136591:
- Replaced the misleading comment with your suggested wording — you're right that `{ workspace = true, features = ["extra"] }` is the primary real-world pattern (serde + derive being the canonical case), not a quirky permissive shape.
- Added `test_detect_flags_with_workspace_when_member_is_workspace_root` covering the root-equals-workspace case, so a future refactor of the walk start from `member_dir` to `member_dir.parent()` will fail loudly instead of silently breaking.

All 6 detect_flags tests pass locally.

ShiraiEd (approved):
LGTM

arferreira (approved):
Looks fine to me.

PR #562, opened 2026-05-11
Add Axum vs Rapina blog post

arferreira:
Just a blog post with my comparison between Axum and Rapina, opened for more contributions.


richasdev (approved):
LGTM. Balanced take - the "When Axum / When Rapina" sections are honest and the auth-secure-by-default argument is the strongest part.

richasdev (dismissed):

richasdev:
one minor nit: the middleware section compares a ready-made Tower layer (TraceLayer) with a custom Rapina middleware implementation from scratch  not quite apples-to-apples. Worth leveling that comparison in a follow-up, but not a blocker.

arferreira:
> one minor nit: the middleware section compares a ready-made Tower layer (TraceLayer) with a custom Rapina middleware implementation from scratch not quite apples-to-apples. Worth leveling that comparison in a follow-up, but not a blocker.

Def a good catch! Worth a follow-up rn.

juv:
I do agree that this is a well balanced take. Good job 👏 maybe we can iterate back on this topic - as I think it’s a good comparison that many devs are curious about - and update this blog for big features by mentioning new (smaller) blog posts on the top of this blog. Then it’s obvious that Rapina evolves quickly while we still care enough to give an updated take on every once in a while

juv:
Maybe we can also mention the (relatively new) jwks feature in the section about JWTs. This is something I looked for as an Axum middleware and didn’t find the same thing we added for Rapina: jwks setup as a first class citizen, easy to use based on an external idp, automated jwks refresh, with generic token claims that can easily be used in my custom logic. Axum has two, three community middleware crates but none of them had a lot of traction and maintenance the last time I looked (4-5 months ago). IIRC, They also focused primarily on redirecting the user to the login page upon missing/invalid token instead of having this more traditional backend 401 errors without redirects. Maybe I’m looking too much from an Enterprise perspective, where this kind of feature is basically a must have. You can ignore it if you think it’s getting too much into details.

ShiraiEd (commented):
Great writing

It would be worth dedicating a part to talk about the cli commands since Axum has none

Also, when talking about `rapina new`, a mention about the CLAUDE.md, AGENTS.md and .rapina-docs/ would be a nice nice addition

arferreira:
Okay, good catch guys, gonna add more sections to the post bringing these points.

juv (approved):
LGTM

PR #564, opened 2026-05-13
fix: skip write in generate_rapina_docs when content unchanged

### LINKED ISSUE #544: [Good First Issue]: generate_rapina_docs rewrites .rapina-docs/ on every call even when content is already up to date
## This depends on pr #535 
Every time rapina doctor --fix-agents runs, fix_agents calls generate_rapina_docs unconditionally at the end. Inside that function, every fragment file (core.md, extractors.md, errors.md, testing.md, and any conditional ones) is written with std::fs::write regardless of whether the file already exists with identical content.

Since fragment content is embedded via include_str! at compile time, it never changes within the same CLI version. This means repeated calls always do unnecessary disk writes even when the project is fully up to date.

https://github.com/rapina-rs/rapina/pull/535#discussion_r3157828670

### Why This Matters

- Unnecessary IO on every invocation, even on the UpToDate path.
- File modification timestamps change on each run, which can confuse tools that watch the filesystem (editors, inotify-based watchers, CI caches).
- It's misleading: the command appears to "do work" even when nothing changed.

### Relevant Files

- `rapina-cli/src/commands/agents.rs`
    - generate_rapina_docs — lines 104–140 (the write loop)
    - fix_agents — line 391 (the unconditional call site)

### Acceptance Criteria

- [ ] Documentation is updated if needed
- [ ] If all `.rapina-docs/` files already exist with content identical to what the CLI would write, no file is touched (mtimes must not change).
- [ ] A new test verifies that a second call to `generate_rapina_docs` with the same flags does not modify files (can assert via metadata mtime or by reading content before/after).
- [ ] All existing tests in agents.rs continue to pass.

### Estimated Effort

M — ~2 hours

### Hints & Resources

The fix is localized entirely to generate_rapina_docs. For each file to be written, read the existing content first and skip the write if it matches:
```rust
  let path = docs_path.join(name);
  let existing = std::fs::read(&path).unwrap_or_default();
  if existing != content.as_bytes() {
      std::fs::write(&path, content)
          .map_err(|e| format!("Failed to write .rapina-docs/{}: {}", name, e))?;
  }
```
  Apply the same pattern to the conditional fragments (migrations.md, websocket.md, jobs.md). No changes needed anywhere else.

### Maintainer Checklist

- [ ] Issue is self-contained and scoped for a first-time contributor
- [ ] Relevant files and context are provided above
- [ ] Acceptance criteria are clear and verifiable

richasdev:
## Summary

What does this PR do?

## fix: skip fragment writes when content is unchanged

Every call to `rapina doctor --fix-agents` was unconditionally rewriting all `.rapina-docs/` fragment files via `std::fs::write`, even when content was already up to date. Since fragments are embedded at compile time via `include_str!`, they never change within the same CLI version, making every write a no-op that still dirtied mtimes.

### Changes

`generate_rapina_docs` now reads existing file content before writing; write is skipped if `existing == content.as_bytes()`.

Applied to all 4 always-on fragments: `core.md`, `extractors.md`, `errors.md`, `testing.md`.

Applied to all 3 conditional fragments: `migrations.md`, `websocket.md`, `jobs.md`.

## Related Issues

Closes #544

## Checklist

- [x] `cargo fmt` passes
- [x] Tests pass
- [] Documentation updated (if needed)

### DIFF
--- a/rapina-cli/src/commands/agents.rs
+++ b/rapina-cli/src/commands/agents.rs
@@ -756,6 +756,10 @@ rapina = "0.11""#,
         let mtime_before = std::fs::metadata(&core_path).unwrap().modified().unwrap();
 
         // Small sleep so that a write would produce a different mtime.
+        // NOTE: assumes sub-second filesystem precision (APFS, ext4). On 1-second
+        // resolution filesystems (HFS+, FAT32) a write within the same second would
+        // be undetectable via mtime — content comparison cannot substitute here
+        // because written bytes are identical either way.
         std::thread::sleep(std::time::Duration::from_millis(10));
 
         // Second call: content identical → no write → mtime unchanged.
@@ -827,6 +831,10 @@ rapina = "0.11""#,
             })
             .collect();
 
+        // NOTE: assumes sub-second filesystem precision (APFS, ext4). On 1-second
+        // resolution filesystems (HFS+, FAT32) a write within the same second would
+        // be undetectable via mtime — content comparison cannot substitute here
+        // because written bytes are identical either way.
         std::thread::sleep(std::time::Duration::from_millis(10));
         generate_rapina_docs(dir.path(), &flags).unwrap();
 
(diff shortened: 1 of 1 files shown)


ShiraiEd (changes requested):
Nice work!                                  
            
  Nits:
- The skip tests rely on mtime comparison with a 10ms sleep. Worth adding a comment in the tests noting that this assumes sub-second filesystem
  precision (ext4, APFS), on 1-second resolution filesystems (HFS+, FAT32) a write within the same second would go undetected, and content comparison     
  can't replace this since the written bytes are identical either way.
                             
-   Also, there's no test with all three flags (with_db, with_websocket, with_jobs) set to true at the same time, that combination is never exercised.

ShiraiEd (approved):
LGTM

PR #572, opened 2026-05-15
Panic on duplicate operationId at spec build time

### LINKED ISSUE #565: [Feature]: Validate and deduplicate operationId in generated OpenAPI spec
### Problem Statement

`operationId` in the generated OpenAPI spec is derived directly from the handler function name with no uniqueness check. When two handlers share the same name (e.g. `get` or `list` defined in different modules), the resulting spec contains duplicate `operationId` values — which violates the OpenAPI 3.0 specification:

> Each operation ID MUST be unique among all operations described in the API.

This breaks client code generators (e.g. `openapi-generator`, `orval`) silently, producing either compilation errors or overwritten methods with no indication of what went wrong.

**Relevant code:** `rapina/src/openapi/spec.rs:289`

```rust
operation_id: Some(route.handler_name.clone()),
```

No deduplication or validation is performed before or after this assignment.

### Proposed Solution

At spec-build time, track all assigned `operationId` values and either:

1. Return an error if a duplicate is detected (fail-fast, explicit), or
2. Automatically suffix duplicates (e.g. `get_user` → `get_user_2`) with a warning

Option 1 is preferred — it surfaces the naming conflict at startup rather than silently producing an invalid spec.

### Alternatives Considered

_No response_

### Contribution

richasdev:
## Summary

Validates `operationId` uniqueness during OpenAPI spec construction. When two handlers share the same name, `build_openapi_spec` now panics at startup with a clear message identifying the duplicate — rather than silently emitting an invalid spec that breaks client code generators.

Internal routes (`/__rapina/*`) are excluded from the check, consistent with existing behavior.

## Related Issues

Closes #565

## Checklist

- [x] `cargo fmt` passes
- [x] Tests pass

### DIFF
--- a/rapina/src/openapi/spec.rs
+++ b/rapina/src/openapi/spec.rs
@@ -231,9 +231,20 @@ pub fn build_openapi_spec(
 
     spec.components = Some(Components { schemas });
 
+    let mut seen_operation_ids: std::collections::HashSet<String> =
+        std::collections::HashSet::new();
+
     for route in routes {
         if route.is_internal() {
             continue;
+        }
+
+        if !seen_operation_ids.insert(route.handler_name.clone()) {
+            panic!(
+                "Duplicate operationId '{}' in OpenAPI spec. Each handler must have a unique name. \
+                 Found duplicate on {} {}.",
+                route.handler_name, route.method, route.path
+            );
         }
         // Extract path parameters (e.g., :id -> id)
         let params: Vec<Parameter> = route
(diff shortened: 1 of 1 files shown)


ShiraiEd (commented):
Nice one, Richas!

Pr #542 was merged before this one, and in there was added one more param in the RouteInfo struct, so the CI checks are blocking because its missing a Vec::new() on the RouteInfo constructor params

**Non-blocking-comment**
Also, I pondered if the HashSet<String> approach would be the best option, since it allocates more memory to its creation, and clones every route name. In terms of time complexity it is very fast, like O(n) for the look ups, but in terms of memory I don't really know if its worth it, also there is the heap memory allocation of the <string>, so consider using &str instead, RouteInfo outlives seen_operations_ids so no lifetime problem.
Another path would be a double iteration over RouteInfo,an average web app maybe not even surpass 100 endpoints, so it wouldn't be that bad if it is O(n2) but in memory complexity is not that consuming because of the references.

ShiraiEd (changes requested):
PR #542 was merged before this one and added a new param to the `RouteInfo` struct so CI is blocking because the constructor call is missing a `Vec::new().`

  Non-blocking: consider replacing `HashSet<String>` with `HashSet<&str>.` You're cloning every route name into the set, but `RouteInfo` outlives
  `seen_operation_ids` so borrowing is fine, no lifetime issues. Saves the allocations and the clones.

richasdev:
Reopening from fork (richasdev) to enable CI on updated branch with merged main.

PR #573, opened 2026-05-15
fix(jwt): return error when validate_aud=true but set_audience() never called

### LINKED ISSUE #566: [Bug]: JWT aud misconfiguration silently bypasses audience validation
### Description

When `validate_aud = true` is set on the `Validation` config but `set_audience()` is never called, the framework logs a `debug!` message and silently continues — accepting tokens regardless of their `aud` claim.

This is a silent security misconfiguration: the developer believes audience validation is active, but tokens with any (or no) `aud` pass through.

**Relevant code:** `rapina/src/jwt/extract.rs:80-84`

```rust
let validation = if let Some(validation) = validation {
    if validation.validate_aud && validation.aud.is_none() {
        tracing::debug!(
            "aud claim validation is enabled but validation.set_audience was not called"
        );
    }
    // continues without returning an error
```

The `debug!` log is invisible in most production environments (default log level is `info` or higher), so this misconfiguration goes completely unnoticed.

### Steps to Reproduce

1. Configure JWT validation with `validate_aud = true`
2. Do **not** call `set_audience()`
3. Register the `Validation` in app state
4. Send a request with a JWT containing any `aud` value (or none)
5. Request is accepted — audience is not validated

### Expected Behavior

When `validate_aud = true` and `aud` is `None`, the framework should:
- Return a configuration error at startup or at first use, **not** silently continue
- At minimum, log at `warn!` or `error!` level so it surfaces in production

### Code Example

```rust
let mut validation = jwt::default_validation();
validation.validate_aud = true;
// forgot to call: validation.set_audience(&["https://my-api.com"]);

let app = Rapina::new()
    .state(jwks_client)
    .state(validation); // silently broken
```

### Environment

- Rapina version: latest
- Rust version: stable
- OS: any

### Error Output / Logs

No error is produced. Only a `DEBUG` log that is typically not visible in production.

richasdev:
## Summary

Addresses reviewer feedback on the JWT `aud` misconfiguration fix:

- **`required_spec_claims`**: Added `"aud"` to `default_validation()` so tokens without an `aud` claim are rejected by default, as recommended by the `jsonwebtoken` crate docs
- **Health check**: Added JWT misconfiguration check to `readiness_check` in `health/endpoint.rs` — if `validate_aud=true` and `set_audience()` was never called, the readiness endpoint returns 503 before real traffic hits, surfacing the misconfiguration via Kubernetes readiness probes
- **Panic on misconfiguration**: Replaced the silent 500 path with `tracing::error!` + `panic!` so the developer misconfiguration is caught at startup rather than serving broken responses indefinitely

## Related Issues

Closes #566

## Checklist

- [x] `cargo fmt` passes
- [x] Tests pass

### DIFF
--- a/rapina/src/health/endpoint.rs
+++ b/rapina/src/health/endpoint.rs
@@ -19,7 +19,6 @@ use crate::{
     response::{APPLICATION_JSON, BoxBody, full},
     state::AppState,
 };
-
 /// Handler for `GET /__rapina/health/live`.
 ///
 /// Always returns `200 OK` with `{"status": "ok"}` as long as the process is running.
@@ -73,6 +72,22 @@ pub async fn readiness_check(
                 if db_ok { "ok".into() } else { "error".into() },
             );
             if !db_ok {
+                all_ok = false;
+            }
+        }
+    }
+
+    // JWT configuration check — surfaces validate_aud misconfiguration before real traffic hits.
+    #[cfg(feature = "jwks")]
+    {
+        use jsonwebtoken::Validation;
+        if let Some(validation) = state.get::<Validation>() {
+            let jwt_ok = !(validation.validate_aud && validation.aud.is_none());
+            checks.insert(
+                "jwt".to_string(),
+                if jwt_ok { "ok".into() } else { "error".into() },
+            );
+            if !jwt_ok {
                 all_ok = false;
             }
         }
@@ -231,6 +246,42 @@ mod tests {
         assert_eq!(json["status"], "error");
         assert_eq!(json["checks"]["redis"], "ok");
         assert_eq!(json["checks"]["stripe"], "error");
+    }
+
+    #[cfg(feature = "jwks")]
+    #[tokio::test]
+    async fn test_readiness_check_jwt_misconfiguration_returns_503() {
+        use crate::jwt::default_validation;
+
+        let misconfigured = default_validation(); // validate_aud=true, aud=None
+        let app = Rapina::new().with_health_check(true).state(misconfigured);
+        let client = TestClient::new(app).await;
+        let response = client.get("/__rapina/health/ready").send().await;
+        let status = response.status();
+        let json = response.json::<Value>();
+
+        assert_eq!(status, StatusCode::SERVICE_UNAVAILABLE);
+        assert_eq!(json["status"], "error");
+        assert_eq!(json["checks"]["jwt"], "error");
+    }
+
+    #[cfg(feature = "jwks")]
+    #[tokio::test]
+    async fn test_readiness_check_jwt_correctly_configured_returns_ok() {
+        use crate::jwt::default_validation;
+
+        let mut validation = default_validation();
+        validation.set_audience(&["https://example.com/api"]);
+        let app = Rapina::new().with_health_check(true).state(validation);
+        let client = TestClient::new(app).await;
+        let json = client
+            .get("/__rapina/health/ready")
+            .send()
+            .await
+            .json::<Value>();
+
+        assert_eq!(json["status"], "ok");
+        assert_eq!(json["checks"]["jwt"], "ok");
     }
 
     #[cfg(feature = "sqlite")]
--- a/rapina/src/jwt/extract.rs
+++ b/rapina/src/jwt/extract.rs
@@ -78,9 +78,9 @@ where
 
         let validation = if let Some(validation) = validation {
             if validation.validate_aud && validation.aud.is_none() {
-                tracing::debug!(
-                    "aud claim validation is enabled but validation.set_audience was not called"
-                );
+                return Err(Error::internal(
+                    "JWT misconfiguration: validate_aud is enabled but set_audience() was never called",
+                ));
             }
 
             let mut v = validation.clone();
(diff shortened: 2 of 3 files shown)


juv:
If I’m not mistaking, I think the current implementation matches the one in jsonwebcrate: 

- token has `aud` claim, `set_audience` never called, `validate_aud = true` = error
- token has no `aud` claim, `set_audience` never called, `validate_aud = true` = passes unless `aud` is required with required_spec_claims

Making Rapinas default even more secure and less error-prone is great though. Current implementation will cause the webserver to throw 500 errors for all requests where a JWT is validated (which should be most endpoints typically) - is this expected? I’m not sure the a typical Rapina app will  recover from this. 

Besides of that, we might also need to check the presence of aud within `required_spec_claims` if we want to enforce it. And add it to the default validation of Rapina in the same way?

richasdev:
Yeah, that makes sense. Do you think the best approach would be to keep the stricter `aud` validation, but make sure those failures are mapped to a proper auth error, like `401 Unauthorized`, instead of becoming `500 Internal Server Error`?

Also, if we want to enforce `aud` by default, should we add it to `required_spec_claims` as part of Rapina default validation?

juv:
> Yeah, that makes sense. Do you think the best approach would be to keep the stricter `aud` validation, but make sure those failures are mapped to a proper auth error, like `401 Unauthorized`, instead of becoming `500 Internal Server Error`?

I’m thinking more one going into one of these directions:
1) we keep the 500 errors but add it to `rapina doctor` too, to show the misconfiguration as a finding there. However if I was a regular Rapina user, which we can assume is not 100% into all the details of Rapina, I can see how this implementation is a bit unexpected? It basically means that the code compiles but my server will return 500 responses forever. In infrastructure like Kubernetes, the liveness/readiness probe of a container are most often not using externally signed JWTs to access the health endpoints - and this means that detecting this misconfiguration will never trigger „early“ until real traffic hits the Webserver. 
2) this misconfiguration should panic, as the developer (realistically) will not implement a recovery code path for these errors. So the Webserver will never get out of these `500 internal server error`

what do you think?
 
> Also, if we want to enforce `aud` by default, should we add it to `required_spec_claims` as part of Rapina default validation?

👍 this is the recommended approach according to the jsonwebtoken crate docs

ShiraiEd (commented):
I think we should solve the root cause, in the current state validate_aud and the audiences can exist apart from each other, allowing the error to exist and compiling the code just fine to return surprise errors later

If the two were mixed together in a struct like
```rust
pub struct AudienceValidation {
       pub audiences: Vec<String>,
}
```
and passed as an `Option` in the validation config,
when `Some`, the validate_aud would be true and the audience would be included together, but when is `None`, it would be false with no audience, preventing the contradiction, a long term fix

The problems would be that current apis using jwt config would break, and the Validation struct itself doesnt belongs to Rapina, so the `Option<AudienceValidation>` wrapper would need to be a new Rapina-owned struct that gets converted into jsonwebtoken::Validation internally before use.

Better to leave this idea to another issue later.

But for now, I would go with the 500 approach since 401 would mean that the client's token is broken, which is not true

juv:
> I think we should solve the root cause, in the current state validate_aud and the audiences can exist apart from each other, allowing the error to exist and compiling the code just fine to return surprise errors later
> 
> If the two were mixed together in a struct like
> 
> ```rust
> pub struct AudienceValidation {
>        pub audiences: Vec<String>,
> }
> ```
> 
> and passed as an `Option` in the validation config, when `Some`, the validate_aud would be true and the audience would be included together, but when is `None`, it would be false with no audience, preventing the contradiction, a long term fix
> 
> The problems would be that current apis using jwt config would break, and the Validation struct itself doesnt belongs to Rapina, so the `Option<AudienceValidation>` wrapper would need to be a new Rapina-owned struct that gets converted into jsonwebtoken::Validation internally before use.
> 
> Better to leave this idea to another issue later.
> 
> But for now, I would go with the 500 approach since 401 would mean that the client's token is broken, which is not true

Honestly I think if we simply add the aud field to Rapinas default required_spec_claims, we got good coverage here and leave the rest up to the user (aligned with what the jsonwebtoken crate offers now and in the future)

ShiraiEd:
> Honestly I think if we simply add the aud field to Rapinas default required_spec_claims, we got good coverage here and leave the rest up to the user (aligned with what the jsonwebtoken crate offers now and in the future)

Thats a good idea, on top of that we could also add a JWT check to the `readiness_check in health/endpoint.rs` like
```rust
if let Some(validation) = state.get::<Validation>() {
      let jwt_ok = !(validation.validate_aud && validation.aud.is_none());
      checks.insert(
          "jwt".to_string(),
          if jwt_ok { "ok".into() } else { "error".into() },
      );
      if !jwt_ok {
          all_ok = false;
      }
  }
```
and as a last resort, the warn + panic

juv:
> > Honestly I think if we simply add the aud field to Rapinas default required_spec_claims, we got good coverage here and leave the rest up to the user (aligned with what the jsonwebtoken crate offers now and in the future)
> 
> Thats a good idea, on top of that we could also add a JWT check to the `readiness_check in health/endpoint.rs` like
> 
> ```rust
> if let Some(validation) = state.get::<Validation>() {
>       let jwt_ok = !(validation.validate_aud && validation.aud.is_none());
>       checks.insert(
>           "jwt".to_string(),
>           if jwt_ok { "ok".into() } else { "error".into() },
>       );
>       if !jwt_ok {
>           all_ok = false;
>       }
>   }
> ```
> 
> and as a last resort, the warn + panic

Sounds good to me 👍

ShiraiEd:
hey @richasdev 
did you understand the requests?

arferreira:
Hey @richasdev, will you finish this?

richasdev:
not yet, I’ll finish it

ShiraiEd (approved):
LGTM

PR #583, opened 2026-05-19
Fix: log validation error serialization failures instead of silently dropping them

### LINKED ISSUE #577: [Bug]: Validation error details silently dropped on serialization failure
## Description

In `rapina/src/extract/mod.rs:529`, validation error details are silently swallowed via `.unwrap_or_default()` when serialization of the error fails. The client receives an empty `details: {}` object with no indication of what went wrong.

## Steps to Reproduce

1. Implement a custom validator that returns non-serializable error data
2. Send a request that triggers validation failure
3. Observe the response — `details` field is empty `{}`

## Expected Behavior

When validation error serialization fails, Rapina should either:
- Log the serialization failure and include a fallback error message in the response, or
- Return a descriptive error indicating the validation details could not be serialized

Clients should never silently receive empty `details: {}` when a validation error occurred.

## Code Example

```rust
// extract/mod.rs:529
// Current behavior — serialization failure is silently swallowed:
.unwrap_or_default()

// Expected — log and surface the failure:
.unwrap_or_else(|e| {
    tracing::warn!("Failed to serialize validation error details: {e}");
    serde_json::Value::default()
})
```

## Environment

- Rapina version: 0.12.0
- Rust version: stable
- OS: any

richasdev:
## Summary

Replace silent `.unwrap_or_default()` with `.unwrap_or_else()` + `tracing::warn!` in both `Validated<Json<T>>` and `Validated<Form<T>>` extractors, so serialization failures are logged rather than silently swallowed. Also adds an `into_incoming_request()` helper to `TestRequest` and 8 new tests covering valid/invalid input and parse errors for both extractors.

## Related Issues

Closes #577

## Checklist

- [x] `cargo fmt` passes
- [x] Tests pass

### DIFF
--- a/rapina/src/extract/mod.rs
+++ b/rapina/src/extract/mod.rs
@@ -525,8 +525,12 @@ impl<T: DeserializeOwned + Validate + Send> FromRequest for Validated<Json<T>> {
     ) -> Result<Self, Error> {
         let json = Json::<T>::from_request(req, params, state).await?;
         json.0.validate().map_err(|e| {
-            Error::validation("validation failed")
-                .with_details(serde_json::to_value(e).unwrap_or_default())
+            Error::validation("validation failed").with_details(
+                serde_json::to_value(&e).unwrap_or_else(|err| {
+                    tracing::warn!("Failed to serialize validation error details: {err}");
+                    serde_json::Value::default()
+                }),
+            )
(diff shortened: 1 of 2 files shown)


ShiraiEd (approved):
LGTM

PR #586, opened 2026-05-20
refactor: replace curl subprocess with ureq HTTP client

### LINKED ISSUE #579: Replace curl shell-out with ureq in CLI fetchers
Mentioned in : https://github.com/rapina-rs/rapina/pull/528#pullrequestreview-4192964375

## Description

The commands:
- `rapina openapi export`
- `rapina llms export`
- `rapina routes`
- `rapina doctor`

All fetch data from the running `rapina dev` server via the `curl` command using the the subprocess in `std::process::Command` but that causes problems:
- `curl` must be already installed to work
- Loss of detail, no response body or status code, that is very bad especially on 5**/4** errors
- a wrong address or slow server hangs indefinitely until timeout fires

## Fix

Add `ureq` as a dependency and replace all four `Command::new("curl")` calls with proper HTTP calls

`ureq` over `reqwest`:

`reqwest` is used for async, and `tokio` is only used in optional features like `import` and `jobs`, it would add complexity with no benefits, `ureq` on the other hand, is synchronous and would fit the exisitng structure 

## Relevant Files

  - `rapina-cli/src/commands/openapi.rs:112`: fetch_openapi_spec
  - `rapina-cli/src/commands/llms.rs:27`: fetch_llms_txt
  - `rapina-cli/src/commands/routes.rs:70`: fetch_routes
  - `rapina-cli/src/commands/doctor.rs:360`: fetch_json
  - `rapina-cli/Cargo.toml`: add `ureq` here

## Acceptance Criteria

- [ ] Fix is implemented
- [ ] All four `Command::new("curl")` call sites are replaced with `ureq`
- [ ] HTTP status codes are surfaced in error messages (e.g. "server returned 404")
- [ ] A connect/read timeout is configured on the ureq client
- [ ] curl is no longer a runtime requirement for any CLI command
- [ ] Documentation updated if needed
- [ ] cargo fmt and cargo clippy pass

Ismaellima4:
## Summary

What does this PR do?

- Swap Command::new("curl") for ureq::get() in fetch functions
- Accept &ureq::Agent parameter for testability
- Add unit tests with local TcpListener for success/error cases

## Related Issues

Closes #579 

## Checklist

- [x] `cargo fmt` passes
- [x] Tests pass

### DIFF
--- a/rapina-cli/Cargo.toml
+++ b/rapina-cli/Cargo.toml
@@ -37,6 +37,7 @@ sqlx = { version = "0.8", optional = true }
 openapiv3 = { version = "2", optional = true }
 serde_yaml_ng = { version = "0.10", optional = true }
 sha2 = "0.11"
+ureq = "3.3.0"
 
 [features]
 default = []
--- a/rapina-cli/Cargo.toml
+++ b/rapina-cli/Cargo.toml
@@ -37,7 +37,7 @@ sqlx = { version = "0.8", optional = true }
 openapiv3 = { version = "2", optional = true }
 serde_yaml_ng = { version = "0.10", optional = true }
 sha2 = "0.11"
-ureq = "3.3.0"
+ureq = "3.3"
 
 [features]
 default = []
--- a/rapina-cli/src/commands/doctor.rs
+++ b/rapina-cli/src/commands/doctor.rs
@@ -34,9 +34,8 @@ pub fn execute(config: DoctorConfig) -> Result<(), String> {
     );
     println!();
 
-    let agent = ureq::Agent::new_with_defaults();
-    let routes = fetch_json(&agent, &urls::build_routes_url(&config.host, config.port))?;
-    let openapi = fetch_json(&agent, &urls::build_openapi_url(&config.host, config.port));
+    let routes = fetch_json(&urls::build_routes_url(&config.host, config.port))?;
+    let openapi = fetch_json(&urls::build_openapi_url(&config.host, config.port));
 
     let mut result = DiagnosticResult {
         warnings: Vec::new(),
@@ -356,8 +355,8 @@ fn print_results(result: &DiagnosticResult) {
 }
 
 /// Fetch JSON from URL.
-fn fetch_json(agent: &ureq::Agent, url: &str) -> Result<Value, String> {
-    let mut response = agent.get(url).call().map_err(|e| {
+fn fetch_json(url: &str) -> Result<Value, String> {
+    let mut response = crate::common::AGENT.get(url).call().map_err(|e| {
         format!(
             "Failed to fetch data. Is the server running? ({}) ({})",
             url, e
@@ -613,23 +612,21 @@ mod tests {
             use std::io::Read;
             let _ = stream.read(&mut buf);
             let response = format!(
-                "HTTP/1.1 200 OK\r\nContent-Length: {}\r\n\r\n{}",
+                "HTTP/1.1 200 OK\r\nContent-Type: application/json\r\nContent-Length: {}\r\n\r\n{}",
                 json.len(),
                 json
             );
             use std::io::Write;
             stream.write_all(response.as_bytes()).unwrap();
         });
-        let agent = ureq::Agent::new_with_defaults();
-        let result = fetch_json(&agent, &format!("http://127.0.0.1:{}", port));
+        let result = fetch_json(&format!("http://127.0.0.1:{}", port));
         handle.join().unwrap();
         assert_eq!(result.unwrap(), serde_json::json!({"status": "ok"}));
     }
 
     #[test]
     fn test_fetch_json_connection_refused() {
-        let agent = ureq::Agent::new_with_defaults();
-        let result = fetch_json(&agent, "http://127.0.0.1:1");
+        let result = fetch_json("http://127.0.0.1:1");
         assert!(result.is_err());
     }
 }
--- a/rapina-cli/src/commands/doctor.rs
+++ b/rapina-cli/src/commands/doctor.rs
@@ -4,7 +4,6 @@ use crate::commands::agents::{DriftStatus, check_drift, fix_agents, simple_diff}
 use crate::common::urls;
 use colored::Colorize;
 use serde_json::Value;
-use std::process::Command;
 
 struct DiagnosticResult {
     warnings: Vec<String>,
@@ -35,8 +34,9 @@ pub fn execute(config: DoctorConfig) -> Result<(), String> {
     );
     println!();
 
-    let routes = fetch_json(&urls::build_routes_url(&config.host, config.port))?;
-    let openapi = fetch_json(&urls::build_openapi_url(&config.host, config.port));
+    let agent = ureq::Agent::new_with_defaults();
+    let routes = fetch_json(&agent, &urls::build_routes_url(&config.host, config.port))?;
+    let openapi = fetch_json(&agent, &urls::build_openapi_url(&config.host, config.port));
 
     let mut result = DiagnosticResult {
         warnings: Vec::new(),
@@ -356,21 +356,17 @@ fn print_results(result: &DiagnosticResult) {
 }
 
 /// Fetch JSON from URL.
-fn fetch_json(url: &str) -> Result<Value, String> {
-    let output = Command::new("curl")
-        .args(["-s", "-f", url])
-        .output()
-        .map_err(|e| format!("Failed to run curl: {}", e))?;
-
-    if !output.status.success() {
-        return Err(format!(
-            "Failed to fetch data. Is the server running? ({})",
-            url
-        ));
-    }
-
-    let body =
-        String::from_utf8(output.stdout).map_err(|e| format!("Invalid UTF-8 response: {}", e))?;
+fn fetch_json(agent: &ureq::Agent, url: &str) -> Result<Value, String> {
+    let mut response = agent.get(url).call().map_err(|e| {
+        format!(
+            "Failed to fetch data. Is the server running? ({}) ({})",
+            url, e
+        )
+    })?;
+    let body = response
+        .body_mut()
+        .read_to_string()
+        .map_err(|e| format!("Failed to read response body: {}", e))?;
 
     serde_json::from_str(&body).map_err(|e| format!("Invalid JSON response: {}", e))
 }
@@ -604,5 +600,36 @@ mod tests {
         check_openapi_metadata(&openapi, &mut result);
         assert!(result.warnings.is_empty());
         assert_eq!(result.passed.len(), 1);
+    }
+
+    #[test]
+    fn test_fetch_json_success() {
+        let json = r#"{"status":"ok"}"#;
+        let listener = std::net::TcpListener::bind("127.0.0.1:0").unwrap();
+        let port = listener.local_addr().unwrap().port();
+        let handle = std::thread::spawn(move || {
+            let (mut stream, _) = listener.accept().unwrap();
+            let mut buf = [0; 4096];
+            use std::io::Read;
+            let _ = stream.read(&mut buf);
+            let response = format!(
+                "HTTP/1.1 200 OK\r\nContent-Length: {}\r\n\r\n{}",
+                json.len(),
+                json
+            );
+            use std::io::Write;
+            stream.write_all(response.as_bytes()).unwrap();
+        });
+        let agent = ureq::Agent::new_with_defaults();
+        let result = fetch_json(&agent, &format!("http://127.0.0.1:{}", port));
+        handle.join().unwrap();
+        assert_eq!(result.unwrap(), serde_json::json!({"status": "ok"}));
+    }
+
+    #[test]
+    fn test_fetch_json_connection_refused() {
+        let agent = ureq::Agent::new_with_defaults();
+        let result = fetch_json(&agent, "http://127.0.0.1:1");
+        assert!(result.is_err());
     }
 }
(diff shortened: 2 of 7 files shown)


rapina-cli/Cargo.toml
```diff
@@ -37,6 +37,7 @@ sqlx = { version = "0.8", optional = true }
 openapiv3 = { version = "2", optional = true }
 serde_yaml_ng = { version = "0.10", optional = true }
 sha2 = "0.11"
+ureq = "3.3.0"
```
richasdev: The ureq version is pinned to an exact version (`"3.3.0"`) rather than a semver range. Consider using `"3"` or `"3.3"` to allow minor/patch upgrades automatically, which is the idiomatic pattern in this file (see `openapiv3 = { version = "2", ... }` above). Pinning the exact version in `Cargo.toml` is unusual unless there is a known breakage in newer releases.

rapina-cli/src/commands/doctor.rs
```diff
@@ -356,21 +356,17 @@ fn print_results(result: &DiagnosticResult) {
...
-            "Failed to fetch data. Is the server running? ({})",
-            url
-        ));
-    }
-
-    let body =
-        String::from_utf8(output.stdout).map_err(|e| format!("Invalid UTF-8 response: {}", e))?;
+fn fetch_json(agent: &ureq::Agent, url: &str) -> Result<Value, String> {
+    let mut response = agent.get(url).call().map_err(|e| {
+        format!(
```
richasdev, line 360: `mut` on `response` is needed here only because `body_mut()` takes `&mut self`. This is correct, but worth noting that ureq 3.x returns a `Response` that owns the body — storing it as `mut` for two-line use is fine. No action needed; just documenting the pattern.

rapina-cli/src/commands/doctor.rs
```diff
@@ -356,21 +356,17 @@ fn print_results(result: &DiagnosticResult) {
...
+            "Failed to fetch data. Is the server running? ({}) ({})",
+            url, e
+        )
+    })?;
+    let body = response
+        .body_mut()
```
richasdev, line 366: `read_to_string()` buffers the entire body in memory. For the diagnostic endpoints served locally this is fine, but if the OpenAPI spec is very large this could be noticeable. A size cap (e.g. `read_to_string_limit(1_000_000)`) would be more defensive. Same pattern applies in `llms.rs`, `openapi.rs`, and `routes.rs`.

rapina-cli/src/commands/doctor.rs
```diff
@@ -605,4 +601,35 @@ mod tests {
...
     }
+
+    #[test]
+    fn test_fetch_json_success() {
+        let json = r#"{"status":"ok"}"#;
+        let listener = std::net::TcpListener::bind("127.0.0.1:0").unwrap();
+        let port = listener.local_addr().unwrap().port();
+        let handle = std::thread::spawn(move || {
+            let (mut stream, _) = listener.accept().unwrap();
+            let mut buf = [0; 4096];
```
richasdev, line 611: The raw TCP server in this test does not send a `Content-Type: application/json` header. ureq 3.x may or may not require it, but since the callers do not check the content-type and just call `read_to_string`, this works today. However, a more realistic mock would include `Content-Type: application/json\r\n` in the response headers so the test accurately reflects production behaviour and does not silently skip content-type validation if that is ever added.

rapina-cli/src/commands/doctor.rs
```diff
@@ -605,4 +601,35 @@ mod tests {
...
+            use std::io::Read;
+            let _ = stream.read(&mut buf);
+            let response = format!(
```
richasdev, line 614: `stream.read(&mut buf)` reads a single chunk of the request, which is fine for small HTTP requests from ureq. However, if ureq ever sends the request in multiple `write` calls (e.g. due to pipelining or a large body), only part of the request headers would be consumed before the response is written. For correctness and future-proofing, read until `\r\n\r\n` or use a proper HTTP mini-server. This risk is low in tests but can cause flaky failures under load or on slow CI.

rapina-cli/src/commands/doctor.rs
```diff
@@ -605,4 +601,35 @@ mod tests {
...
+                json.len(),
+                json
+            );
+            use std::io::Write;
+            stream.write_all(response.as_bytes()).unwrap();
+        });
+        let agent = ureq::Agent::new_with_defaults();
+        let result = fetch_json(&agent, &format!("http://127.0.0.1:{}", port));
+        handle.join().unwrap();
+        assert_eq!(result.unwrap(), serde_json::json!({"status": "ok"}));
```
richasdev, line 624: `handle.join()` is called *after* `fetch_json`, meaning if `fetch_json` panics the thread is never joined and resources leak in the test. Consider joining inside a `defer`-like guard or restructuring to `let result = fetch_json(...); handle.join().unwrap(); result` (which is what the code does — this ordering is correct). However, if `fetch_json` returns `Err` and the test were to `unwrap` before joining, the thread would be orphaned. The current ordering is fine — just worth a comment.

rapina-cli/src/commands/llms.rs
```diff
@@ -22,19 +22,54 @@ pub fn export(output: Option<String>, host: &str, port: u16) -> Result<(), Strin
...
+            url, e
+        )
+    })?;
+    let body = response
+        .body_mut()
+        .read_to_string()
+        .map_err(|e| format!("Failed to read response body: {}", e))?;
+
+    Ok(body)
+}
```
richasdev, line 38: `Ok(body)` can be simplified by removing the intermediate binding. Since `read_to_string()` already returns a `Result<String, _>` and we `map_err` it above, the function body could end with `.map_err(...)` directly and return `body` without wrapping in `Ok`. Minor style nit — the current code is not wrong.

rapina-cli/src/commands/llms.rs
```diff
@@ -22,19 +22,54 @@ pub fn export(output: Option<String>, host: &str, port: u16) -> Result<(), Strin
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_fetch_llms_txt_success() {
+        let body = "User-agent: *\nDisallow: /admin";
+        let listener = std::net::TcpListener::bind("127.0.0.1:0").unwrap();
```
richasdev, line 47: The test body `"User-agent: *\nDisallow: /admin"` uses a bare `\n`. When `Content-Length` is calculated via `body.len()`, this is correct. But note that the actual file served by the application likely uses `\r\n` line endings (as is conventional in HTTP). If `fetch_llms_txt` were ever changed to validate or normalise line endings, this test would need updating. Low risk, but worth noting.

rapina-cli/src/commands/openapi.rs
```diff
@@ -7,7 +7,8 @@ use std::process::Command;
```
richasdev, line 7: `use std::process::Command` is still imported at the top of this file (line 7 of the original, retained as context line in the diff at position 1). After this refactor, `Command` is only used in `get_spec_from_branch` (which calls `git`). That usage is intentional and correct, but it is worth verifying with `cargo clippy` that no unused-import warning is emitted (it should not be, since `Command` is still used).

rapina-cli/src/commands/openapi.rs
```diff
@@ -34,8 +35,10 @@ pub fn check(file: &str, host: &str, port: u16) -> Result<(), String> {
     let committed_json: Value =
         serde_json::from_str(&committed).map_err(|e| format!("Failed to parse {}: {}", file, e))?;
 
+    let agent = ureq::Agent::new_with_defaults();
+
```
richasdev: `ureq::Agent::new_with_defaults()` is constructed separately in each of the three public functions (`export`, `check`, `diff`). Each call creates a new connection pool. For the CLI use-case this is acceptable since only one endpoint is called per invocation, but if these are ever called in sequence within the same process, sharing a single agent would be more efficient. Consider accepting the agent from the caller or constructing it once at the top of the call-site.

rapina-cli/src/commands/routes.rs
```diff
@@ -66,21 +66,56 @@ pub fn execute(config: RoutesConfig) -> Result<(), String> {
 }
 
 /// Fetch routes from running application.
-fn fetch_routes(url: &str) -> Result<Vec<RouteInfo>, String> {
-    let output = Command::new("curl")
-        .args(["-s", "-f", url])
-        .output()
-        .map_err(|e| format!("Failed to run curl: {}", e))?;
+fn fetch_routes(agent: &ureq::Agent, url: &str) -> Result<Vec<RouteInfo>, String> {
+    let mut response = agent.get(url).call().map_err(|e| {
```
richasdev: Nit: the `agent` parameter precedes `url` in `fetch_routes(agent, url)` but in `fetch_json` (doctor.rs) the same convention is followed. This is consistent across the PR. Good.

rapina-cli/src/commands/routes.rs
```diff
@@ -66,21 +66,56 @@ pub fn execute(config: RoutesConfig) -> Result<(), String> {
...
-        String::from_utf8(output.stdout).map_err(|e| format!("Invalid UTF-8 response: {}", e))?;
+#[cfg(test)]
+mod tests {
+    use super::*;
 
-    serde_json::from_str(&body).map_err(|e| format!("Invalid JSON response: {}", e))
+    #[test]
+    fn test_fetch_routes_success() {
+        let json = r#"[{"method":"GET","path":"/users","handler_name":"list_users"}]"#;
+        let listener = std::net::TcpListener::bind("127.0.0.1:0").unwrap();
```
richasdev, line 90: The test JSON string uses `handler_name` as a field, but `RouteInfo` in the same file must deserialize it. Make sure the struct field name matches (or has a `#[serde(rename = "handler_name")]` attribute). If it does not, `serde_json::from_str` will silently ignore the field and the assertion `routes[0].handler_name` (if it existed) would fail. Since the test only asserts `method` and `path`, a mismatch in `handler_name` would be invisible. Consider adding an assertion for `handler_name` if it is part of the struct.

richasdev (changes requested):

rapina-cli/src/commands/openapi.rs
ShiraiEd: This still being used for `git` as you said, and the CI check did not fail on this so its ok

rapina-cli/src/commands/openapi.rs
ShiraiEd: For now this maybe be fine,
we can leave a comment telling something like
"if ever calls these fns in the same process, consider build one ureq client to share istead of creating new ones in each command"

rapina-cli/src/commands/llms.rs
ShiraiEd: either is fine,
the way it is now is more readable though

rapina-cli/src/commands/doctor.rs
ShiraiEd: the result is unwrapped after handle.join, so the thread would be already closed when the panic happens

rapina-cli/src/commands/llms.rs
ShiraiEd: For this one I think is ok,
if `fetch_llms_txt` ever needs to validate/normalize line endings, we could write a separated tests for that

rapina-cli/src/commands/doctor.rs
ShiraiEd: I dont think the spec would ever reach that amount for this to be necessary,
but security is never good to dismiss.

rapina-cli/src/commands/doctor.rs
ShiraiEd: this will just read
```rust
let json = r#"{"status":"ok"}"#;
```
so that may not be necessary

rapina-cli/src/commands/doctor.rs
Ismaellima4: The method `.read_to_string()` is limited by default to `10MB`.

Since I couldn't find the `read_to_string_limit` method in the `ureq` API, what does this `1_000_000` represent? If it's bytes, can I limit it using

```rust
  let body = response
      .body_mut()
      .with_config()
      .limit(1024 * 1024) // 1MB
      .read_to_string()
      .map_err(|e| format!("Failed to read response body: {}", e))?;
```

rapina-cli/src/commands/routes.rs
Ismaellima4: the field name is correct, I will add it to the assert

rapina-cli/src/commands/doctor.rs
ShiraiEd: I think the default one is fine

ShiraiEd (dismissed):
Well done Ismael

LGTM

richasdev (dismissed):
LGTM!

rapina-cli/src/commands/routes.rs
```diff
@@ -67,20 +66,54 @@ pub fn execute(config: RoutesConfig) -> Result<(), String> {
...
+        assert_eq!(routes.len(), 1);
+        assert_eq!(routes[0].method, "GET");
+        assert_eq!(routes[0].path, "/users");
+        assert_eq!(routes[0].handler_name, "list_users");
+    }
+
+    #[test]
+    fn test_fetch_routes_connection_refused() {
+        let result = fetch_routes("http://127.0.0.1:1");
+        assert!(result.is_err());
```
arferreira, line 117: I think you should assert not only port which is a lib behavior, but also mapped message so the test will covers our code.

arferreira, line 117:
```suggestion
        assert!(result.is_err());
        assert!(result.unwrap_err().contains("Failed to fetch"));
```

arferreira (dismissed):
Good work, thanks! I left just one suggestion for a specific test, I see improvements here and there but nothing that blocks merge.

arferreira (approved):

PR #587, opened 2026-05-20
Add --llms flag to output routes as llms.txt Markdown

### LINKED ISSUE #582: [Feature]: Add --llms flag to rapina routes
### Problem Statement

Mentioned in the end of the issue:
https://github.com/rapina-rs/rapina/issues/522#issue-4306247758
We could discuss about the flags, it could be something like `--format table` for default output and `--format llms` for this feature , but i chose to just implement `--llms` since `rapina routes` already prints the table, so no need for `--format table` for now

`rapina routes` currently prints a formatted table of registered routes. `to_llms_txt already` exists in `rapina/src/introspection/llms.rs` and produces full Markdown document with routes, schemas, and error codes but it's only accessible by fetching `/__rapina/llms.txt` from a running server.

Adding a `--llms` flag to `rapina routes` would let users pipe or redirect the llms.txt output directly from the CLI without needing to curl the separately. `--llms` outputs the same Markdown that the endpoint serves.

Note: rapina routes currently also shells out to curl, this may be fixed by the time this issue is picked up (see issue #579). Either way, the --llms flag should be added regardless.

  Relevant Files

  - rapina-cli/src/commands/routes.rs — add --llms to RoutesConfig and a new output branch
  - rapina-cli/src/main.rs:75 — CLI argument wiring for the routes subcommand
  - rapina/src/introspection/llms.rs:18 — to_llms_txt(title, routes), the formatter to reuse
  - rapina/src/introspection/route_info.rs — RouteInfo struct, already deserialized in routes.rs

  Acceptance Criteria

  - [ ] Fix is implemented
  - [ ] rapina routes defaults to the existing table output when --llms is omitted
  - [ ] rapina routes --llms outputs the same Markdown as /__rapina/llms.txt
  - [ ] Documentation updated if needed
  - [ ] cargo fmt and cargo clippy pass

### Proposed Solution

Add a --llms flag to RoutesConfig and wire it into main.rs. When llms is selected, reuse to_llms_txt from rapina/src/introspection/llms.rs instead of printing the table. The routes fetch itself stays the same

### Alternatives Considered

- A separate rapina llms routes subcommand: rejected to avoid duplicating fetch logic
- Fetching directly from /__rapina/llms.txt: rejected because the endpoint is off in release builds by default, whereas --llms generates from the registry response and always works regardless

### Contribution

richasdev:
## Summary

- Add --llms flag to Routes subcommand in main.rs
- Extend RouteInfo with schema fields (all optional via serde default)
- Add to_llms_txt() in routes.rs, filtering internal /__rapina routes
- Default table output unchanged when --llms is omitted
- Add 11 unit tests covering all branches of to_llms_txt

## Related Issues

Closes #582

## Checklist

- [x] `cargo fmt` passes
- [x] Tests pass


ShiraiEd (changes requested):
nice,

for the `to_llms_txt` I ask you to use the one from `introspections/llms.rs` instead of duplicating the logic

Once that is imported, the tests regarding the version in here will become redundant because the function is already tested in llms.rs

ShiraiEd:
you did a nice work Richas,
but looking closer this new command really doesn't do anything different from `rapina llms export`
my bad for didn't noticing that earlier and creating the issue

PR #596, opened 2026-05-26
Feature: add handler descriptions to llms.txt output

### LINKED ISSUE #585: [Feature]: Add handler descriptions to llms.txt output
### Problem Statement

The `llms.txt` output in https://github.com/rapina-rs/rapina/issues/522#issue-4306247758 sketched handler descriptions under each route:
```
### POST /v1/users
Create a new user. Requires JWT.
```

But currently `RouteInfo` carries no description about the route,
Issue #522 noted this explicitly: 

"RouteInfo.handler_name gives us create_user and nothing richer. Output will look thin for non-trivial routes until we have real handler descriptions. Either rustdoc extraction or a macro attribute, either way its own scope and its own issue."

## Two approaches suggested:

- Rustdoc extraction: read /// doc comments above the handler at compile time inside the proc macro and pass them through to RouteInfo. Zero new user-facing API, but relies on users already documenting their handlers.
- Explicit macro attribute: add a description argument to the existing route macros: #[post("/users", description = "Create a new user")]. Opt-in, explicit, but adds friction.

## Relevant Files

- rapina/src/introspection/route_info.rs
- rapina/src/introspection/llms.rs
- rapina-macros/src/lib.rs (if you consider changing the macro)

Open Design Questions

- Rustdoc extraction or explicit description = "..." attribute?
- If rustdoc, what happens when the doc comment has multiple lines or markdown?
- If explicit attribute, is it on the route macro or a separate #[description("...")]?

### Proposed Solution

Add a description: `Option<String>` field to `RouteInfo` and populate it at compile time in the proc macro either by reading /// doc comments above the handler function or via an explicit description = "..." argument on the route macro.
to_llms_txt renders it between the route heading and the request block when present. When absent, fall back to the handler name (create_user → "Create user"), this is possible because it is the convention mentioned in the docs and in AGENTS.md:
- core.md: "Handler names follow verb_resource convention" with examples like list_todos, get_todo, create_todo
- project-structure.md: "Handlers are verb + resource: create_user, list_users, get_user, delete_user"

### Alternatives Considered

- Runtime annotation: a separate #[description("...")] attribute instead of adding it to the existing route macros. 
- Rustdoc-only: require users to write /// doc comments, no explicit attribute. Clean but silently produces no output for undocumented handlers, which may go unnoticed for a long time.
- No fallback: returns nothing when description is absent. Would be better to return at least the handler name in the humanized version mentioned before

### Contribution

richasdev:
## Summary

Implements #585. Adds human-readable descriptions to each route entry in `llms.txt` output, with three resolution layers:

1. **Explicit attribute** — `#[get("/users", description = "List all users")]` — highest priority
2. **Rustdoc extraction** — first non-empty `///` line above the handler — zero extra syntax for documented handlers
3. **Humanized fallback** — `create_user` → `"Create user"` — ensures every route has a description even without any annotation

The description is rendered between the route heading and the request block:

\`\`\`
### GET /users

List all users

Request (application/json):
...
\`\`\`

**Changes:**
- `RouteInfo` and `RouteDescriptor` gain a `description` field
- `Handler` trait gains a `description()` default method (returns `None`)
- Proc macro parses `description = "..."` attribute and extracts rustdoc via `extract_doc_description()`
- `to_llms_txt` renders the description; `humanize_handler_name` provides the fallback
- All existing `RouteInfo::new()` call sites updated (new 10th parameter)

## Related Issues

Closes #585

## Checklist

- [x] `cargo fmt` passes
- [x] Tests pass

### DIFF
--- a/docs/content/docs/core-concepts/extractors.md
+++ b/docs/content/docs/core-concepts/extractors.md
@@ -7,22 +7,37 @@ date = 2025-02-13
 
 Extractors automatically parse request data and inject it into your handlers. If parsing fails, they return appropriate error responses.
 
+## Handler Attributes
+
+The route macros (`#[get]`, `#[post]`, etc.) accept optional attributes alongside the path:
+
+| Attribute     | Type   | Description                                              |
+| ------------- | ------ | -------------------------------------------------------- |
+| `description` | `&str` | Human-readable description included in `llms.txt` output |
+
+```rust
+#[get("/users", description = "List all users")]
+async fn list_users() -> Json<Vec<User>> { ... }
+```
+
+See [llms.txt](/docs/core-concepts/llms-txt/#handler-descriptions) for the full description resolution order.
+
 ## Available Extractors
 
-| Extractor | Description |
-|-----------|-------------|
-| [`Path<T>`](#path-parameters) | URL path parameters |
-| [`Query<T>`](#query-parameters) | Query string parameters |
-| [`Json<T>`](#json-body) | JSON request body |
-| [`Form<T>`](#form-data) | URL-encoded form data |
-| [`Headers`](#headers) | Request headers |
-| [`State<T>`](#application-state) | Application state |
-| [`Context`](#request-context) | Request context (trace_id) |
-| [`Cookie<T>`](#cookies) | Typed cookie access |
-| [`CurrentUser`](#currentuser) | Authenticated user (JWT) |
-| [`Validated<T>`](#validation) | Validated extractor |
-| [`Paginate`](#paginate) | Pagination params (requires feature) |
-| [`Db`](#db) | Database connection (requires feature) |
+| Extractor                        | Description                            |
+| -------------------------------- | -------------------------------------- |
+| [`Path<T>`](#path-parameters)    | URL path parameters                    |
+| [`Query<T>`](#query-parameters)  | Query string parameters                |
+| [`Json<T>`](#json-body)          | JSON request body                      |
+| [`Form<T>`](#form-data)          | URL-encoded form data                  |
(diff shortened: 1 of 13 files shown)


rapina/src/introspection/route_info.rs
```diff
@@ -191,8 +201,79 @@ mod tests {
...
+            None,
+            None,
+            None::<String>,
+            None,
+            Vec::new(),
+            Vec::new(),
+            None::<String>,
+        );
+        let json = serde_json::to_string(&info).unwrap();
+        assert!(!json.contains("description"));
```
ShiraiEd: nit

To keep the pattern, this should be tested like the one above, like:
```rust
assert!(!json.contains("\"description\""));
```

ShiraiEd (changes requested):
Nice!

I left one comment in `route_info.rs`

Also, this should be added to the docs, in the `llms.txt` and probably in `extractors` too

docs/content/docs/core-concepts/extractors.md
ShiraiEd, line 1: I read through this and it doesn't seem to fit that well here,
I know I suggested to include in `Extractors` but could you live this unchanged and
add in `Routing` instead, right after `route groups`

It was my error to suggested the wrong file, sorry man

ShiraiEd (changes requested):
`route_info.rs` and `llms.txt` were well addressed

just a comment in `Extractors`

docs/content/docs/core-concepts/extractors.md
richasdev: No worries at all, thanks for catching that! I'll move it to Routing right after route groups.

ShiraiEd (approved):
LGTM

PR #607, opened 2026-06-03
Add OpenTelemetry OTLP tracing exporter

### LINKED ISSUE #93: Add OpenTelemetry tracing exporter

arferreira:
Fixes #93

Adds an `otel` feature that exports traces over OTLP gRPC to a collector like Jaeger or Datadog. `with_telemetry(TelemetryConfig { endpoint, service_name, sample_rate })` composes a `tracing-opentelemetry` layer onto the same subscriber as the stdout `fmt` layer. Incoming W3C `traceparent` headers continue the trace across services, request spans follow the HTTP semantic conventions and record the response status, the OTel `trace_id`/`span_id` land on the request span so logs correlate with the trace, and spans flush on graceful shutdown. Export is plaintext by default. Enable `otel-tls` for system-root TLS to reach `https://` collectors. Because a global subscriber can only be set once, `with_tracing` now installs at `listen()` instead of at call time so the two layers compose.

### DIFF
--- a/rapina/src/middleware/otel.rs
+++ b/rapina/src/middleware/otel.rs
@@ -7,7 +7,7 @@ use tracing::field::Empty;
 use tracing::{Instrument, info_span};
 use tracing_opentelemetry::OpenTelemetrySpanExt;
 
-use crate::context::RequestContext;
+use crate::context::{MatchedPattern, RequestContext};
 use crate::response::BoxBody;
 
 use super::{BoxFuture, Middleware, Next};
@@ -57,7 +57,11 @@ impl Middleware for TraceContextMiddleware {
         let parent_cx = global::get_text_map_propagator(|propagator| {
             propagator.extract(&HeaderExtractor(req.headers()))
         });
-        let span = server_span(req.method(), req.uri().path(), parent_cx);
+        let route = req
+            .extensions()
+            .get::<MatchedPattern>()
+            .map(|m| m.as_str().to_owned());
+        let span = server_span(req.method(), req.uri().path(), route.as_deref(), parent_cx);
 
         let response_span = span.clone();
         Box::pin(
@@ -75,24 +79,33 @@ impl Middleware for TraceContextMiddleware {
 /// conventions, links it to the remote parent, and records the OTel ids onto it
 /// so log lines emitted within the span correlate with the exported trace.
 ///
-/// The route template isn't known before routing, so the span name is the
-/// method per the spec's low-cardinality fallback.
+/// Named `{method} {route}` when the router matched a template, plain method
+/// otherwise per the spec's low-cardinality fallback.
 fn server_span(
     method: &hyper::Method,
     path: &str,
+    route: Option<&str>,
     parent_cx: opentelemetry::Context,
 ) -> tracing::Span {
+    let name = match route {
+        Some(route) => format!("{method} {route}"),
+        None => method.to_string(),
+    };
     let span = info_span!(
         "http.request",
-        otel.name = %method,
+        otel.name = %name,
         otel.kind = "server",
         otel.status_code = Empty,
         http.request.method = %method,
+        http.route = Empty,
         url.path = %path,
         http.response.status_code = Empty,
         otel_trace_id = Empty,
         otel_span_id = Empty,
     );
+    if let Some(route) = route {
+        span.record("http.route", route);
+    }
     let _ = span.set_parent(parent_cx);
 
     let span_context = span.context().span().span_context().clone();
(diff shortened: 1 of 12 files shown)


rapina/src/middleware/otel.rs
```diff
@@ -0,0 +1,202 @@
...
+/// The route template isn't known before routing, so the span name is the
+/// method per the spec's low-cardinality fallback.
+fn server_span(
+    method: &hyper::Method,
+    path: &str,
+    parent_cx: opentelemetry::Context,
+) -> tracing::Span {
+    let span = info_span!(
+        "http.request",
+        otel.name = %method,
```
ShiraiEd: According to the convention, `name` should be the router template, like `GET /users/:id` for example,
worth a follow-up?

ShiraiEd (commented):
Nice work

i have left a comment on `server_span`

It is also missing the docs, a `telemetry.md` or `observabiity.md`  as a new doc file containing the app method, the config, the feature flag, sample rate, etc, would be nice

ShiraiEd:
also noticed that `with_tracing` that was modified here has no docs mention,
that user on discord found no info about this
this could left to a follow-up

rapina/src/middleware/otel.rs
arferreira: No follow-up needed, the router resolves the pattern before the middleware chain runs and puts it in request extensions as `MatchedPattern`.

rapina/src/observability/tracing.rs
```diff
@@ -81,26 +83,52 @@ impl TracingConfig {
...
+        .unwrap_or_else(|_| EnvFilter::new(config.level.to_string()));
+
+    let mut layers: Vec<Box<dyn Layer<Registry> + Send + Sync>> = vec![fmt_layer(&config)];
+    if let Some(layer) = extra_layer {
+        layers.push(layer);
     }
+
+    Registry::default()
+        .with(layers.with_filter(filter))
+        .try_init()
```
ShiraiEd, line 131: ### Non-blocking-comment

In this, the filter is applied on both tracing and otel, there is no distinction between each layer,
If a user wants to blocks `info spans` in tracing but pass them to otel for example, it has no ways to it

Since `extra_layer` is already an Option, and `tracing_subscriber` implements Layer for Option<L>
you could drop the `Vec` and go with options instead, like:

```rust
let layer = fmt_layer(&config);

Registry::default()
      .with(layer.with_filter(filter))
      .with(extra_layer)
      .try_init()
```

ShiraiEd (commented):
Nice docs


Just one more thing that I noticed on the filter in tracing.rs

rapina/src/observability/tracing.rs
arferreira: I see two problems with this:

First, moving the filter onto only the fmt layer leaves `extra_layer` unfiltered, right? so otel would suddenly receive every span regardless of the configured level. It can flood the collector...

Second, the reason a single shared filter exists in the first place is almost certainly that `EnvFilter` isn't `Clone`. you can just hand the same filter to two layers, to give each layer it's own filter we'd have to parse the directive string twice and build two `EnvFilter`, or I don't know, switch to a cheaper `Targets` filter.. Yeah, this is a real design decision, I don't believe it's just a one line refactor.

ShiraiEd: For the single filter using the RUST_LOG env var I think we could leave like that for now,
  one filter defined in env for all the layers

  What we could do to add separated filters for the layers when there is none RUST_LOG env:
  1. create a `level` filed inside the `TelemetryConfig`, like the `TracingConfig`,
  2. inside `build_otel_pipeline` we could check if there is a `RUST_LOG` var, if yes use that if not use the `config.level` and return a layer with the filter already applied.
  3. and for the tracing filter, it could be applied inside the `fmt_layer` function, returning a layer with the filter applied and making the whole filter approach more consistent with each fn dealing with it's own layers, or leave on `init_subscriber` to apply it just on the formatted layer like the current way, that works too
  
  but this is just an suggestion to add more stuff in dev mode, we could leave for later

ShiraiEd: lets leave this for another issue then

ShiraiEd (approved):
My comment was out of the scope for this

Ship It!

PR #617, opened 2026-06-08
Add TLS certificate support for Redis

### LINKED ISSUE #271: Support TLS and client certificates for Redis cache backend
The Redis cache backend currently connects with `redis::Client::open(url)`, which only supports system root certificates when using `rediss://`. No way to provide custom CA certs or client certificate authentication.

Should support `redis::Client::build_with_tls` for environments that require it (internal PKI, mutual TLS, etc). This needs the `rust-tls` feature on the redis crate.

Ref: https://docs.rs/redis/latest/redis/struct.Client.html#method.build_with_tls

From https://github.com/arferreira/rapina/pull/268#discussion_r2867787287

ShiraiEd:
## Summary

- I changed `tokio-comp` to `tokio-rustls-comp` because just `rust-tls` was returning an impl error inside the crate itself
- Created the `RedisTlsConfig` to carry the certificates
- Added the `tls` field in the `CacheConfig` enum
- Added the `tls` param in `fn connect`, and on the same function implemented a url check so when `rediss://` is used, it accepts the certicates
- Added a test to verify if the connection returns an error if the cert is passed but key is missing
- Updated the cache docs

## Related Issues

Closes #271

## Checklist

- [x] `cargo fmt` passes
- [x] Tests pass


ShiraiEd:
I considered creating a new App Method like `.with_redis_tls` to separate the redis with and without the certs
but decided to go with this for the feedback first

also I am feeling that my approach to check if is a `rediss://` conn looks kinda bad

docs/content/docs/core-concepts/caching.md
```diff
@@ -153,8 +153,19 @@ Rapina::new()
...
+If you also want to use a certificate alongside the redis, pass it after the url like this:
+
+```rust
+CacheConfig::redis("rediss://localhost", Some(RedisTlsConfig {
+      ca_cert: Some(std::fs::read("ca.pem")?),
+      client_cert: Some(std::fs::read("client.pem")?),
+      client_key: Some(std::fs::read("client.key")?),
+  }))
+  ```
+___
```
arferreira: Cool!

rapina/src/cache_redis.rs
```diff
@@ -52,8 +52,38 @@ pub struct RedisCache {
...
+                }
+            };
+            let tls_certs = TlsCertificates {
+                client_tls,
+                root_cert: tls.ca_cert,
+            };
+            redis::Client::build_with_tls(url, tls_certs)?
+        } else {
+            redis::Client::open(url)?
+        };
```
arferreira: Well I think this could be more elegant.
`contains` matches the scheme anywhere in the string, not just the front. Also this removes the `unwrap_or` with an all-None config and fixes a silent failure: today, passing a `RedisTlsConfig` with a `redis://` URL drops the certificates without a word. Worth rejecting that combination explicitly while you're here, certs on a plaintext URL is always a misconfiguration.
```suggestion
        let client = match tls {
    Some(tls) => {
        let client_tls = match (tls.client_cert, tls.client_key) {
            (Some(client_cert), Some(client_key)) => {
                Some(ClientTlsConfig { client_cert, client_key })
            }
            (None, None) => None,
            _ => {
                return Err(redis::RedisError::from((
                    redis::ErrorKind::InvalidClientConfig,
                    "client_cert and client_key must be provided together",
                )));
            }
        };
        redis::Client::build_with_tls(url, TlsCertificates { client_tls, root_cert: tls.ca_cert })?
    }
    None => redis::Client::open(url)?,
};
```

rapina/src/cache.rs
```diff
@@ -191,9 +202,10 @@ impl CacheConfig {
 
     /// Creates a Redis cache configuration.
     #[cfg(feature = "cache-redis")]
-    pub fn redis(url: &str) -> Self {
+    pub fn redis(url: &str, tls: Option<RedisTlsConfig>) -> Self {
```
arferreira: Okay, this forces `CacheConfig::redis(url, None)` on every caller that doesn't use TLS right?
I think we could keep `redis(url)` as it is on main and add `redis_with_tls(url: &str, tls: RedisTlsConfig)` with no option, because if you're calling it you have a config, does it makes sense?

This would change docs as well, sorry.

rapina/src/cache_redis.rs
```diff
@@ -187,4 +217,19 @@ mod tests {
...
+            "rediss://localhost",
+            Some(RedisTlsConfig {
+                ca_cert: None,
+                client_cert: Some(b"cert".to_vec()),
+                client_key: None,
+            }),
+        )
+        .await;
+
+        assert!(result.is_err());
```
arferreira: IMO as written the test passes if connect fails for any reason at all.
```suggestion
        assert_eq!(
            result.unwrap_err().kind(),
            redis::ErrorKind::InvalidClientConfig
        );
```

rapina/src/cache.rs
```diff
@@ -174,13 +174,24 @@ impl CacheBackend for InMemoryCache {
     }
 }
 
+/// Configuration for Redis Tls
+#[cfg(feature = "cache-redis")]
+pub struct RedisTlsConfig {
+    pub ca_cert: Option<Vec<u8>>,
+    pub client_cert: Option<Vec<u8>>,
+    pub client_key: Option<Vec<u8>>,
```
arferreira, line 185: Two things here: there's a typo, should be TLS, also this doc comment says nothing useful, we could improve it.

arferreira (changes requested):
Hey thanks for picking this up! A few smaller changes to do, worth addressing those before merge.

ShiraiEd:
I will follow the new app method approach then, and do the changes

docs/content/docs/core-concepts/caching.md
```diff
@@ -118,7 +118,7 @@ rapina = { version = "0.12.0", features = ["cache-redis"] }
 use rapina::cache::CacheConfig;
 
 Rapina::new()
-    .with_cache(CacheConfig::redis("redis://localhost:6379")).await?
+    .with_cache(CacheConfig::redis("redis://localhost:6379", None)).await?
```
arferreira: must fix, this won't compile. `CacheConfig::redis` is single arg now, but the example passes two:
```suggestion
    .with_cache(CacheConfig::redis("redis://localhost:6379")).await?
```

docs/content/docs/core-concepts/caching.md
```diff
@@ -153,8 +153,19 @@ Rapina::new()
     .await
 ```
 
----
 
+### TLS Certificates
```
arferreira: Sorry, heading level.
```suggestion
#### TLS Certificates
```

docs/content/docs/core-concepts/caching.md
```diff
@@ -153,8 +153,19 @@ Rapina::new()
...
+If you also want to use a certificate alongside the redis, pass it after the url like this:
+
+```rust
+CacheConfig::redis_with_tls("rediss://localhost", RedisTlsConfig {
+    ca_cert: Some(std::fs::read("ca.pem")?),
+    client_cert: Some(std::fs::read("client.pem")?),
+    client_key: Some(std::fs::read("client.key")?),
+})
+```
+___
```
arferreira: I think you changed the separator for underscore
```suggestion
---
```

arferreira (changes requested):
Hey sorry, just noticed a few nits to be addressed before merge, I left inline suggestions.

arferreira (approved):
LGTM

PR #622, opened 2026-06-09
fix: allow entity types as primary keys in join tables

### LINKED ISSUE #619: [Bug]: [primary_key] rejects entity types in join table definitions (docs example doesn't compile)
### Description

The docs show entity-typed fields (e.g., user_id: User) as valid composite primary key columns in a join table. The macro rejects them at compile time with "must be a scalar type, not a relationship", so the documented pattern is impossible to use.

### Steps to Reproduce

Try to declare this schema:

```rust
schema! {
      #[primary_key(user_id, role_id)]
      #[timestamps(none)]
      UserRole {
          user_id: User,
          role_id: Role,
      }
  }
```

### Expected Behavior

The macro accepts entity-typed fields in #[primary_key], infers the FK column type from the target entity's PK, and uses the field name as the column name (no extra _id suffix appended).
(As shown in the docs)

### Code Example

```rust
(From the discord)

#[table_name = "transactions"]
Tx {
    ...
    labels: Vec<TxLabel>,
}

#[table_name = "labels"]
Label {
    #[unique]
    name: String,
}

#[table_name = "transaction_labels"]
#[timestamps(none)]
#[primary_key(tx_id, label_id)]
TxLabel {
    tx_id: Tx,
    label_id: Label,
}
```

### Environment

- Rapina version: latests
- Rust version: any
- OS: any

### Error Output / Logs

```shell
INFO Building project...
   Compiling server v0.1.0 (...)
error: primary_key column 'tx_id' must be a scalar type, not a relationship
  --> src/entities.rs:34:9
   |
34 |         tx_id: Tx,
   |         ^^^^^

error: could not compile `server` (bin "server") due to 1 previous error

Error: Build failed
```

mvanhorn:
## Summary

Allows entity-typed fields to be used as primary-key columns in `#[primary_key(...)]`, making the documented many-to-many join-table pattern compile. Previously the `schema!` macro rejected them at compile time with `primary_key column '...' must be a scalar type, not a relationship`, so the documented pattern was impossible to use.

The exact reproduction from the issue now compiles:

```rust
schema! {
    #[table_name = "transactions"]
    Tx { name: String }

    #[table_name = "labels"]
    Label { #[unique] name: String }

    #[table_name = "transaction_labels"]
    #[timestamps(none)]
    #[primary_key(tx_id, label_id)]
    TxLabel {
        tx_id: Tx,
        label_id: Label,
    }
}
```

What changed:

- `analyze.rs`: A required `belongs_to` field is now accepted as a custom primary-key column. The scalar-only check is relaxed but still rejects `has_many` (no column) and optional `belongs_to` (a primary key cannot be nullable), each with a clear compile-time message. Unknown-type and column-not-found errors are unchanged.
- `generate.rs`:
  - `generate_custom_pk_fields` emits a relationship PK column using the field name verbatim (e.g. `tx_id`, no extra `_id` suffix) and the scalar type inferred from the target entity's primary key.
  - The shared target-PK-type resolution is extracted into `resolve_target_pk_type` (reused by the existing belongs_to model-field path) and now resolves transitively when a target's PK is itself a relationship, with a cycle guard so self/mutually-referential relationship PKs terminate instead of recursing forever.
  - The generated relation `from` column for a relationship PK uses the verbatim PascalCase field name (`Column::TxId`, not `Column::TxIdId`), and the `to` column resolves to the target's actual single PK column instead of a hardcoded `Column::Id`.
- `rapina/tests/schema_test.rs`: Adds an end-to-end compile test for the issue's `TxLabel` join table, verifying the generated `Model` has verbatim PK columns, no auto `id`, and working `Relation` variants.

## Related Issues

Closes #619

## Checklist

- [x] `cargo fmt` passes
- [x] Tests pass (`cargo test -p rapina-macros`: 128 passed; `cargo test -p rapina --features database --test schema_test`: 6 passed)

### DIFF
--- a/rapina-macros/src/schema/generate.rs
+++ b/rapina-macros/src/schema/generate.rs
@@ -9,10 +9,6 @@ use super::types::{FieldType, ScalarType};
 
 /// Generate the complete schema code from analyzed entities.
 pub fn generate_schema(schema: AnalyzedSchema) -> TokenStream {
-    if let Some(error) = composite_pk_target_error(&schema) {
-        return error;
-    }
-
     let entity_modules: Vec<TokenStream> = schema
         .entities
         .iter()
@@ -143,42 +139,6 @@ fn build_field_attr(parts: &[TokenStream], column_type: Option<TokenStream>) -> 
         (false, None) => quote! {#[sea_orm(#(#parts), *)]},
     }
 }
-
-fn composite_pk_target_error(schema: &AnalyzedSchema) -> Option<TokenStream> {
-    for entity in &schema.entities {
-        for field in &entity.fields {
-            let FieldType::BelongsTo { target, .. } = &field.ty else {
-                continue;
-            };
-
-            let Some(target_entity) = schema.entities.iter().find(|e| &e.name == target) else {
-                continue;
-            };
-            let Some(pk_cols) = &target_entity.attrs.primary_key else {
-                continue;
-            };
-
-            if pk_cols.len() > 1 {
-                let pk_list = pk_cols
-                    .iter()
-                    .map(|col| format!("`{}`", col))
-                    .collect::<Vec<_>>()
-                    .join(", ");
-                let message = format!(
-                    "schema relationship `{}.{}` targets `{}` with composite primary key ({}); belongs_to relationships currently require a target with a single primary key column",
-                    entity.name, field.name, target_entity.name, pk_list
-                );
-
-                return Some(quote! {
-                    compile_error!(#message);
-                });
-            }
-        }
-    }
-
-    None
-}
-
 fn generate_custom_pk_fields(
     entity: &AnalyzedEntity,
     pk_cols: &[String],
@@ -214,22 +174,6 @@ fn generate_custom_pk_fields(
 }
 
 fn resolve_target_pk_type(target: &Ident, schema: &AnalyzedSchema) -> TokenStream {
-    resolve_target_pk_type_inner(target, schema, &mut Vec::new())
-}
-
-fn resolve_target_pk_type_inner(
-    target: &Ident,
-    schema: &AnalyzedSchema,
-    visiting: &mut Vec<String>,
-) -> TokenStream {
-    let target_name = target.to_string();
-    // Guard against self-referential or mutually-referential relationship PKs,
-    // which would otherwise recurse forever during macro expansion. Falling back
-    // to i32 keeps expansion terminating; such a cyclic PK has no scalar anchor.
-    if visiting.contains(&target_name) {
-        return quote! { i32 };
-    }
-
     let resolved = schema
         .entities
         .iter()
@@ -247,10 +191,7 @@ fn resolve_target_pk_type_inner(
                             target: inner_target,
                             ..
                         } => {
-                            visiting.push(target_name.clone());
-                            let ty = resolve_target_pk_type_inner(inner_target, schema, visiting);
-                            visiting.pop();
-                            return Some(ty);
+                            return Some(resolve_target_pk_type(inner_target, schema));
                         }
                         FieldType::HasMany { .. } => {}
                     }
@@ -856,35 +797,6 @@ mod tests {
     }
 
     #[test]
-    fn test_generate_belongs_to_composite_pk_target_is_compile_error() {
-        let input = quote! {
-            #[primary_key(left_id, right_id)]
-            #[timestamps(none)]
-            Pair {
-                left_id: i32,
-                right_id: i32,
-            }
-
-            PairRef {
-                pair: Pair,
-            }
-        };
-
-        let parsed = parse_schema(input).unwrap();
-        let analyzed = analyze_schema(parsed).unwrap();
-        let generated = generate_schema(analyzed);
-        let output = generated.to_string();
-
-        assert!(output.contains("compile_error !"));
-        assert!(output.contains("PairRef.pair"));
-        assert!(output.contains("Pair"));
-        assert!(output.contains("left_id"));
-        assert!(output.contains("right_id"));
-        assert!(!output.contains("pub pair_id : i32"));
-        assert!(!output.contains("Column::Id"));
-    }
-
-    #[test]
     fn test_generate_belongs_to_pk_resolves_transitively() {
         // A relation to an entity whose own PK is a belongs_to must resolve the
         // FK scalar type transitively, not fall back to i32.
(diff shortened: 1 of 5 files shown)


ShiraiEd (dismissed):
Nice fix! One small ask before merging:

  Better error message for Option<Entity> in PK

  Append a hint:

  "primary_key column '{}' cannot be an optional relationship; a primary key cannot be nullable.
   Remove the `Option<...>` wrapper."

  Matches the convention of the created_at/updated_at collision messages in the same file.

  Otherwise LGTM!

rapina-macros/src/schema/generate.rs
```diff
@@ -118,27 +118,104 @@ fn generate_entity_module(entity: &AnalyzedEntity, schema: &AnalyzedSchema) -> T
...
+            } else {
+                // Default PK is i32
+                return Some(quote! { i32 });
+            }
+            None
+        });
+
+    // Fallback to i32 if target not found or complex PK.
+    resolved.unwrap_or_else(|| quote! { i32 })
+}
```
arferreira, line 208: If the target has a composite PK there's no single type for this column, and we emit `i32` anyway? So schema compiles then fails at the database, I am a bit concerned but also that cycle guard above as well. We already reject `has_many` and optional relationships with a proper error, so cycles and composite targets should get the same treatment there... What do you think?

rapina-macros/src/schema/generate.rs
```diff
@@ -118,27 +118,104 @@ fn generate_entity_module(entity: &AnalyzedEntity, schema: &AnalyzedSchema) -> T
...
+        .find(|e| &e.name == target)
+        .and_then(|e| {
+            let pk_cols = e.attrs.primary_key.as_ref()?;
+            if pk_cols.len() == 1 {
+                Some(to_pascal_case(&pk_cols[0]))
+            } else {
+                None
+            }
+        })
+        .unwrap_or_else(|| "Id".to_string())
```
arferreira, line 226: Here's the same issue as the type fallback, a composite PK target gets `Column::Id`, which may not exist on the target

arferreira (changes requested):
Hey solid contribution, thanks! I requested some changes in the way I think it's important before we close it. Also add a line to the schema docs about the verbatim column name rule for PK fields, since that's user facing behavior now.

ShiraiEd:
Hello @mvanhorn,
how is the progress on this one?

ShiraiEd:
Hi @mvanhorn, 
It has been a week, how is the current situation?

mvanhorn:
Good catch. Since `belongs_to` maps a single field to a single FK column, there's no correct single-column type for a composite-PK target — so rather than emit `i32`/`Column::Id` that compiles but breaks at the database, it now raises a `compile_error!` naming the relationship and the target's PK columns. Added a test. Full composite-PK-target support would be a larger follow-up.

rapina-macros/src/schema/generate.rs
```diff
@@ -118,27 +122,139 @@ fn generate_entity_module(entity: &AnalyzedEntity, schema: &AnalyzedSchema) -> T
...
+    schema: &AnalyzedSchema,
+    visiting: &mut Vec<String>,
+) -> TokenStream {
+    let target_name = target.to_string();
+    // Guard against self-referential or mutually-referential relationship PKs,
+    // which would otherwise recurse forever during macro expansion. Falling back
+    // to i32 keeps expansion terminating; such a cyclic PK has no scalar anchor.
+    if visiting.contains(&target_name) {
+        return quote! { i32 };
+    }
```
arferreira: Coming back to the cycle half of my earlier review. Before this PR, analyze rejected relationship PKs with `must be a scalar type, not a relationship`, so `#[primary_key(parent)] parent: Category` failed with a clear message. Now it reaches here and expands to `pub parent: i32` with `to = Column::Parent`, a primary key that's a foreign key to itself with a type we chose. It compiles with no diagnostic (I checked the expansion locally). Isn't that the same compiles-then-breaks-at-the-database case you're erroring on for composite targets? Is there a reason the cycle terminates with `i32` here instead of raising the same `compile_error!` naming it?

rapina-macros/src/schema/generate.rs
```diff
@@ -118,27 +122,139 @@ fn generate_entity_module(entity: &AnalyzedEntity, schema: &AnalyzedSchema) -> T
...
+                    .collect::<Vec<_>>()
+                    .join(", ");
+                let message = format!(
+                    "schema relationship `{}.{}` targets `{}` with composite primary key ({}); belongs_to relationships currently require a target with a single primary key column",
+                    entity.name, field.name, target_entity.name, pk_list
+                );
+
+                return Some(quote! {
+                    compile_error!(#message);
+                });
```
arferreira: Should this point at the field instead of the whole `schema! {}` block? The has_many and optional rejections use `syn::Error::new(field.name.span(), ...)` and land on the offending field. `field.name` is in scope here too, so maybe:
```suggestion
                return Some(
                    syn::Error::new_spanned(&field.name, message).to_compile_error(),
                );
```
Would that work? Or would you rather move the whole check to the end of `analyze_schema` next to the other PK validations, so it returns a normal `syn::Error` and generate assumes the invariant?

arferreira (changes requested):
Hey thanks for the updates. ﻿Left two inline questions, both on edges of what this PR already touched rather than new scope. Could we also get a docs line on the verbatim PK column-name rule, now that it's user facing? Composite-PK-target support in full feels like a fair follow-up, agree?

mvanhorn:
Went with your second option in dbeadf0 - both checks now live at the end of analyze_schema next to the other PK validations, returning field-spanned syn::Errors, and generate assumes the invariant. The self-referential cycle case you found now errors with 'Category.parent forms a primary key cycle' instead of expanding to i32 (the i32 fallback guard in resolve_target_pk_type is gone since analyze rejects every cycle that recursion could follow). Added tests asserting both the message and that the span lands on the offending field.

arferreira:
Hey heads up before you rebase, I just merged #750, which rewrites `generate_custom_pk_fields` to stop hand rolling the field and reuse the attribute builder from `generate_model_field`. It'll conflict here. The shape you want on top is the same one you have now, just prepending your parts instead of emitting the attribute directly. Ping me if the rebase gets ugly and I'll take a pass at it.

mvanhorn:
Rebased onto main on top of #750, and the docs line is in.

The conflict resolved the way you predicted. #750's `sea_orm_parts` / `build_field_attr` now do the attribute work, and the relationship-PK branch just feeds them: the match yields `(rust_type, column_type)`, where a scalar contributes its own `column_type_attr()` and a `belongs_to` PK resolves to the target's PK type with no column type of its own. Then `primary_key`, `auto_increment = false`, and `sea_orm_parts(field)` are prepended and handed to `build_field_attr`. So column attributes are still respected on composite PK fields, including on relationship PKs, which they were not before.

Worth noting the two test suites now cover each other: your `test_composite_pk_respects_column_rename` and my `test_join_table_entity_typed_primary_key_compiles` both pass on the merged shape, which is the evidence that #750's behavior survived the rebase. Full run is 7/7 on `rapina --features database` and 134/134 on `rapina-macros`.

Docs: added the verbatim column-name rule next to the join-table example in `core-concepts/database.md`, since that is the surprising part - `user_id: User` stays `user_id` instead of becoming `user_id_id`.

On composite-PK targets as a follow-up: agreed. This PR keeps rejecting them, now from analyze with a field-spanned error rather than reaching generate.

arferreira (approved):
Thanks for the work! LGTM.

PR #621, opened 2026-06-08
Ignore-proc-macro-error2

ShiraiEd:
## Summary

Added proc-macro-error2 to `[advisories] ignore` in deny.toml so the CI check passes

## Related Issues

Closes #620

## Checklist

- [x] `cargo fmt` passes
- [x] `cargo clippy` has no warnings
- [x] Tests pass
- [ ] Documentation updated (if needed)

PR #623, opened 2026-06-10
Ignore proc-macro-error2 advisory in cargo-audit

arferreira:
Cargo Audit job uses `cargo-audit`, which keeps its own ignore list inline in the workflow rather than reading `deny.toml`. #621 pinned `RUSTSEC-2026-0173` for `cargo-deny` but not here, so the unmaintained `proc-macro-error2` advisory still fails Security Audit on push to main. Adds the same advisory to the audit-check ignore list to bring both gates in sync.


ShiraiEd (approved):
LGTM

PR #624, opened 2026-06-10
Ignore proc-macro-error2 advisory in ci cargo-audit

arferreira:
Security Audit job in ci.yml runs cargo-audit with its own inline ignore list, separate from both deny.toml and the audit.yml workflow. #623 synced audit.yml but not this one, so the unmaintained proc-macro-error2 advisory still failed on push to main. Adds `RUSTSEC-2026-0173` here too so all three advisory gates match.


ShiraiEd (approved):
LGTM

PR #631, opened 2026-06-12
[Feature]: Support multiple RDBMS for background jobs persistence  

### LINKED ISSUE #430: [Feature]: Support multiple RDBMS for background jobs persistence
### Problem Statement

The first implementation of the background jobs is PostgreSQL-only.
The reason is:
> the claiming query relies on `FOR UPDATE SKIP LOCKED`, `gen_random_uuid()`, `now()`, and a partial index, which is the thing that makes this work under concurrent workers without contention. MySQL has limited support for it, SQlite doesn't have at all.

Background jobs are a great feature in Rapina but forcing it to be PostgreSQL-only limits the adoption for interested users

### Proposed Solution

Reduce the hard PostgreSQL dependency in the background job queries by supporting multiple RDBMS (using SeaORM)

### Alternatives Considered

_No response_

### Contribution

Ismaellima4:
## Summary

What does this PR do?

Extract backend-specific SQL into dedicated structs (Postgres, Mysql, Sqlite)
and dispatch via match on DbBackend instead of hardcoding PostgreSQL.

Migration:
- Remove gen_random_uuid(), DEFAULT now(), DEFAULT '{}' — application supplies all values
- Conditional index: partial for Postgres, regular for MySQL, skip for SQLite

Model:
- Manual FromQueryResult — try native Uuid first, then String + parse (SQLite stores UUIDs as TEXT)

Backend extraction (backend.rs):
- build_insert_stmt, build_retry_stmt, build_fail_stmt, build_success_stmt, claim_batch per backend
- mod.rs / retry.rs / worker.rs dispatch via match on DbBackend

## Related Issues

Closes #430 

## Checklist

- [x] `cargo fmt` passes
- [x] Tests pass


rapina/src/jobs/backend.rs
```diff
@@ -0,0 +1,579 @@
...
+                   {sa} = NULL
+               WHERE {id} = ?"#
+        );
+
+        Statement::from_sql_and_values(
+            DbBackend::MySql,
+            &sql,
+            [
+                Value::String(Some(Box::new(error.to_owned()))),
+                Value::Int(Some(delay_secs as i32)),
```
arferreira: Okay, this drops the fractional part of the backoff before MySQL ever sees it, `delay_secs` is an f64, and a float to int as cast in [Rust](https://doc.rust-lang.org/reference/expressions/operator-expr.html#r-expr.as.numeric) rounds toward zero, so `1.7f64 as i32` is 1. 

Also, that fraction is where the jitter lives, `exponential_delay` returns base `* 4^(n-2) + jitter`, and `jitter` is in `[0, base)`. With the default `base_delay` of 1s the jitter is entirely sub-second, so the cast discards all of it and every job retrying at the same attempt lands on the same `run_at`. The jitter that spreads out retries is silently disabled on MySQL.

Postgres and SQLite keep the fraction because they pass it through `make_interval(secs => $2)` and `'+' || ? || ' seconds'`. I ran the SQLite path locally with a 1.5s delay and the retried job correctly stayed unclamable until it passed. 

If you want MySQL to keep sub-second precision, `INTERVAL ? SECOND` is the whole seconds unit, fractional values need `SECOND_MICROSECOND` or `MICROSECOND`, take a look [here](https://dev.mysql.com/doc/refman/8.0/en/expressions.html) at Table 11.2.

rapina/src/jobs/backend.rs
```diff
@@ -0,0 +1,579 @@
...
+
+        let sql = format!(
+            r#"UPDATE {t}
+               SET {st}   = 'running',
+                   {sa}   = CURRENT_TIMESTAMP,
+                   {lu} = CURRENT_TIMESTAMP + INTERVAL ? SECOND
+               WHERE {id} IN ({id_placeholders})"#
+        );
+
+        let mut values: Vec<Value> = vec![Value::Int(Some(config.job_timeout.as_secs() as i32))];
```
arferreira: I think it's the same case as that cast for `delay_secs` is harmless since job_timeout is whole seconds in practice.

rapina/src/jobs/create_rapina_jobs.rs
```diff
@@ -145,8 +179,26 @@ impl MigrationTrait for Migration {
...
+///     "UPDATE {} SET {} = 'running' WHERE {} = ?",
+///     RapinaJobs::table_name(),
+///     RapinaJobs::Status.to_string(),
+///     RapinaJobs::Id.to_string(),
+/// );
+/// ```
+///
+/// When the iden enum changes, the compiler catches the mismatch because
+/// `.to_string()` is generated by `DeriveIden`.
 #[derive(DeriveIden)]
```
arferreira: Well, this doc is incorrect, the accessors below don't use `.to_string()`. Did you mean to centralize the literal or to get compile time checking? Right now it's neither.

rapina/src/jobs/mod.rs
```diff
@@ -337,58 +315,9 @@ mod tests {
...
-            queue: "default",
-            max_retries: 3,
-        };
-        let stmt = build_insert_stmt(req, None);
-        let trace_val = &stmt.values.as_ref().unwrap().0[4];
-        assert_eq!(*trace_val, sea_orm::Value::String(None));
+    fn insert_job_generates_new_uuid() {
+        // Uuid::new_v4() should produce a non-nil UUID.
+        let id = Uuid::new_v4();
+        assert_ne!(id, Uuid::nil());
```
arferreira: This asserts the `uuid` crate works, not our code, so I'd drop it. I see a gap where you should be exercising the three dialects your PR is adding.

rapina/src/jobs/backend.rs
```diff
@@ -0,0 +1,579 @@
...
+        let r = RapinaJobs::run_at();
+
+        let sql = format!(
+            r#"SELECT * FROM {t}
+               WHERE  {st}  = 'pending'
+                 AND  {q}   IN ({placeholders})
+                 AND  {r} <= CURRENT_TIMESTAMP
+               ORDER  BY {r} ASC
+               LIMIT  ?
+               FOR UPDATE SKIP LOCKED"#
```
arferreira: I think we are pulling columns nobody uses, `SELECT {id}` is enough.

rapina/src/jobs/create_rapina_jobs.rs
```diff
@@ -58,12 +56,7 @@ impl MigrationTrait for Migration {
                             .string_len(255)
                             .not_null(),
                     )
-                    .col(
-                        ColumnDef::new(RapinaJobs::Payload)
-                            .json_binary()
-                            .not_null()
-                            .default("{}"),
-                    )
+                    .col(ColumnDef::new(RapinaJobs::Payload).json().not_null())
```
arferreira, line 61: This movement makes it portable, but Postgres loses `jsonb` and its indexing. I'd take that as the price of portability, just worth a line in the migration doc so it reads as a recorded choice instead of a silent regression.

arferreira (changes requested):
Good work so far, thanks! I pulled your branch and ran the SQlite e2e locally, looking great. I left some comments important to address before we merge, it might be better if you run against a real MySQL 8.

arferreira (approved):
LGTM

PR #632, opened 2026-06-13
Add --diff mode to compare entities against the live database

### LINKED ISSUE #245: Add --diff mode to compare entities against live database

arferreira:
Closes #245.

Adds `--diff` to `rapina import database`. It rebuilds the table each `schema!` block in `src/entity.rs` describes and compares it against the live database. It reports columns that exist on only one side, type and nullability mismatches, missing tables, and primary key differences. The exit code is 0 when code and the database agree, 2 on drift, and 1 on an error, so it works in CI. Two differences that round-trip at runtime are reported as notes rather than drift: a `String` field against a `text` column, and a `bool` against a MySQL `tinyint`. A column whose type has no `schema!` equivalent is skipped on both sides instead of being flagged as false drift. `--tables` scopes the comparison, and since `--diff` writes nothing, combining it with `--force` is rejected.

The URL-scheme introspection is now shared between the normal import and the diff path, and the `schema!` parser is shared with `seed`, which gained attribute parsing for `#[table_name]`, `#[timestamps]`, `#[primary_key]`, and `#[column]`.


rapina-cli/src/commands/schema_source.rs
ShiraiEd, line 1: This checks one entity for `schema!`, but if multiple entities are declared in just one block, it catches just the first one, e.g.
```rust
schema! {
    User { name: String, }
    Post { title: String, }
} //just returns User
```

this happens because `parse_schema_content` just checks for `schema!` pattern and the inner loop with `parse_one_entity` returns on the first '}' without checking for more entities

This already existed in `seed.rs`, now is affecting `--dif` as well

There is also no tests for multiple entities in a single block, the existing test just covers two separated schema blocks

any thoughts on this? A follow up maybe?

rapina-cli/src/commands/import/diff.rs
```diff
@@ -0,0 +1,906 @@
...
+/// Uuid when the entity declares `id: Uuid`, otherwise the default i32.
+fn entity_pk_type(entity: &ParsedEntity) -> NormalizedType {
+    entity
+        .fields
+        .iter()
+        .find(|f| f.name == "id")
+        .and_then(|f| schema_type_to_normalized(strip_option(&f.type_str).0))
+        .filter(|t| *t == NormalizedType::Uuid)
+        .unwrap_or(NormalizedType::I32)
+}
```
ShiraiEd, line 143: This assumes the pk name is always `id`,and its types are always `i32` or `uuid`,
but with the `#[primary_key(slug)]` attribute the pk's name can be anything, and its type any one of the scalars not just `i32` or `uuid.`

this falls back to `i32` and causes type mismatches

adjusting to find and use the entity's id name and correct type would be the call, also a new test for a non `i3,`, `uuid` pk can be added

ShiraiEd (commented):
Hi!

I found two things that can result in incorrect diffs reports if entity.rs doesn't follow the template generated by `rapina add resource`

The first one can happen if the entities are all declared in the same `schema!` block,
and the second can happen if the name of the pk isn't `id` and its type isn't `i32` or `uuid`

rapina-cli/src/commands/schema_source.rs
arferreira: Nop, totally true! Good, catch.

rapina-cli/src/commands/schema_source.rs
```diff
@@ -34,7 +38,7 @@ pub(crate) fn parse_schema_content(content: &str) -> Result<Vec<ParsedEntity>, S
     Ok(entities)
 }
 
-fn parse_one_entity(lines: &mut std::str::Lines) -> Result<ParsedEntity, String> {
+fn parse_one_entity(lines: &mut std::str::Lines) -> Result<Option<ParsedEntity>, String> {
```
ShiraiEd: Nit
This could be renamed to something like `try_parse_entity` or `parse_next_entity`
since the return is now an `Option` instead of a single entity

ShiraiEd (dismissed):
Nice, just one nit

the rest LGTM

rapina-cli/src/commands/schema_source.rs
arferreira: good catch, I think `parse_next_entity` reads more honestly.

ShiraiEd (approved):
LGTM

PR #637, opened 2026-06-16
feat(GraphQL): GraphQLRequest extractor and GraphQLResponse responder

### LINKED ISSUE #597: [Feature]: GraphQLRequest extractor + GraphQLResponse responder
### Problem Statement

This is the first building block for Graphql support,
A way for handlers to receive/return Graphql requests/responses without having to deal with the boilerplate

the next issues depend on this one to be implemented, like the App method convention later to be done

update docs and changelog alongside

### Proposed Solution

Following [Luan's plan](https://github.com/luanperosa/rapina-poc-graphql/blob/master/PLAN.md)

Add a new `rapina::graphql` module gated behind `#[cfg(feature = "graphql")]`. PR 1 ships as a single file `rapina/src/graphql/mod.rs`, later PRs will split it into `request.rs`, `response.rs`, `context.rs`, etc. as the surface grows.

  ### Cargo feature

  ```toml
  # rapina/Cargo.toml
  [features]
  graphql = ["dep:async-graphql"]

  [dependencies]
  async-graphql = { version = "7", optional = true }
  ```

  ### Module registration

  ```rust
  // rapina/src/lib.rs
  #[cfg(feature = "graphql")]
  pub mod graphql;
  ```

  ### `rapina/src/graphql/mod.rs`

  async-graphql's `Request` derives `Deserialize`, so the POST path is a direct `serde_json::from_slice`. For GET, we parse a helper struct from the URL query string, then convert it into `async_graphql::Request`, the `variables` query param is URL-encoded JSON, so it requires an explicit `serde_json::from_str` step into `async_graphql::Variables`.

  ```rust
  /// Intermediate struct for GET query-param extraction.
  #[derive(Deserialize)]
  struct GraphQLParams {
      query: String,
      #[serde(default)]
      variables: Option<String>,
      #[serde(default, rename = "operationName")]
      operation_name: Option<String>,
  }

  pub struct GraphQLRequest(pub async_graphql::Request);

  impl FromRequest for GraphQLRequest {
      async fn from_request(req: Request<Incoming>, ...) -> Result<Self, Error> {
          // Method dispatch:
          //   GET  → parse query string into GraphQLParams, convert to Request
          //          (variables param goes through explicit serde_json::from_str)
          //   POST → serde_json::from_slice of the body into async_graphql::Request
          // Malformed input → Error::bad_request (400 with Rapina's error envelope).
      }
  }

  pub struct GraphQLResponse(pub async_graphql::Response);

  impl IntoResponse for GraphQLResponse {
      fn into_response(self) -> http::Response<BoxBody> {
          // ALWAYS 200 for application/json per the GraphQL-over-HTTP spec.
          // Field-level resolver errors live in the `errors` array of the body,
          // NOT in the HTTP status code.
          let body = serde_json::to_vec(&self.0).unwrap_or_default();
          http::Response::builder()
              .status(StatusCode::OK)
              .header("content-type", "application/json")
              .body(full(body)).unwrap()
      }
  }
  ```

  ### Why a dedicated module instead of adding to `extract/`

  The umbrella feature (#202) eventually adds 6+ items spread across extractor / responder / context / error / builder method / handler concerns. Keeping them in one `rapina::graphql::*` module means:

  - One feature flag gates the whole surface (`#[cfg(feature = "graphql")]` in `lib.rs`)
  - One `use rapina::graphql::*;` import gives users everything
  - Matches existing rapina patterns (`src/jwt/`, `src/relay/`, `src/cache/`)

### Alternatives Considered

Two separate extractors (`GraphQLPostRequest + GraphQLGetRequest`): rejected because it forces handler authors to duplicate the route registration. Single extractor with method dispatch is cleaner and matches `async-graphql's` own integration adapters.

### Acceptance criteria
  - [x] Single unified `GraphQLRequest` extractor (dispatches POST/GET internally)
  - [x] HTTP status always 200 for `application/json` (no 4xx for field-level errors)
  - [x] GET `variables` param explicitly parsed via `serde_json::from_str` → `async_graphql::Variables`
  - [x] Cargo feature: `graphql = ["dep:async-graphql"]`
  - [x] Unit tests: POST JSON, GET query-string, malformed body → 400, field-level error → still 200
  - [x] Docs page + CHANGELOG entry
  - [x] Cargo fmt + cargo clippy

### Contribution

ShiraiEd:
## Summary

- New `graphql` feature flag
- `GraphQLResquest` extractor and `GraphQLResponse` responder structured wrapping `async-graphq`'s `Request` and `Response`, implementing `FromRequest` and `IntoResponse`
- Those 2 implement `Deref` and `into_inner( )` like the other extractors
- Re-exports in `rapina::prelude`
- 4 Unit tests covering POST, GET, malformed body and field level errors returning status 200
- Docs page + CHANGELOG entry

## Related Issues

Closes #597 

## Checklist

- [x] `cargo fmt` passes
- [x] Tests pass

### DIFF
--- a/docs/content/docs/core-concepts/graphql.md
+++ b/docs/content/docs/core-concepts/graphql.md
@@ -45,7 +45,7 @@ Field-level resolver errors live in the response body's `errors` array, never
 in the HTTP status.
 
 The 400 status is reserved for *transport-level* failures: malformed JSON in
-the POST body, invalid query string in GET, or unsupported HTTP method.
+the POST body or an invalid query string in GET. Unsupported HTTP methods return 405.
 
  ### Limitations
 
--- a/docs/content/docs/core-concepts/graphql.md
+++ b/docs/content/docs/core-concepts/graphql.md
@@ -1,0 +1,52 @@
++++
+title = "GraphQL"
+description = "GraphQL request/response extractor and responder behind the graphql feature flag"
+weight = 13
+date = 2026-06-16
++++
+
+Rapina ships a thin integration with [async-graphql](https://docs.rs/async-graphql)
+so handlers can receive and return GraphQL payloads without boilerplate. Gated
+behind the `graphql` feature flag.
+
+## Setup
+
+Add the feature to your `Cargo.toml`:
+
+```toml
+[dependencies]
+rapina = { version = "0.12.0", features = ["graphql"] }
+async-graphql = "7.0"
+```
+
+## The extractor and responder
+
+Two newtypes wrap `async_graphql::Request` and `async_graphql::Response`:
+
+```rust
+use rapina::prelude::*;
+use async_graphql::{Schema, EmptyMutation, EmptySubscription};
+
+#[post("/graphql")]
+async fn graphql_handler(req: GraphQLRequest) -> GraphQLResponse {
+    let schema: Schema<Query, EmptyMutation, EmptySubscription> = /* ... */;
+    GraphQLResponse(schema.execute(req.0).await)
+}
+```
+
+`GraphQLRequest` accepts both POST (JSON body) and GET (query string with
+`query`, `variables`, `operationName`). Malformed input returns 400.
+
+## HTTP semantics
+
+Per the [GraphQL-over-HTTP spec](https://graphql.github.io/graphql-over-http/),
+`GraphQLResponse` always returns HTTP 200 with `Content-Type: application/json`.
+Field-level resolver errors live in the response body's `errors` array, never
+in the HTTP status.
+
+The 400 status is reserved for *transport-level* failures: malformed JSON in
+the POST body, invalid query string in GET, or unsupported HTTP method.
+
+ ### Limitations
+
+This pr just delivers the extractors, Schema management and the `with_graphql` builder method are still to be done!
--- a/docs/content/docs/core-concepts/graphql.md
+++ b/docs/content/docs/core-concepts/graphql.md
@@ -15,7 +15,7 @@ Add the feature to your `Cargo.toml`:
 
 ```toml
 [dependencies]
-rapina = { version = "0.12.0", features = ["graphql"] }
+rapina = { version = "0.13.0", features = ["graphql"] }
 async-graphql = "7.0"
 ```
 
--- a/docs/templates/changelog.html
+++ b/docs/templates/changelog.html
@@ -11,21 +11,6 @@
 
     <div class="changelog-timeline">
 
-
-        <div class="timeline-item">
-            <div class="timeline-marker"></div>
-            <div class="timeline-content">
-                <div class="timeline-date">Unreleased</div>
-                <h2>v0.13.0</h2>
-                <span class="timeline-tag tag-feature">Feature</span>
-                <ul>
-                    <li><code>GraphQLRequest</code> extractor and <code>GraphQLResponse</code> responder for <a href="https://docs.rs/async-graphql">async-graphql</a> integration behind <code>graphql</code> feature flag</li>
-                </ul>
-                <p class="timeline-contributors">Contributors: <a href="https://github.com/ShiraiEd">@ShiraiEd</a></p>
-            </div>
-        </div>
-
-
         <div class="timeline-item">
             <div class="timeline-marker"></div>
             <div class="timeline-content">
--- a/docs/templates/changelog.html
+++ b/docs/templates/changelog.html
@@ -11,6 +11,21 @@
 
     <div class="changelog-timeline">
 
+
+        <div class="timeline-item">
+            <div class="timeline-marker"></div>
+            <div class="timeline-content">
+                <div class="timeline-date">Unreleased</div>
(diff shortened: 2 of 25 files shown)


rapina/src/graphql/mod.rs
```diff
@@ -0,0 +1,261 @@
...
+                if let Some(name) = params.operation_name {
+                    inner = inner.operation_name(name);
+                }
+
+                Ok(GraphQLRequest(inner))
+            }
+
+            _ => Err(crate::error::Error::bad_request(
+                "Method not allowed in GraphQL",
+            )),
```
arferreira: I think we can change this, the status is `400` but the message says something different, and a reject method is `405`. [If an unsupported HTTP method is used, status code 405 is RECOMMENDED.](https://graphql.github.io/graphql-over-http/draft/#sec-application-graphql-response-json)

rapina/src/graphql/mod.rs
```diff
@@ -0,0 +1,261 @@
...
+                        })?;
+                    inner = inner.variables(Variables::from_json(value));
+                }
+
+                if let Some(name) = params.operation_name {
+                    inner = inner.operation_name(name);
+                }
+
+                Ok(GraphQLRequest(inner))
+            }
```
arferreira, line 176: This `reimplements async_graphql::http::parse_query_string`. I'm not saying swap to it blindly, because yours is actually more correct on one axis: you read `operationName` per the GraphQL-over-HTTP GET spec, while theirs reads `operation_name`. But theirs also decodes `extensions`, and yours drops it, which means APQ over GET silently won't work. Is that an intentional scope cut? A one line comment saying "extensions intentionally unsupported for now" would make the divergence a decision instead of an accident.

rapina/src/graphql/mod.rs
```diff
@@ -0,0 +1,261 @@
...
+        let req = TestRequest::get("/graphql?query=%7B+hello+%7D&operationName=Hello")
+            .into_incoming_request()
+            .await;
+
+        let result = GraphQLRequest::from_request(req, &empty_params(), &empty_state()).await;
+        assert!(result.is_ok());
+        let request = result.unwrap().0;
+        assert_eq!(request.query, "{ hello }");
+    }
+
```
arferreira, line 229: Could we add a case that sends `variables` (URL encoded JSON) and `operationName` and asserts both land on the inner Request, plus one for the method arm asserting 405? I guess that's the part most likely to break under a refactor.

docs/templates/changelog.html
```diff
@@ -11,6 +11,21 @@ <h1>Changelog</h1>
 
     <div class="changelog-timeline">
 
+
+        <div class="timeline-item">
+            <div class="timeline-marker"></div>
+            <div class="timeline-content">
+                <div class="timeline-date">Unreleased</div>
+                <h2>v0.13.0</h2>
```
arferreira: Here's the thing, this will not be available in that version, I pointed to `0.14.0` with an optimist sense, it might be an option just (unknown), what do you think?

arferreira (changes requested):
This looks nice, good work so far! I left few questions and one important detail about that 405 reject method arm, happy to to approve once we address that.

rapina/src/graphql/mod.rs
ShiraiEd: Will add that comment and open a follow up for this later

docs/templates/changelog.html
ShiraiEd: so, for the way I am thinking, our new release is scheduled for tomorrow, and if this gets merged and released with 0.13 it will be just the tip of the feature, with the rest missing

I think I should 
1. remove the changelog updates on the .md and .html for now,
2. merge #653 and release the new version
3. rebase this branch on top of the main afterwards, since 653 will already be merged I won't need to deal with conflics later
4. and then try to release the complete implementation in 0.14 with the changelog.md to that version, all updated, and if I can't at least most of it will be already implemented

arferreira: Well, this is planned to arrive with `0.14.0`, makes sense removing from changelog, or just moving to "In Progress".

docs/content/docs/core-concepts/graphql.md
```diff
@@ -0,0 +1,52 @@
...
+
+## HTTP semantics
+
+Per the [GraphQL-over-HTTP spec](https://graphql.github.io/graphql-over-http/),
+`GraphQLResponse` always returns HTTP 200 with `Content-Type: application/json`.
+Field-level resolver errors live in the response body's `errors` array, never
+in the HTTP status.
+
+The 400 status is reserved for *transport-level* failures: malformed JSON in
+the POST body, invalid query string in GET, or unsupported HTTP method.
```
arferreira: Sorry, you forgot updating here.

arferreira (changes requested):
Almost there, just the fix in the docs.

arferreira:
@ShiraiEd you have to update Rapina's version

arferreira (approved):
Lit! LGTM

PR #647, opened 2026-06-17
Add flexible, framework-invoked authorization logic

### LINKED ISSUE #116: [Feature]: Add pluggable authorization for endpoints
### Problem Statement

Handling authentication is one crucial part for a web framework.
An easy-to-implement mechanism to add authorization for endpoints is an equivalently important feature for many developers. 

Rapina should provide a built-in way to handle authorization for endpoints in a extendable/pluggable fashion.

### Proposed Solution

As a start, Rapina could consider role based access control for Json Web Tokens (JWT) as a demonstrator of this feature because authentication with JWTs is already supported. 

For example: a JWT can contain a certain claim that contains the roles for the associated user. As the JWT is already verified for its authencity (to be issued by a trusted Authorization server), Rapina can inspect the JWT and authorize based on the claim and its value.
In case the claim does not contain expected values, Rapina should intercept the request and return a `403 FORBIDDEN` response

### Alternatives Considered

_No response_

Security (guardrails, validation)

### Contribution

juv:
## Summary

Adds a new `#[authorize]` macro that adds flexible authorization for any endpoint handler.
See /examples/jwt-validation for a full example.

## Related Issues

Closes #116

## Checklist

- [ ] `cargo fmt` passes
- [ ] Tests pass


rapina/examples/jwt-validation/src/main.rs
```diff
@@ -1,15 +1,45 @@
...
 #[get("/email")]
-async fn get_email(token: JsonWebToken<GoogleClaims>) -> Json<String> {
+#[authorize(authz::authorize(JsonWebToken))]
+async fn get_email(token: JsonWebToken<GoogleClaims>, _unused: Headers) -> Result<Json<String>> {
     tracing::info!(sub = %token.sub, "authenticated request");
-    Json(token.claims.email)
+    Ok(Json(token.claims.email))
+}
+
+// Example handler that takes two parameters and is authorized by an authorization handler within another module
```
juv:
```suggestion
// Example handler that takes two parameters and is authorized by an authorization handler within the same module
```

rapina-macros/src/lib.rs
```diff
@@ -69,6 +74,187 @@ fn join_paths(prefix: &str, path: &str) -> String {
...
+        return syn::Error::new(
+            attr.span(),
+            "#[authorize] contradicts #[public]. A public handler must not include authorization.",
+        )
+        .to_compile_error()
+        .into();
+    }
+
+    item
+}
```
arferreira, line 349: Question on how the two attributes compose, since I want to be sure I'm reading the expansion right. Attribute macros expand top-down and each one consumes its own attribute, so with the documented order

```rust
#[get("/email")]
#[authorize(authz::authorize(JsonWebToken))]
```
so `#[get]` expands first and strips `#[authorize]` itself. Does this `authorize` proc macro ever actually run in that case? And if it doesn't, what's enforcing the `#[public]` conflict for the order users will actually write, can `#[get] #[public] #[authorize(...)]` compile into a route that's both public and authorized?

arferreira: Now the other way around:

```rust
#[authorize(authz::authorize(JsonWebToken))]
#[get("/email")]
```
Here `authorize` expands first and returns item without its own attribute, so by the time `#[get]` runs there's nothing left to find. Is the route still protected? I don't think it is, and that's the part that worries me. For an auth feature, getting the attribute order wrong shouldn't silently drop the check, it should fail to compile. Could we move all of this into `route_macro_core` (it already extracts the attr), put the `#[public]` conflict check there so it actually runs, and make a bare `#[authorize]` a hard error?

rapina-macros/src/lib.rs
```diff
@@ -69,6 +74,187 @@ fn join_paths(prefix: &str, path: &str) -> String {
...
+        Ok(Self { auth_fn, deps })
+    }
+}
+
+/// TODO add docs
+fn authorize_needs_parts(func: &ItemFn, contains_authorize: &Option<Attribute>) -> bool {
+    if let Some(attr) = &contains_authorize {
+        let auth_args = attr
+            .parse_args::<AuthorizeArgs>()
+            .expect("#[authorize] args must be parseable");
```
arferreira: Is there a reason we `expect()` here instead of surfacing a `syn::Error`? When the attribute args don't parse this panics, and a proc-macro panic renders as `proc macro panicked` with the message buried under it..

rapina-macros/src/lib.rs
```diff
@@ -69,6 +74,187 @@ fn join_paths(prefix: &str, path: &str) -> String {
...
+        let mut found = None;
+
+        for input in inputs {
+            let FnArg::Typed(PatType { pat, ty, .. }) = input else {
+                continue;
+            };
+
+            if normalize_type(ty) == dep_norm {
+                let ident = extract_ident(pat)?;
+                found = Some(quote!(&#ident));
```
arferreira: We hand the dependency to the auth fn two different ways depending on the branch: when a handler param matches, here and when we extract a fresh one, `#tmp` by value. The auth fn has a single signature for that parameter though, so which one is it meant to be?

rapina-macros/src/lib.rs
```diff
@@ -69,6 +74,187 @@ fn join_paths(prefix: &str, path: &str) -> String {
...
+        Pat::Ident(PatIdent { ident, .. }) => Ok(ident.clone()),
+        _ => Err(Error::new(
+            pat.span(),
+            "#[authorize] only supports simple identifier parameters, e.g. `state: State<AppConfig>`, `token: JsonWebToken<T>`",
+        )),
+    }
+}
+
+fn normalize_type(ty: &Type) -> String {
+    ty.to_token_stream().to_string().replace(' ', "")
```
arferreira, line 297: is this comparison really pulling its weight?

rapina-macros/src/lib.rs
```diff
@@ -336,11 +515,52 @@ fn route_macro_core(
...
     };
 
+    // compute whether the authorization handler requires extra request parts
+    let authorize_needs_parts = authorize_needs_parts(&func, &contains_authorize);
+    // build authorization prelude in case #[authorize] macro is present
+    let authorize_prelude: proc_macro2::TokenStream = if let Some(attr) = contains_authorize {
+        println!("authorize attr tokens: {}", attr.to_token_stream()); //TODO remove
+        let args = attr
+            .parse_args::<AuthorizeArgs>()
+            .expect("#[authorize] args must be parseable");
```
arferreira: Why this gets parsed twice? We could parse once and thread it through, minor though.

rapina-macros/src/lib.rs
```diff
@@ -69,6 +74,187 @@ fn join_paths(prefix: &str, path: &str) -> String {
...
+        }
+
+        let deps = if input.peek(syn::token::Paren) {
+            let content;
+            parenthesized!(content in input);
+            let parsed: Punctuated<Type, Token![,]> =
+                content.parse_terminated(Type::parse, Token![,])?;
+            parsed.into_iter().collect()
+        } else {
+            input.parse::<Token![,]>()?;
```
arferreira: Is the comma form deliberate, or a leftover?

rapina-macros/src/lib.rs
```diff
@@ -412,6 +636,10 @@ fn route_macro_core(
             quote! {
                 let (__rapina_parts, _) = __rapina_req.into_parts();
                 #(#header_extractions)*
+
+                //call authorization handler
```
arferreira: Just a nit, these comments skip the space. fmt won't catch it.

rapina/examples/jwt-validation/src/main.rs
```diff
@@ -60,6 +90,7 @@ async fn main() -> std::io::Result<()> {
     Rapina::new()
         .state(jwks_client)
         .state(jwks_validation)
+        .discover()
         .router(router)
```
arferreira:
```suggestion
```

arferreira (commented):
Nice work @juv! A few things need sorting before it can land.

juv:
Hi, thank you! Hope I will find the time on this Sunday to work on the changes :)

rapina-macros/src/lib.rs
juv: You are right, let's standardize on `&T`

rapina-macros/src/lib.rs
juv: Nope, just an oversight. Fixed

rapina-macros/src/lib.rs
juv: I agree, good points. I have added that suggestion.
But the same problem still persists with the `#[public]` macro:

Right now, the code allows that `#[public]` is sitting above and below the main route macro. Our documentation and code examples have multiple occasions where the `#[public]` macro is on top of the main macro: https://userapina.com/docs/core-concepts/authentication/ 

To make that happen, the code currently does a stretch:

1. Sets a `PublicMarker` if the macro is above the main route macro: https://github.com/rapina-rs/rapina/blob/main/rapina-macros/src/lib.rs#L165-L178
2. adds `is_public` to the `RouteDescriptor` if the macro is below main route macro: https://github.com/rapina-rs/rapina/blob/main/rapina-macros/src/lib.rs#L546 
3. Has logic in the `app.rs` to collect both: https://github.com/rapina-rs/rapina/blob/main/rapina/src/app.rs#L944-L964

To cover the cases where `#[public]` is sitting on top of the main route macro, we can no longer detect that `#[authorize]` and `#[public]` are present within `route_core_macro`. At the time the code reaches the `route_core_macro`, public was already handled and removed:

```
#[public]
#[get("/email")]
#[authorize(authz::authorize(JsonWebToken))]
```

Obviously this is a breaking change for existing code where #[public] sits on the top. But I think it's necessary in this case and it will also make the code easier and more consistent for the future.

rapina-macros/src/lib.rs
juv: Can you elaborate a bit? I added this because `syn::Type` rendered via `to_token_stream().to_string()` often inserts spaces, for example `JsonWebToken < GoogleClaims >`. The user likely writes `JsonWebToken<GoogleClaims>` instead, or some combination with spaces too. Removing spaces makes those spellings compare equal

juv: I've added a comment to the function that explains this too

rapina-macros/src/lib.rs
juv: Good catch, it was a leftover indeed

rapina-macros/src/lib.rs
```diff
@@ -69,6 +74,271 @@ fn join_paths(prefix: &str, path: &str) -> String {
...
+/// `syn::Type` stringification renders generics with spaces (e.g. `JsonWebToken < GoogleClaims >`),
+/// so this makes `JsonWebToken<GoogleClaims>` compare equal textually.
+///
+/// This is syntactic normalization only, not semantic type equality.
+fn normalize_type(ty: &Type) -> String {
+    ty.to_token_stream().to_string().replace(' ', "")
+}
+
+/// TODO add docs (in docs/ folder too)
+/// TODO add tests
```
juv: todo. 
Waiting here because the open PR comments might require substantial implementation changes still

rapina-macros/src/lib.rs
arferreira: Thanks, this is a big step up. Standalone `#[authorize]` as a hard error, plus the conflict check in `route_macro_core`, closes the ordering problem I was worried about. Putting it below the route macro I think is the right constraint.

On `#[public]` on top: I don't think we need the breaking change to catch this. The reason `route_macro_core` can't see it is the same thing that gives us another spot that can.  When `#[public]` sits on top it expands first and receives the whole item with `#[get]` and `#[authorize]` still attached below it, which is why the `PublicMarker` path exists in the first place. So the `#[public]` macro itself can see the `#[authorize]` underneath it and reject the pairing right there. You've already got `extract_authorize_attr` to reuse.

That gives us both sides: `route_macro_core` catches `#[public]` below, the `#[public]` macro catches `#[public]` on top, and standalone `#[authorize]` is already an error. Every ordering ends up a compile error, and `#[public]` on top keeps working for the normal case. No break, no runtime check. Does that cover what you were worried about? If there's an ordering where neither spot sees it, point me at it.

Separately, you're right that the two path `#[public]` is a stretch, and collapsing it to below only would be cleaner long term, especially as more attributes start composing with routes. I'd just rather not fold that into this PR. It's a breaking change to a documented pattern, so it deserves its own PR with a migration note as a follow-up, decided on its own terms rather than as a side effect of adding `#[authorize]`. 
Happy to help spec it out after this lands if you want to pick it up.

rapina-macros/src/lib.rs
arferreira: Nah that makes sense, and you're right! One thing worth adding to the rustdoc you added, what happens on a miss, I mean, if the spelling doesn't line up, I think it's good to say out loud.

arferreira (commented):
Hey this is looking really good! I left some comments, no rush, it's still a draft. Just flip it to ready when these land and I'll do a full pass.

rapina-macros/src/lib.rs
```diff
@@ -69,6 +74,271 @@ fn join_paths(prefix: &str, path: &str) -> String {
...
+    }
+}
+
+/// Normalizes a type into a whitespace-insensitive token string.
+///
+/// `syn::Type` stringification renders generics with spaces (e.g. `JsonWebToken < GoogleClaims >`),
+/// so this makes `JsonWebToken<GoogleClaims>` compare equal textually.
+///
+/// This is syntactic normalization only, not semantic type equality.
+fn normalize_type(ty: &Type) -> String {
```
juv, line 296: Type reuse relies on token-string equality, so aliases and differently qualified paths such as `JsonWebToken<Claims>` and `rapina::jwt::JsonWebToken<Claims>` miss and cause a second extraction

juv:
> Hey this is looking really good! I left some comments, no rush, it's still a draft. Just flip it to ready when these land and I'll do a full pass.

Hi, I think this is in an acceptable state for a review now :)

rapina-macros/src/lib.rs
```diff
@@ -2312,4 +2736,91 @@ mod tests {
...
+        assert_eq!(
+            args.auth_fn
+                .segments
+                .iter()
+                .map(|segment| segment.ident.to_string())
+                .collect::<Vec<_>>(),
+            ["authz", "authorize"]
+        );
+        assert!(args.deps.is_empty());
+    }
```
arferreira, line 2756: Could we get a unit test for the `#[public]` conflict? I guess one call to `route_macro_core` asserting the message. Same for that non identifier pattern error the docs promise.

rapina-macros/src/lib.rs
```diff
@@ -487,8 +879,16 @@ fn route_macro_core(
...
                 let (__rapina_parts, __rapina_body) = __rapina_req.into_parts();
+
+                // Extract authorization dependencies while request parts remain available
+                #authorization_extracts
+
                 #(#parts_extractions)*
                 #last_extraction
+
+                // Reused handler parameters are now bound, authorization handler can be called
+                #authorization_call
```
arferreira, line 890: Since body extractors can't be authorize deps anyway, could the call move up right after `#(#parts_extractions)*` when no dep reuses the body binding? Actually,  what does a denying policy plus malformed JSON return today? I didn't see a test for that one.

rapina/examples/jwt-validation/src/main.rs
```diff
@@ -1,15 +1,37 @@
...
+async fn pong() -> Result<Json<String>> {
+    tracing::info!("this is called within the handler body");
+    Ok(Json("success".to_string()))
+}
+
+// Example handler that takes two parameters and is authorized by an authorization handler within another module
 #[get("/email")]
-async fn get_email(token: JsonWebToken<GoogleClaims>) -> Json<String> {
+#[authorize(authz::authorize(JsonWebToken))]
+async fn get_email(token: JsonWebToken<GoogleClaims>, _unused: Headers) -> Result<Json<String>> {
```
arferreira, line 32: Is `_unused: Headers` doing anything here?

rapina/examples/jwt-validation/src/main.rs
```diff
@@ -1,15 +1,37 @@
...
+}
+
+// Example handler that takes no parameter and is authorized by an authorization handler within another module
+#[get("/example")]
+#[authorize(authorization_handler(JsonWebToken<GoogleClaims>))]
+async fn pong() -> Result<Json<String>> {
+    tracing::info!("this is called within the handler body");
+    Ok(Json("success".to_string()))
+}
+
```
arferreira, line 28: Is `pong` unused?

rapina-macros/src/lib.rs
```diff
@@ -69,6 +74,288 @@ fn join_paths(prefix: &str, path: &str) -> String {
...
+        .into()
+}
+
+/// Extract #[authorize] attribute from function attributes, removing it if found.
+fn extract_authorize_attr(attrs: &mut Vec<syn::Attribute>) -> Option<Attribute> {
+    attrs
+        .iter()
+        .position(|attr| attr.path().is_ident("authorize"))
+        .map(|idx| attrs.remove(idx))
+}
```
arferreira, line 357: What happens with two `#[authorize]` attrs? Looks like the second survives into the emitted fn and expands standalone, so the error says "must be used with a route macro" on a function that has one. Worth a "duplicate #[authorize]" error spanned to the second?

arferreira (changes requested):
Yo @juv this came out solid! Thanks, nits and questions batched.

rapina/examples/jwt-validation/src/main.rs
juv: `pong` is there to show case that the `authorize` macro dependencies and the actual function dependencies does not need to match. here the authorization handler got one dependency while `pong` does not. also, `pong` uses the authorization handler _within_ the same module while `get_email` uses the other examplary authorization handler in another module.  that's all.

Should i clarify it more in the comments? `// Example handler that takes no parameter and is authorized by an authorization handler within another module` and `// Example handler that takes two parameters and is authorized by an authorization handler within another module`? 

we can remove it if you think it's not worth to keep.

rapina/examples/jwt-validation/src/main.rs
juv: no, i added it while testing macro dependency expansion

rapina/examples/jwt-validation/src/main.rs
juv: I've changed the comments a bit. Hope this helps

rapina-macros/src/lib.rs
juv: in the current implementation, body extractors can only be used as an `authorize` dependency in case they are also present in the handler function:

```
#[get("/example")]
#[authorize(authorization_handler(Json<Whatever>))]
async fn pong(x: Json<Whatever>) -> Result<Json<String>> {
...
}
```

However, a authorization-only dependency like the following does not compile. it fails with `the trait bound rapina::extract::Json<JsonWebToken<GoogleClaims>>: FromRequestParts is not satisfied`:
```
#[get("/example")]
#[authorize(authorization_handler(Json<Whatever>))]
async fn pong() -> Result<Json<String>> {
    tracing::info!("this is called within the handler body");
    Ok(Json("success".to_string()))
}
```


--------


Consider this example:

```
async fn authorize_with_body_dependency_reuse(claims: &Json<GoogleClaims>) -> Result<()> {
    tracing::info!(email = %claims.email, "authorizing request before it hits the handler");
    if "this-compares-content-of-the-body" == claims.email.as_str() {
        return Ok(());
    }
    Err(Error::forbidden("Missing permissions"))
}

#[get("/with-body")]
#[authorize(authorize_with_body_dependency_reuse(Json<GoogleClaims>))]
async fn handler_with_body(body: Json<GoogleClaims>) -> Result<Json<String>> {
    tracing::info!("this is called within the handler body");
    Ok(Json("success".to_string()))
}
```

Empty body, or malformed json body will lead to `400 Bad Request`: 
```
{
    "error": {
        "code": "BAD_REQUEST",
        "message": "Invalid JSON in request body: expected value at line 1 column 1"
    },
    "trace_id": "cf1a9002-7e2a-43a1-af0c-00a30e88867c"
}
```

Input body `{"email": "x"}` (which leads to a `false` in the comparison of `authorize_with_body_dependency_reuse`) will lead to `403 Forbidden`: 
```
{
    "error": {
        "code": "FORBIDDEN",
        "message": "Missing permissions"
    },
    "trace_id": "55327e68-b703-41c5-89f0-d0821e654c88"
} 
```

Input body `{"email": "this-compares-content-of-the-body"}` (which leads to a `true` in the comparison of `authorize_with_body_dependency_reuse`) will lead to `200 Ok` with logs:
```
2026-08-16T18:43:57.584890Z  INFO jwt_validation: authorizing request before it hits the handler email=this-compares-content-of-the-body
2026-08-16T18:43:57.584998Z  INFO jwt_validation: this is called within the handler body
```

juv: I just saw that this is not described properly in the documentation file, I missed that reusing the in the section `Body Extractors`, sorry. I can add test cases too. 

But first let's decide if we want to keep this possibility. Or whether we want to get rid of all body-"consuming" authorization dependencies entirely - no matter if they are present on the handler or not.

juv:
Hi @arferreira, thank you! I have addressed most of your comments. Two of them need feedback once more.

After rebasing with `main`, the Github actions started failing:
`Clippy` had some findings unrelated to the main pr, regarding the usage of `is_multiple_of()` instead of modulo operator `x % y == 0`. I've added those changes too.

And regarding CI step [Publish Dry Run](https://github.com/rapina-rs/rapina/actions/runs/31967537023/job/95214886553), I am not so sure honestly. Looks like it's trying to look up the rapina macro in version `0.13.1` from `crates.io` but that version obviously does not include the `authorize` export...? It was added recently as part of this [PR](https://github.com/rapina-rs/rapina/pull/757/changes#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR191).

rapina-macros/src/lib.rs
arferreira: Let's forbid them. The docs already say so: under "Body Extractors" you list `Json<T>`, `Form<T>` and the `Validated` wrappers as not valid authorization only dependencies. The reuse path accepts them anyway whenever the handler declares the same extractor, so the code and the docs disagree and I guess the docs picked the better side.

rapina/examples/jwt-validation/src/main.rs
arferreira: Reads fine now, thanks.

arferreira:
Hey @juv, I think the MSRV bump caused them rather than the other way around. `manual_is_multiple_of` is MSRV-gated and wants `1.87`, and main is green today with the plain `%`. Try reverting the three rust version lines first and see whether those findings go away on their own. 
If they do, put the `%` back in `cache`, `rate_limit` and `seed`, since this PR doesn't otherwise touch those files.

Also, `cargo publish --workspace` does co publish siblings, but `rapina-macros 0.13.1` is already on crates.io, so verification resolves the registry copy and it has no authorize. Nothing for you to do, it goes green after the next release. I'll sort the check out separately.

juv:
> Hey @juv, I think the MSRV bump caused them rather than the other way around. `manual_is_multiple_of` is MSRV-gated and wants `1.87`, and main is green today with the plain `%`. Try reverting the three rust version lines first and see whether those findings go away on their own. If they do, put the `%` back in `cache`, `rate_limit` and `seed`, since this PR doesn't otherwise touch those files.
> 
> Also, `cargo publish --workspace` does co publish siblings, but `rapina-macros 0.13.1` is already on crates.io, so verification resolves the registry copy and it has no authorize. Nothing for you to do, it goes green after the next release. I'll sort the check out separately.

Hi @arferreira, you are right. It's due to the switch to MSRV `1.87`. I guess we can keep these changes then to keep clippy happy :)

rapina-macros/src/lib.rs
juv: Alright. It's now enforced and throws a compile error like this:

```
error[E0277]: the trait bound `rapina::extract::Json<GoogleClaims>: FromRequestParts` is not satisfied
  --> src\main.rs:23:35
   |
23 | #[authorize(authorization_handler(Json<GoogleClaims>))]
   |                                   ^^^^^^^^^^^^^^^^^^ the trait `FromRequestParts` is not implemented for `rapina::extract::Json<GoogleClaims>`
   |
   = help: the following other types implement trait `FromRequestParts`:
             Cookie<T>
             CurrentUser
             Headers
             JsonWebToken<T>
             State<T>
             rapina::extract::Context
             rapina::extract::Path<T>
             rapina::extract::Query<T>
```

rapina-macros/src/lib.rs
```diff
@@ -165,44 +489,77 @@ pub fn delete(attr: TokenStream, item: TokenStream) -> TokenStream {
...
+
+        for attr in &authorize_attrs[1..] {
+            error.combine(Error::new_spanned(attr, message));
+        }
+
+        return error.into_compile_error();
+    }
+
+    // Fail in case multiple #[authorize] macros are present until more than one #[authorize] macro is supported
+    // TODO remove entirely when adding support for multiple #[authorize] macros
```
juv: let's cover this in a seperate issue

juv: i've created this issue for it https://github.com/rapina-rs/rapina/issues/771

rapina-macros/src/lib.rs
```diff
@@ -71,6 +76,325 @@ mod config;
...
+    let mut extracts = Vec::new();
+    let mut arguments = Vec::with_capacity(auth.deps.len());
+    let mut needs_request_parts = false;
+
+    for (index, dependency_type) in auth.deps.iter().enumerate() {
+        let normalized_dependency = normalize_type(dependency_type);
+
+        // Prefer an existing handler parameter over extracting the same syntactically
+        // matching dependency a second time.
+        let matching_handler_parameter = inputs.iter().find_map(|input| {
```
juv: Now that I see this once more, I think, if a handler has multiple parameters with the same type, this "silently" selects the first. I think `Header<T>` is the extractor that is primarily affected here. What do you think? Should we just keep the current code? Or reject those ambiguous parameters? We can also follow up with an issue to support ambiguous parameters somehow.

juv:
Hi @arferreira , I moved my code changes to the new macro folder structure.

Two comments to be reviewed besides of the code: 
https://github.com/rapina-rs/rapina/pull/647#discussion_r3816668741 
https://github.com/rapina-rs/rapina/pull/647#discussion_r3815964074

PR #657, opened 2026-06-25
feat: Redis backend for relay pub/sub (relay-redis feature)

### LINKED ISSUE #567: [Feature]: Redis backend for relay pub/sub (relay-redis feature flag)
### Problem Statement

 The relay pub/sub system currently only ships an `InMemoryBackend`, which means                                                                          
  all subscribers must be on the same process. In multi-node deployments (multiple                                                                       
  server instances behind a load balancer), a push sent to node A will never reach                                                                         
  a subscriber connected to node B, silently with no error or warning.
                                                                                                                                                           
  Redis is the natural fit here: it already ships as an optional dependency for                                                                            
  `cache-redis` users, supports native pub/sub with `PUBLISH`/`SUBSCRIBE`, and is
  the most common infrastructure component already present in deployments that need                                                                        
  a cache. Adding a `relay-redis` backend costs near zero for those users.                                                                                 
                                                                                                                                                           
  This is also an explicitly planned feature, `relay/backend.rs` contains a                                                                               
  `// TODO: relay-redis feature — add RedisBackend behind a feature flag, following the same pattern as cache_redis.rs` comment that has been waiting implementation.                                                                                                                                          


### Proposed Solution

  **Usage**                                                                                                                                                
   
  Once implemented, users would opt in via the feature flag and swap                                                                                       
  `with_relay()` for `with_relay_redis()`:                                                                                                               
                                                                                                                                                           
  ```toml                                                                                                                                                  
  # Cargo.toml
  rapina = { version = "0.11", features = ["relay-redis"]}
```          
```rust                                                                                                                                                                       
  #[tokio::main]
      Rapina::new()
          .with_relay_redis(RelayConfig::default(), "redis://127.0.0.1:6379")                                                                              
          .discover()                                                                                                                                      
          .listen("127.0.0.1:3000")
          .await                                                                                                                                           
  }                                                                                                                                                      
   ```                                                                                                                                                        
  Handler code stays identical and the Relay extractor works the same regardless                                                                             
  of backend:                                                                                                                                              
   ```rust                                                                                                                                                        
  #[post("/orders")]                                                                                                                                     
  async fn create_order(relay: Relay, body: Json<NewOrder>) -> Result<Json<Order>> {
      let order = save_order(&body).await?;                                                                                                                
      relay.push("orders:new", "created", &order).await?;                                                                                                  
      Ok(Json(order))                                                                                                                                      
  }                                                                                                                                                        
   ```                                                                                                                                                      
  Pushes from any node now reach subscribers on all nodes.                                                                                                 
 
**How to implement**

  The `RelayBackend` trait is already designed for this: `RelayHub` takes a                                                                                
  `Box<dyn RelayBackend>`, so a Redis implementation slots in without changing the                                                                         
  hub or connection loop. The `redis` crate is already an optional dependency                                                                              
  (used by `cache-redis`). The implementation follows the same structure as                                                                                
  `cache_redis.rs`.                                                                                                                                        
                                                                                                                                                           
  **1. Add the feature flag — `rapina/Cargo.toml`**                                                                                                        
                                                                                                                                                         
  Add `relay-redis = ["redis"]` to the `[features]` table. The `redis` crate is                                                                            
  already declared as an optional dependency, so no new crate is needed.                                                                                 
                                                                                                                                                           
  **2. Create `rapina/src/relay_redis.rs`**                                                                                                              

  Implement `RelayBackend` for a new `RedisRelayBackend` struct:                                                                                           
   
  - `push(topic, json)` — send `PUBLISH topic json` over a cloned                                                                                          
    `MultiplexedConnection`. The trait expects a `RelayFuture`, so the return                                                                              
    must be wrapped: `Box::pin(async move { conn.publish(...).await... })`. 
    Unlike `InMemoryBackend`, which wraps a synchronous                                                                         
    send in `std::future::ready(...)`, this returns a genuine async future that                                                                            
    suspends until Redis acknowledges the publish.                                                                                                       
  - `subscribe(topic)` — open a **dedicated** `AsyncPubSub` connection via                                                                                 
    `client.get_async_pubsub()`. Redis pub/sub cannot share a multiplexed
    connection with regular commands. Call `.subscribe(topic)` and return a                                                                                
    `RedisTopicReceiver` that wraps it. The return must be wrapped:                                                                                            
    same `Box::pin(async move { ... }).`                                                                                                                
  - `RedisTopicReceiver::recv()` — poll `pubsub.on_message()` for the next message.                                                                        
  - `RedisTopicReceiver::drop()` — send `UNSUBSCRIBE` to clean up the Redis                                                                                
    subscription, following the same cleanup-on-drop pattern as                                                                                            
    `BroadcastReceiver` in `backend.rs`.                                                                                                                   
                                                                                                                                                           
  **3. Register in `rapina/src/lib.rs`**                                                                                                                   
                                                                                                                                                           
  Expose the module behind the feature flag:                                                                                                               
  `#[cfg(feature = "relay-redis")] pub mod relay_redis;`
                                                                                                                                                           
  **4. Wire up in `rapina/src/app.rs`**                                                                                                                    
  Add a `with_relay_redis(config: RelayConfig, redis_url: &str)` builder method
  on `Rapina`, behind `#[cfg(feature = "relay-redis")]`. It should construct a                                                                             
  `RedisRelayBackend` and pass it to `RelayHub::new()` instead of                                                                                        
  `InMemoryBackend`.                                                                                                                                       
                                                                                                                                                           
  Integration tests should be annotated with `#[ignore]` and documented as:                                                                                
  `cargo test --features relay-redis -- --ignored`                                                                                                         
  following the same convention as `cache_redis.rs`.                                                                                                     
                                                                                                                                                           
  **Known limitation:** `PresenceMap` remains in-memory per-node. In a multi-node                                                                          
  deployment, `relay.presence(topic)` will only reflect clients connected to the                                                                           
  local node, not the full cluster. This will be addressed in a follow-up issue. 


### Alternatives Considered

_No response_

### Contribution

### PR DESCRIPTION
**Pull request by mvanhorn**
## Summary

This adds a Redis-backed relay pub/sub backend so the relay can route messages across nodes, addressing the multi-node gap described in #567. Today the relay ships only `InMemoryBackend`, which means subscribers must live in the same process: a push on node A never reaches a subscriber on node B, and the message is silently dropped. Because the `RelayBackend` trait and `RelayHub::new(Box<dyn RelayBackend>)` were already designed for pluggable backends, and a `// TODO: relay-redis feature` comment in `backend.rs` already marked this as planned, a Redis implementation slots in without touching the hub or the connection loop.

The new `RedisRelayBackend` implements the existing `RelayBackend` trait, publishing each message with a Redis `PUBLISH` over a cloned multiplexed connection and subscribing each topic over a dedicated async Pub/Sub connection (Pub/Sub cannot share the multiplexed connection). It follows the established `cache_redis.rs` pattern and reuses the `redis` crate that is already an optional dependency, so no new crate is introduced. A new async `Rapina::with_relay_redis(config, redis_url)` builder connects to Redis and substitutes this backend for the default `InMemoryBackend`. The feature is gated behind a new `relay-redis` flag, so the default build is unchanged and the module and builder are absent without the flag.

One known limitation is documented in the builder rustdoc: presence tracking remains per-node because the relay `PresenceMap` is not shared across nodes by this implementation. Subscribe-time failures (a refused Pub/Sub connection or a failed `SUBSCRIBE`) are surfaced through `tracing::error!` rather than swallowed silently, since the trait's `subscribe` method is infallible and cannot otherwise propagate the error. Dropping a receiver drops its dedicated Pub/Sub connection, which Redis treats as an implicit unsubscribe, so no leaked subscription remains.

## Related Issues

Closes #567

## Checklist

- [x] `cargo fmt` passes
- [x] Tests pass (default `cargo build` compiles with the module absent; building and test-compiling with the `relay-redis` feature both succeed; the Redis integration test is `#[ignore]` and runs against a live Redis via `cargo test --features relay-redis -- --ignored`, matching the `cache_redis.rs` convention)
- [x] Documentation updated (module docs, builder rustdoc, and the per-node presence limitation)

### CODE REVIEW

File: rapina/src/relay/redis.rs [hunk 1]
```diff
@@ -0,0 +1,148 @@
...
+                        pubsub: None,
+                        topic,
+                    }) as Box<dyn TopicReceiver>;
+                }
+            };
+
+            if let Err(e) = pubsub.subscribe(&topic).await {
+                tracing::error!(error = %e, topic = %topic, "relay redis subscribe failed");
+                return Box::new(RedisTopicReceiver {
+                    pubsub: None,
```
ShiraiEd: Could you change these two `pubsub`s to `stream` to match the updated `RedisTopicReciever`

rapina/src/relay/redis.rs
```diff
@@ -0,0 +1,148 @@
...
+                    topic,
+                }) as Box<dyn TopicReceiver>;
+            }
+
+            Box::new(RedisTopicReceiver {
+                pubsub: Some(pubsub),
```
ShiraiEd: Could you split the `pubsub` with `.split()` and pass the `stream`  to the Reciever:
```rust
let (_sink, stream) = pubsub.split();
Box::new(RedisTopicReciever { stream: Some(stream), topic, })
```

rapina/src/relay/redis.rs
```diff
@@ -0,0 +1,148 @@
...
+    }
+}
+
+/// Receives messages from a dedicated Redis pub/sub connection.
+///
+/// Dropping the receiver drops that connection, which Redis treats as an
+/// implicit unsubscribe/disconnect. Explicit `UNSUBSCRIBE` is async and cannot
+/// be awaited from `Drop`.
+struct RedisTopicReceiver {
+    pubsub: Option<redis::aio::PubSub>,
```
ShiraiEd: We will need to use the stream part of the split pubsub, so could you change:
```rust
//pubsub: Option<redis::aio::PubSub>, 
//to: 
stream: Option<redis::aio::PubSubStream>,
``` 

this way the PubSubStream can be used

rapina/src/relay/redis.rs
```diff
@@ -0,0 +1,148 @@
...
+    topic: String,
+}
+
+impl TopicReceiver for RedisTopicReceiver {
+    fn recv(&mut self) -> RelayFuture<'_, Option<Arc<String>>> {
+        Box::pin(async move {
+            let pubsub = self.pubsub.as_mut()?;
+            let mut messages = pubsub.on_message();
```
ShiraiEd: `pubsub` -> `stream`

rapina/src/relay/redis.rs
```diff
@@ -0,0 +1,148 @@
...
+            let msg = messages.next().await?;
+
+            if msg.get_channel_name() != self.topic {
```
ShiraiEd: add a comment for future follow-ups pls:
```rust
  // Redundant today: each receiver subscribes to exactly one channel, so every
  // message here already belongs to `self.topic`. Kept defensively, becomes
  // necessary if a shared connection ever multiplexes multiple channels and we
  // need to demux by channel name.
```

rapina/src/relay/redis.rs
```diff
@@ -0,0 +1,148 @@
...
+                return None;
+            }
+
+            msg.get_payload::<String>().ok().map(Arc::new)
+        })
+    }
+}
+
+impl Drop for RedisTopicReceiver {
```
ShiraiEd, line 129: add comment for future follow-ups in this too pls:
```rust
      // One dedicated connection per topic: dropping the stream closes the
      // connection, and Redis treats that as an implicit unsubscribe. No explicit
      // UNSUBSCRIBE is needed here (and it couldn't run anyway cuz Drop is sync,
      // UNSUBSCRIBE is async). Explicit UNSUBSCRIBE only becomes necessary with a
      // shared connection (planed as follow-up), and even then it lives in the
      // backend's refcounted subscription management, not in this Drop.
```

ShiraiEd (changes requested):
Hi @mvanhorn, thanks for the contribution

This still needs an update in the websocket docs (rapina/docs/content/docs/core-concepts/websockets.md) and in the CHANGELOGs

Also I added some comments about `pubsub`:

On the current implementation, `recv()` calls `pubsub.on_message()` on every message, so the message stream is recreated and dropped on each call (the underlying connection stays open the whole time, it's just re-wrapped each time). Using `split()` to keep a long-lived `PubSubStream` in the receiver instead of the full `PubSub` would let us loop over one persistent stream via `.next()`, avoiding the per-message re-wrap and the small window between two recv calls.

This way `sink` would not be used here, but it could be used in the future follow-ups to  enable multiple channels subscriptions

mvanhorn:
Thanks for the thorough review @ShiraiEd - addressed in 8308c75: websocket docs updated (core-concepts/websockets.md), CHANGELOGs added, and recv() reworked per your pubsub comments. Full test suite including the socket-based relay tests passes locally.

ShiraiEd (dismissed):
LGTM

rapina/src/relay/redis.rs
```diff
@@ -0,0 +1,154 @@
...
+            let stream = self.stream.as_mut()?;
+            let msg = stream.next().await?;
+
+            // Redundant today: each receiver subscribes to exactly one channel, so every
+            // message here already belongs to `self.topic`. Kept defensively, becomes
+            // necessary if a shared connection ever multiplexes multiple channels and we
+            // need to demux by channel name.
+            if msg.get_channel_name() != self.topic {
+                return None;
+            }
```
arferreira: Doesn't `None` here mean "stream closed" to the forwarding task? If someone publishes binary to the topic, or a foreign message hits the mismatch check above, the subscription dies for good and the decode path doesn't even log. Shouldn't these skip and keep reading instead? Feels like the defensive check is doing the opposite of defending.

rapina/src/relay/redis.rs
```diff
@@ -0,0 +1,154 @@
...
+struct RedisTopicReceiver {
+    stream: Option<redis::aio::PubSubStream>,
+    topic: String,
+}
+
+impl TopicReceiver for RedisTopicReceiver {
+    fn recv(&mut self) -> RelayFuture<'_, Option<Arc<String>>> {
+        Box::pin(async move {
+            let stream = self.stream.as_mut()?;
+            let msg = stream.next().await?;
```
arferreira: What happens when Redis blips mid-stream? `next()` gives `None`, task ends, and the client sits on a live ws receiving nothing forever. Reconnect can be a follow up, but at least a `warn!` when the stream dies while the receiver is still held? I've been using this a lot and a silent stop is the first thing I would hit in production.

docs/content/docs/core-concepts/websockets.md
```diff
@@ -84,6 +84,23 @@ async fn main() -> std::io::Result<()> {
...
+```rust
+Rapina::new()
+    .with_relay_redis(RelayConfig::default(), "redis://127.0.0.1:6379")
+    .await?
+    .discover()
+    .listen("127.0.0.1:3000")
+    .await
+```
+
+Redis pub/sub delivers relay messages across every Rapina node connected to the same Redis instance. Presence tracking remains per-node and is not shared through Redis.
```
arferreira, line 109: We should say somewhere that it's one Redis connection per active subscription. 10k subscribers is 10k connections. Is the shared connection follow up tracked in an issue somewhere @ShiraiEd? Also not sure this snippet compiles, `?` on `rapina::Error` inside an `std::io::Result` main.

arferreira (commented):
I think we should fix `recv()` returning `None` for skippable messages, because it kills subscriptions permanently.

docs/content/docs/core-concepts/websockets.md
ShiraiEd: Still no follow-up, was waiting this to get finished first,
gonna create it

ShiraiEd: @arferreira,
following the current pattern in relay/
the follow-up should create a Broadcaster to manage the single connection, 
the topics
and the subscribers?

mvanhorn:
@arferreira good catches, all three in d035902.

You were right that the defensive check was doing the opposite of defending. recv() returned None for both an undecodable payload and a channel mismatch, and None reads as "stream closed" to the forwarding task, so one bad message killed the subscription for good. Both paths now log and keep reading. None means only one thing now: the stream is actually exhausted.

Added the warn! on stream death too, with the topic, so a Redis blip isn't silent.

Docs now say each active subscription holds its own Redis connection, and the example is fixed. You were right that it wouldn't compile, it used ? on a rapina error inside a main returning std::io::Result.

Tests cover both skip paths, #[ignore]d like the other Redis tests since they need a live server. Reconnect and @ShiraiEd's Broadcaster stay follow-ups.

arferreira (commented):
Hey, good work so far, two things before merge:

1 - This needs a rebase, also fixing that changelog conflicts.
2 - #758 tracks the shared connection, but cross node presence and reconnect after a Redis blip aren't tracked anywhere. Worth an issue each, or fold reconnect into #758's Broadcaster work since it owns the connection lifecycle.

mvanhorn:
Thanks @arferreira. On 1, the branch is now merged up with main (698523e9 on Sep 7) and the CHANGELOG conflict is resolved, so it should be mergeable again. On 2, I opened #801 for cross-node presence and #802 for reconnect after a Redis drop. I left #802 as its own issue rather than folding it into #758, but noted there that the Broadcaster work owns the connection lifecycle, so if you would rather track it inside #758 feel free to close #802 into it.

PR #665, opened 2026-07-02
feat: Graphql context(current_user, trace_id) and error adapter

### LINKED ISSUE #601: [Feature]: RapinaGraphQLContext with auth/trace_id bridging + error adapter
### Problem Statement

Step 2 of the [Plan](https://github.com/luanperosa/rapina-poc-graphql/blob/master/PLAN.md)

#597 ships the wire-level extractor and responder, but resolvers running through that path have no access to Rapina's request-scoped data, `CurrentUser` and `trace_id` are invisible inside `Context<'_>`. Two consequences:

  1. **Auth-aware resolvers can't be written cleanly.** Authors have to thread auth state through some out-of-band mechanism, or skip auth checks inside GraphQL entirely.
  2. **Errors lose `trace_id`.** Resolvers that fail produce `async_graphql::Error` objects without the `trace_id` extension that Rapina's REST error envelope guarantees, breaking observability for any production debugging workflow.

  PR 3 (builder methods) can't be written without these primitives, because the builder's internal handler is the thing that has to inject the context per-request.

### Proposed Solution

Split `rapina/src/graphql/mod.rs` (introduced in #597) into per-concern files, and add two new ones for this PR:

  ```
  rapina/src/graphql/
  ├── mod.rs            # pub use re-exports
  ├── request.rs        # GraphQLRequest        (moved from #597's mod.rs)
  ├── response.rs       # GraphQLResponse       (moved from #597's mod.rs)
  ├── context.rs        # NEW  RapinaGraphQLContext
  └── error.rs          # NEW  error bridging
  ```

  ### `context.rs`

  async-graphql's `Context<'_>` is a type-map (`ctx.data::<T>()`), not a user-defined struct like Juniper's. We inject a Rapina-specific data type per-request, then resolvers retrieve it via the standard async-graphql pattern.

  ```rust
  /// Request-scoped data injected into every async_graphql::Request before execution.
  pub struct RapinaGraphQLContext {
      pub current_user: Option<CurrentUser>,
      pub trace_id: String,
  }
  ```

  Resolvers access it like this:

  ```rust
  #[Object]
  impl QueryRoot {
      async fn me(&self, ctx: &Context<'_>) -> Result<User> {
          let rapina_ctx = ctx.data::<RapinaGraphQLContext>()?;
          let user = rapina_ctx.current_user
              .as_ref()
              .ok_or_else(|| Error::new("Unauthorized"))?;
          // ...
      }
  }
  ```

  Injection happens in the built-in handler:

  ```rust
  let rapina_ctx = RapinaGraphQLContext {
      current_user: current_user_from_extensions,
      trace_id: request_ctx.trace_id().to_string(),
  };
  let request = gql_request.0.data(rapina_ctx);
  let response = schema.execute(request).await;
  ```

  ### `error.rs`

  Two pieces, a `From` impl for ergonomic `?` propagation, and a helper for explicit error construction with `trace_id`.

  ```rust
  impl From<rapina::Error> for async_graphql::Error {
      fn from(err: rapina::Error) -> Self {
          async_graphql::Error::new(err.message())
              .extend_with(|_, e| {
                  e.set("code", err.error_code());
              })
      }
  }

  /// Construct an async-graphql error with Rapina's error code vocabulary
  /// and the current trace_id populated in extensions.
  pub fn graphql_error(
      code: &str,
      message: impl Into<String>,
      trace_id: &str,
  ) -> async_graphql::Error {
      async_graphql::Error::new(message)
          .extend_with(|_, e| {
              e.set("code", code);
              e.set("trace_id", trace_id);
          })
  }
  ```

  Resolver usage:

  ```rust
  async fn delete_post(&self, ctx: &Context<'_>, id: i32) -> Result<bool> {
      let rapina_ctx = ctx.data::<RapinaGraphQLContext>()?;
      let post = find_post(id).await
          .ok_or_else(|| graphql_error("NOT_FOUND", "Post not found", &rapina_ctx.trace_id))?;
      // ...
  }
  ```

  ### `mod.rs` re-exports

  ```rust
  mod context;
  mod error;
  mod request;
  mod response;

  pub use context::RapinaGraphQLContext;
  pub use error::graphql_error;
  pub use request::GraphQLRequest;
  pub use response::GraphQLResponse;
```

- [x] `RapinaGraphQLContext` is injected via `async_graphql::Request::data()` (not via a `.data()` call on the schema itself, it's per-request, not per-schema)
- [x] `From<rapina::Error> for async_graphql::Error` populates `code` extension from `err.error_code()`
- [x] `graphql_error(code, message, trace_id)` helper sets both `code` and `trace_id` extensions
- [x] Integration test: error response body includes `trace_id` in `extensions`
- [x] Files split per the layout above; `mod.rs` re-exports all four public types
- [ ] Docs page section and changelog updated
- [ ] cargo fmt + cargo clippy passes
(Request context tests move to pr#3)

### Alternatives Considered

**A user-defined `GraphQLContext` trait** (Juniper's pattern, where the user owns the context struct) — rejected because async-graphql's idiomatic pattern is the type-map (`ctx.data::<T>()`), and users can already insert arbitrary additional data via `request.data(...)`. Forcing a trait would fight the library's conventions for no ergonomic gain.

  **An `IntoGraphQLError` trait** mirroring `IntoApiError` — deferred to a follow-up if needed. The `From<rapina::Error>` impl + `graphql_error()` helper cover all v1 use cases. Adding the trait now would add surface without proven demand.

Auditability (contracts, errors, observability)

### Contribution

### PR DESCRIPTION
**Pull request by ShiraiEd**
## Summary

Broke the first pr(#637) mod file into
- `request.rs`
- `response.rs`
 and created `error.rs` and `context.rs`

Now mod just re-exports the structs and function

### Graphql error
The original plan was to implement `impl From<rapina::Error> for async-graphql::Error`,
but it already has an `impl<T: Display + Send + Sync + 'static> From<T> for Error ` on the crate
and rapina Error implements Display too, so they conflic

So I had to implement `ErrorExtentions` for the rapina Error,
with the `extend()` function

Because of that, the `?` operator just returns the error message to the result,
to extend and include the code and trace_id it is need to use `.extend()?` or `.map_err(|e| e.extend())?`

added unit 3 tests

### Graphql context
For this pr I had to just create the struct that will be used on the next one,
containing the `Option<CurrentUser>` and `trace_id`

### Docs
updated the `graphql.md`

Also re-exported `RapinaGraphQLContext` and `graphql_error` from the prelude
(gated by `#[cfg(feature = "graphql")]`).

## Related Issues

Closes #601
2/5 #202

## Checklist

- [ ] `cargo fmt` passes
- [ ] Tests pass

### CODE REVIEW

File: docs/content/docs/core-concepts/graphql.md [hunk 1]
```diff
@@ -47,6 +47,77 @@ in the HTTP status.
...
+    pub current_user: Option<CurrentUser>,
+    pub trace_id: String,
+}
+```
+
+async-graphql's `Context` is a type-map, so resolvers retrieve it with
+`ctx.data::<RapinaGraphQLContext>()`:
+
+```rust
+use rapina::prelude;
```
arferreira: Should this be use `rapina::prelude::*;`?

rapina/src/graphql/error.rs
```diff
@@ -0,0 +1,63 @@
...
+use async_graphql::{Error as GraphQLError, ErrorExtensions};
+
+impl ErrorExtensions for RapinaError {
+    fn extend(&self) -> GraphQLError {
+        GraphQLError::new(self.message()).extend_with(|_, e| {
+            e.set("code", self.code());
+            if let Some(tid) = self.trace_id() {
+                e.set("trace_id", tid);
+            }
+        })
```
arferreira, line 20: Just checking the intent, our REST envelope also has detail when it's set. Did you mean to leave details out?

rapina/src/graphql/response.rs
```diff
@@ -0,0 +1,67 @@
+//! GraphQLResponder
+//! Utilizes graphql-async's `Response` as `GraphQLResponseInner` alias
```
arferreira:
```suggestion
//! Uses async-graphql's `Response` as `GraphQLResponseInner` alias
```

rapina/src/graphql/request.rs
```diff
@@ -0,0 +1,184 @@
+//! GraphQL Request extractor
+//! Utilizes async-graphql's `Request` as `GraphQLRequestInner` alias
```
arferreira:
```suggestion
//! Uses async-graphql's `Request` as `GraphQLRequestInner` alias
```

arferreira (commented):
Good work so far. Left two inline questions and suggestions, thoughts?

rapina/src/graphql/error.rs
ShiraiEd: Oh sorry, totally missed that

I will try to include it,
is the `instance` needed to?

arferreira: Well, I think including `instance` we'll duplicate that, I mean, `path` has already what is needed for locating the failures. The adapter should mirror the standard envelope, `details` when set.

ShiraiEd: @arferreira 
Should this also accepts `details`?:
``` rust
pub fn graphql_error(code: &str, message: impl Into<String>, trace_id: &str) -> GraphQLError {
    GraphQLError::new(message).extend_with(|_, e| {
        e.set("code", code);
        e.set("trace_id", trace_id);
    })
}
```

arferreira: Sorry for the delay here. I'd keep it as is, three args.

rapina/src/graphql/request.rs
```diff
@@ -0,0 +1,184 @@
...
+/// GraphQL request extractor.
+///
+/// Implements `FromRequest`. Accepts:
+///
+/// - **POST** with a JSON body containing `query`, `variables`, and
+///   `operationName`.
+/// - **GET** with the same fields as URL query string params; the
+///   `variables` param is URL-encoded JSON.
+///
+/// Any other HTTP method or malformed input returns a 400 via
```
arferreira: Moved verbatim, but the doc says any other HTTP method returns 400 via `bad_request`, and the code (plus the test) returns 405. Since the doc is relocating anyway, fix it here?

docs/content/docs/core-concepts/graphql.md
```diff
@@ -47,6 +47,77 @@ in the HTTP status.
 The 400 status is reserved for *transport-level* failures: malformed JSON in
 the POST body or an invalid query string in GET. Unsupported HTTP methods return 405.
 
- ### Limitations
+## Request context
 
-This pr just delivers the extractors, Schema management and the `with_graphql` builder method are still to be done!
+`RapinaGraphQLContext` carries Rapina's request-scoped data into resolvers
```
arferreira, line 52: Could you restructure these as full sentences? Commas or parentheses both work.

docs/content/docs/core-concepts/graphql.md
```diff
@@ -47,6 +47,77 @@ in the HTTP status.
...
+    let user = rapina_ctx.current_user.as_ref().ok_or_else(|| {
+        graphql_error("UNAUTHORIZED", "Login required", &rapina_ctx.trace_id)
+    })?;
+    // ...
+}
+```
+
+Automatic per-request injection  pulling `CurrentUser` and `trace_id` from the
+request and calling `request.data(ctx)` for you  arrives with the `with_graphql`
+builder method. Until then, inject it yourself before executing:
```
arferreira: I guess here as well, what do you think?

docs/content/docs/core-concepts/graphql.md
```diff
@@ -47,6 +47,77 @@ in the HTTP status.
...
+
+```rust
+GraphQLResponse(schema.execute(req.0.data(rapina_ctx)).await)
```
arferreira: since `into_inner()` exists, worth using `req.0.data` in the doc example?

rapina/src/graphql/error.rs
```diff
@@ -0,0 +1,108 @@
...
+use crate::error::Error as RapinaError;
+use async_graphql::{Error as GraphQLError, ErrorExtensions};
+
+impl ErrorExtensions for RapinaError {
+    fn extend(&self) -> GraphQLError {
+        GraphQLError::new(self.message()).extend_with(|_, e| {
+            e.set("code", self.code());
+            if let Some(tid) = self.trace_id() {
+                e.set("trace_id", tid);
+            };
```
arferreira:
```suggestion
            }
```

rapina/src/graphql/error.rs
```diff
@@ -0,0 +1,108 @@
...
+            Some(&Value::from_json(serde_json::json!({"field": "email"})).unwrap())
+        )
+    }
+
+    #[test]
+    fn graphql_error_builds_code_and_trace_id() {
+        let gql = graphql_error("INVALID_INPUT", "qty too large", "trace-xyz");
+
+        assert_eq!(gql.message, "qty too large");
+        let ext = gql.extensions.expect("Should have extensions");
```
arferreira:
```suggestion
        let ext = gql.extensions.expect("Should have extentions");
```

arferreira (commented):
We're almost there, just minor issues before we merge.

rapina/src/graphql/request.rs
ShiraiEd: Right, forgot to update that

arferreira:
Let's keep this sleeping for a while, okay?

ShiraiEd:
Sure

PR #673, opened 2026-07-10
Discover `#[metric]` statics alongside routes

### LINKED ISSUE #541: [Feature]: Automatically discover and register custom Prometheus metrics
### Problem Statement

Currently, Rapina offers a great developer experience for discovering endpoint handlers via the `.discover()` macro. The [first iteration on custom Prometheus metrics](https://github.com/rapina-rs/rapina/pull/538) requires the devs to manually register the custom metrics using the `.add_metric()` on the Rapina app builder. 

This leads to an inconsistent workflow:

- Handlers are automatically found and registered
- Metrics, even if defined as static or lazy_static globals, must be manually passed to the builder, often with clone() calls and wiring through multiple layers

As a user, I expect the powerful `.discover()` functionality to also handle metrics, reducing boilerplate and creating a more unified, "hands-free" setup. The current need to mix auto-discovery for one part of the application (handlers) and manual registration for another (metrics) feels inconvenient.

### Proposed Solution

The `.discover()` macro could be enhanced to scan for and automatically register Prometheus collectors.

A potential implementation could involve a new macro, e.g., `#[rapina::metric]`, that marks a static collector for discovery.

```
// src/metrics.rs
use rapina::prometheus::{IntCounter, Opts};
use rapina::prelude::metric; // new attribute macro

#[metric]
pub static ORDERS_TOTAL: Lazy<IntCounter> = Lazy::new(|| {
    IntCounter::new(Opts::new("orders_total", "Total orders")).unwrap()
});


// src/main.rs
Rapina::new()
    .enable_metrics()
    .discover() // .discover() now finds and registers ORDERS_TOTAL automatically
    .listen(...)
    .await; 
```

### Alternatives Considered

_No response_

### Contribution

arferreira:
Handlers are found by `.discover()` but custom collectors have to be threaded through the builder with `add_metric()` and a clone. Now a static holding a collector can be picked up by the same mechanism:

```rust
#[metric]
static ORDERS_TOTAL: LazyLock<IntCounter> = LazyLock::new(|| {
    IntCounter::new("orders_total", "Total orders placed").unwrap()
});

Rapina::new().enable_metrics().discover().listen("127.0.0.1:3000").await
```

Registration needs both calls. When `#[metric]` statics exist and only one is set, we warn at startup naming the missing call, mirroring the health check warning. Misuse gets spanned errors pointing at the user's code: `#[metric]` on a fn, on a `static mut`, or with arguments. Duplicate metric names keep the fail-fast startup panic, which now names the colliding metric since a collector discovered from a dependency makes collisions harder to trace than manual registration.

Fixes #541


rapina-macros/src/lib.rs
```diff
@@ -990,6 +990,83 @@ fn relay_macro_impl(
...
+    };
+    if let syn::StaticMutability::Mut(m) = &item.mutability {
+        return syn::Error::new_spanned(m, "#[metric] cannot be applied to a `static mut`")
+            .to_compile_error();
+    }
+
+    let ident = &item.ident;
+    let collector_fn = quote::format_ident!("__rapina_metric_{}", ident);
+
+    quote! {
```
juv, line 1053: Maybe this new code has a similar _macro ordering problem_ like we did in the draft pull request for authorization (#647). When there is a metric like this:
```
#[metric]
#[cfg(feature = "orders")]
static ORDERS_TOTAL: LazyLock<IntCounter> = ...
```

Do we need to make sure that `#[cfg]`/`#[cfg_attr]` attributes are added to the generated collector function and `inventory::submit!` as well? Otherwise the expanded output of the macro will be something like this:

```
#[cfg(feature = "orders")]
static ORDERS_TOTAL: LazyLock<IntCounter> = ...

fn __rapina_metric_ORDERS_TOTAL() -> Box<dyn Collector> {
    Box::new(ORDERS_TOTAL.clone())
}

inventory::submit! {
    MetricDescriptor {
        collector: __rapina_metric_ORDERS_TOTAL,
    }
}
```

And then the `ORDERS_TOTAL` will be "hidden" during compilation due to the `#[cfg]` macro taking effect after the `#[metric]` expansion happened. This should cause a compilation error or am I missing something here?

juv (changes requested):
Hey, this is a great feature. I left one comment, the rest looks good to me :)

rapina-macros/src/lib.rs
juv: Nevermind, after trying it out, it appears that `#[cfg]` and `#[cfg_attr]` are handled differently during compilation and do not need special consideration in the macro expansion code.

juv (approved):
LGTM

PR #679, opened 2026-07-14
[Bug]: Silent JSON serialization failures across response handlers return empty body with success status

### LINKED ISSUE #594: [Bug]: Silent JSON serialization failures across response handlers return empty body with success status
## Description

The fix in #583 (closing #577) replaced `unwrap_or_default()` with logged fallbacks for validation error details in `extract/mod.rs:529`. However, the same silent-drop pattern exists across many other response paths in the framework. When a user's `Serialize` implementation fails (e.g., a map with non-string keys, a custom serializer returning an error, or a numeric type out of JSON range), the client receives an **empty body with the original status code** — including `200 OK` for successful handler returns — and no log line indicates anything went wrong.

This is the same class of bug as #577, but broader in scope. The validation-details path was the most visible symptom; the underlying anti-pattern is used in at least 9 other places.

### Affected sites

All call `serde_json::to_vec` / `to_string_pretty` / `to_vec_pretty` and discard the error via `unwrap_or_default()`:

- `rapina/src/extract/mod.rs:469` — `(StatusCode, Json<T>)` response body
- `rapina/src/error.rs:427` — error response serialization
- `rapina/src/error.rs:443` — error response serialization
- `rapina/src/health/endpoint.rs:39` — health endpoint body
- `rapina/src/health/endpoint.rs:110` — health endpoint body
- `rapina/src/introspection/endpoint.rs:51` — `/__rapina/routes` body
- `rapina/src/introspection/llms.rs:46` — llms.txt schema pretty-print
- `rapina/src/introspection/llms.rs:55` — llms.txt schema pretty-print
- `rapina/src/openapi/endpoint.rs:43` — `/__rapina/openapi.json` body
- `rapina/src/pagination.rs:226` — paginated response body
- `rapina/src/pagination.rs:365` — paginated response body

The most user-impacting is `extract/mod.rs:469`, since it is on the hot path for any handler returning `(StatusCode, Json<T>)`.

## Steps to Reproduce

1. Define a response type whose `Serialize` implementation can fail (e.g., a `HashMap<NonStringKey, V>` or a custom `Serialize` impl that returns `Err` for some inputs).
2. Return it from a handler as `(StatusCode::OK, Json(value))`.
3. Trigger the failing-serialization input.
4. Observe: client receives `200 OK` with `Content-Type: application/json` and an empty body. No log line is emitted.

## Expected Behavior

When response serialization fails, Rapina should either:

- Log the failure at `error!` (or at least `warn!`) level so it is visible in production, **and** return a `500 Internal Server Error` with a structured error body, **or**
- At minimum, log the failure so operators can diagnose silent empty responses.

The current `unwrap_or_default()` produces an empty `Vec<u8>` and still returns the handler's original status code, which actively misleads clients and is hard to debug from logs alone.

The fix should be uniform across all sites listed above, ideally via a small helper (e.g., `serialize_or_500(value)`) so future call sites do not regress to the same pattern.

## Code Example

```rust
// rapina/src/extract/mod.rs:469 — current
impl<T: serde::Serialize> IntoResponse for (http::StatusCode, Json<T>) {
    fn into_response(self) -> http::Response<BoxBody> {
        let body = serde_json::to_vec(&(self.1).0).unwrap_or_default();
        http::Response::builder()
            .status(self.0)
            .header(CONTENT_TYPE, APPLICATION_JSON)
            .body(crate::response::full(body))
            .unwrap()
    }
}
```

Suggested direction (consistent with the #583 fix):

```rust
let body = match serde_json::to_vec(&(self.1).0) {
    Ok(body) => body,
    Err(e) => {
        tracing::error!("Failed to serialize JSON response body: {e}");
        return crate::error::Error::internal_server_error(
            "Failed to serialize response body",
        )
        .into_response();
    }
};
```

## Environment

- Rapina version: 0.12.0
- Rust version: stable
- OS: any

## Error Output / Logs

No log line is produced. Client sees an empty body with the handler's original status code.

Ismaellima4:
## Summary

What does this PR do?

Replace silent `unwrap_or_default()` serialization patterns with proper error handling that returns structured 500 responses instead of empty bodies with the original status code.

A new `json_response()` helper in `error.rs` centralizes the serialize-or-fail logic: on success it builds a JSON response with the given status, on failure it logs the error and returns a 500 with a structured error body.

**Changed to `json_response()`** — user-provided types that can actually fail serialization:
- `rapina/src/extract/mod.rs:469` — `(StatusCode, Json<T>)` response body
- `rapina/src/pagination.rs:226` — `Paginated<T>` response body
- `rapina/src/pagination.rs:365` — `CursorPaginated<T>` response body

**Kept `unwrap_or_default()`** — framework-owned types that are infallible (derived `Serialize` with standard fields, `serde_json::Value`, string literals):
- `rapina/src/error.rs:427` — RFC 7807 error response (has hardcoded fallback)
- `rapina/src/error.rs:443` — standard error response (has hardcoded fallback)
- `rapina/src/health/endpoint.rs:39` — `serde_json::json!` macro output
- `rapina/src/health/endpoint.rs:110` — `serde_json::Value` from string literals
- `rapina/src/introspection/endpoint.rs:51` — `RouteInfo` with derived `Serialize`
- `rapina/src/introspection/llms.rs:46,55` — `serde_json::Value` schema
- `rapina/src/openapi/endpoint.rs:43` — `OpenApiSpec` with derived `Serialize`

## Related Issues

Closes #594

## Checklist

- [x] `cargo fmt` passes
- [x] Tests pass


ShiraiEd (commented):
Nice PR @Ismaellima4!

Could you add unit tests for `json_response( )` too?
At least one happy path and one unhappy path pls

ShiraiEd (dismissed):
LGTM

rapina/src/error.rs
```diff
@@ -445,7 +489,19 @@ impl IntoResponse for Error {
...
-            let body = serde_json::to_vec(&response).unwrap_or_default();
+            let body = match serde_json::to_vec(&response) {
+                Ok(body) => body,
+                Err(e) => {
+                    tracing::error!("Failed to serialize error response: {e}");
+                    return http::Response::builder()
+                        .status(500u16)
+                        .header(CONTENT_TYPE, APPLICATION_JSON)
+                        .body(crate::response::full(
+                            r#"{"error":{"code":"INTERNAL_ERROR","message":"Failed to serialize error response"},"trace_id":"unknown"}"#.as_bytes(),
```
arferreira: is `trace_id` in the scope here? Since the envelope promises a real `trace_id` on every error, could this fallback interpolate it instead of hardcoding unknown?

rapina/src/error.rs
```diff
@@ -138,6 +164,30 @@ pub mod rfc7807 {
             .collect::<Vec<_>>()
             .join(" ")
     }
+
+    impl IntoResponse for ProblemDetails {
```
arferreira, line 172: This means we can now return it straight from handlers. That's fine by me, but was it intentional or a refactor side effect? If we're keeping it, a doc line on the impl saying so would make it official rather than accidental.

rapina/src/error.rs
```diff
@@ -138,6 +164,30 @@ pub mod rfc7807 {
...
+        fn into_response(self) -> http::Response<crate::response::BoxBody> {
+            let body = match serde_json::to_vec(&self) {
+                Ok(body) => body,
+                Err(e) => {
+                    tracing::error!("Failed to serialize RFC 7807 error response: {e}");
+                    return http::Response::builder()
+                            .status(500u16)
```
arferreira: Any reason not to use `StatusCode::INTERNAL_SERVER_ERROR` like the rest?

arferreira (changes requested):
Good fix, thanks! Two small asks I believe we could address before merge.

rapina/src/error.rs
```diff
@@ -428,24 +482,31 @@ impl IntoResponse for Error {
...
+            };
+            let body = match serde_json::to_vec(&response) {
+                Ok(body) => body,
+                Err(e) => {
+                    tracing::error!("Failed to serialize error response: {e}");
+                    let body = format!(
+                        r#"{{"error":{{"code":"INTERNAL_ERROR","message":"Failed to serialize error response"}},"trace_id":"{trace_id}"}}"#
+                    );
+                    return http::Response::builder()
+                        .status(500u16)
```
arferreira: Sorry, can you also fix the status here?
```suggestion
                        .status(StatusCode::INTERNAL_SERVER_ERROR)
```

arferreira (dismissed):
LGTM, one tiny thing I left a suggestion so please just click on apply inline suggestion and I will merge!

arferreira:
Oops

arferreira (approved):

Issue #678, opened 2026-07-13
[Bug]: [schema] can't support multiple relationships with same type


sougiovn:
### Description

When creating an entity, if using same Entity type for different relationships, it gets a compilation error: [E0119](https://doc.rust-lang.org/error_codes/E0119.html)

### Steps to Reproduce

Try to declare this schema:

```rust
use rapina::prelude::*;

schema! {
    #[table_name = "accounts"]
    Account {
        #[unique]
        name: String,
        initial_value: f64,
        initial_value_date: DateTime,
    }

    #[table_name = "transactions"]
    Tx {
        from: Option<Account>,
        to: Option<Account>,
        date: Date,
        description: Text,
    }
}
```

### Expected Behavior

The macro accepts both type of relationships in the between the tables.

### Code Example

```rust
use rapina::prelude::*;

schema! {
    #[table_name = "accounts"]
    Account {
        #[unique]
        name: String,
        initial_value: f64,
        initial_value_date: DateTime,
    }

    #[table_name = "transactions"]
    Tx {
        from: Option<Account>,
        to: Option<Account>,
        date: Date,
        description: Text,
    }
}
```

### Environment

- Rapina version: 0.12.0
- Rust version: 1.96.0
- OS: MacOS 26.5.1

### Error Output / Logs

```shell
conflicting implementations of trait rapina::sea_orm::Related<account::Entity> for type tx::Entity
conflicting implementation for tx::Entity (rustc E0119)

error[E0119]: conflicting implementations of trait `rapina::sea_orm::Related<account::Entity>` for type `tx::Entity`
  --> oscar/server/src/entity.rs:3:1
   |
 3 | / schema! {
 4 | |     #[table_name = "accounts"]
 5 | |     Account {
 6 | |         #[unique]
...  |
44 | | }
   | | ^
   | | |
   | |_first implementation here
   |   conflicting implementation for `tx::Entity`
   |
   = note: this error originates in the macro `schema` (in Nightly builds, run with -Z macro-backtrace for more info)
```

PR #719, opened 2026-07-25
fix: issue #678 - an entity having multiple belongsTo to same table r…

Kannaj:
fix: issue #678 - an entity having multiple belongsTo to same table raises compilation error due to multiple <Related> impls

Analyse Phase:
1. Use hashset to track how many belongsTo in a single entity
2. If multiple columns pointing to the same table - tag first as canonical in AnalyzeField

Generate phase:
1. Only generate `<Releated>` impl for the field marked canonical
2. Generate `<Linked>` for the rest

Claude was used to generate the tests and code.

## Checklist

- [x] `cargo fmt` passes
- [x] Tests pass


Kannaj:
Just FYI , theres another bug which this PR wont solve yet. its for cases where an entity can have multiple `hasMany` . 

for example:

```
schema! {
    Category {
        name: String,
        parent: Option<Category>,
        children: Vec<Category>,
    }
}
```

1. Can add a fix for this in the same PR

3. This code also assumes that the first `belongsTo` will get a `Related` impl - which might go against what an actual user might assume. Should we allow the user to specify in the macro which one needs to implement `<Related>` ?

ShiraiEd (commented):
> Just FYI , theres another bug which this PR wont solve yet. its for cases where an entity can have multiple `hasMany` .
> 
> for example:
> 
> ```
> schema! {
>     Category {
>         name: String,
>         parent: Option<Category>,
>         children: Vec<Category>,
>     }
> }
> ```
> 
>     1. Can add a fix for this in the same PR
> 
>     2. This code also assumes that the first `belongsTo` will get a `Related` impl - which might go against what an actual user might assume. Should we allow the user to specify in the macro which one needs to implement `<Related>` ?

## Hi @Kannaj, thanks for contributing!

### About the questions
For the the first one, It would be really nice if you solve that,
but I think it would be better if you open an new issue + PR focusing just in that
because it would go a little to far from the scope of this PR that is focusing on the `Related` impl,

For the second one what do you have in mind?
A optional macro like `#[related]` for explicit specification?

### To add
Whether you keep like this, with the first one being the canonical and implementing `Related`,
or with my suggestion, or with your own approach,
This should be added to the documentation so users can understand the Trait implementation behavior
and use the correct methods.

somewhere around here:
(https://github.com/rapina-rs/rapina/blob/main/docs/content/docs/core-concepts/database.md#the-schema-macro)

Kannaj:
hi @ShiraiEd :) - thank you for reviewing

if we make it explicit in the documentation that you can only use `related`, `find_related` for the first column and use `link` for the rest of the `belongsTo` fields - will it be enough?

wondering if it's better than introducing a new macro for `#[related]`

arferreira:
Yo guys, my take here: I guess docs aren't enough here, actually nobody rereads docs when they reorder fields, sad but true. What do you think about making ambiguity explicit instead: when two or more fields target the same entity, require `#[related]` on exactly one of them and fail compilation with a message saying so, everything else gets `Linked`.

ShiraiEd:
Yeah, just the docs may not be enough here,
with `#[related]` users would know which one impls Related and Linked

This plus the mention in the docs for the new users
@Kannaj what do you say?

Kannaj:
Got it ! will make the changes !

arferreira:
hey do you need any help here?

Kannaj:
Sorry - got a bit busy with my job. Will push a fix by this weekend

ShiraiEd:
This was solved in #763

PR #720, opened 2026-07-29
Fix: Rapina add resource migration missing default id column

### LINKED ISSUE #700: [Docs]: add resource command forgets id generation in migration
### Description

When using `rapina add resource`
if `id` is not included, schema generates an i32 as default, but the column in the migration is not generated
creating this gap between the two

### Steps to Reproduce

1. `rapina new`
2. `rapina add resource` without the `id:` arg
3. check the migrations, there is no ID column

### Expected Behavior

The migration should generate a default ID column for this case

### Code Example

```rust
.col(ColumnDef::new(Items::Id).integer().not_null().auto_increment().primary_key())
```

### Environment

- Rapina version: 0.13.0
- Rust version: 1.95
- OS: macOs Tahoe 26.5.1

### Error Output / Logs

```shell
//without the id column, it always returns `database query error`

2026-07-22T09:49:43.780728Z ERROR request{method=GET path=/items trace_id=808351f2-88ef-4e7b-af1
0-88fb55314c4d}: rapina::database: database query error error=error returned from database: (cod
e: 1) no such column: items.id
```

CapThunder19:
## Summary

Fixes a mismatch between the entity generated by the `schema!` macro and the migration generated by `rapina add resource`.

When `rapina add resource <name> <field:type>...` is run **without an explicit `id:` field**, the generated entity includes the default auto-increment `id: i32` primary key, but the generated migration omitted the corresponding `id` column. This caused runtime errors such as:

```text
database query error error=error returned from database: (code: 1) no such column: items.id
```

## Root Cause

Two independent code generators described the same table but had diverged:

- The `schema!` macro (`rapina-macros/src/schema/generate.rs`) automatically generates:

  ```rust
  pub id: i32
  ```

  with `#[sea_orm(primary_key)]` when no `#[primary_key(...)]` attribute is present.

- `generate_migration` (`rapina-cli/src/commands/codegen.rs`) only generated columns for fields explicitly supplied via the CLI, so no `id` column was created when the user omitted one.

As a result, the generated entity always expected an `id` column while the database schema sometimes didn't contain one.

## Fix

When:

- `primary_key` is `None`, and
- no field named `id` was supplied,

`generate_migration` now prepends the default auto-increment primary key column:

```rust
.col(
    ColumnDef::new(Items::Id)
        .integer()
        .not_null()
        .primary_key()
        .auto_increment()
)
```

This mirrors the existing output generated for an explicit `id: i32` field, keeping implicit and explicit primary key generation consistent.

Custom primary keys, composite primary keys, and explicit `id` fields are unaffected.

## Testing

- ✅ `cargo test -p rapina-cli`
  - **208 tests passed**
  - **0 failed**
- ✅ Existing migration generation tests for:
  - Default primary key
  - Explicit `id` fields
  - Custom primary keys
  - Composite primary keys
  all continue to pass without modification.

Fixes #700


CapThunder19:
pls review this pr and tell me if any changes needed

ShiraiEd (commented):
Very nice, @CapThunder19 !

This probably should have an unit test to assert the id generation 
and one to assert that if an explicit type was passed for the id, it won't be overwritten
e.g. :
- test_generate_migration_injects_default_id_when_absent
- test_generate_migration_preserves_id_type_when_declared

also, a mention in the [docs](https://github.com/rapina-rs/rapina/blob/main/docs/content/docs/cli/commands.md#sensible-defaults) about this would be good too

CapThunder19:
pls review

ShiraiEd (approved):
LGTM

arferreira (approved):

ShiraiEd:
@CapThunder19, thanks again for all the help!

Have you given the repo a star yet?
It would help the project gain more visibility ;)

PR #747, opened 2026-08-02
feat: Improve openapi macros to support more attributes similar to FastAPI + optional serving via  swagger-ui

aparo:
This PR is targeting issue #752  
The actual implementation of Rust macro is quite poor in the ability to define the openapi properties.

To document the PR I copy&paste the documentation

# This the abstract from the documentation of implementation of OpeAPI attribute:

## Operation Metadata

All HTTP verb macros accept a set of optional key-value parameters after the path for richer OpenAPI descriptions:

| Parameter | Type | Default | Description |
|---|---|---|---|
| `id` | string literal | handler name | Stable `operationId`. Use dotted notation like `"users.create"`. |
| `summary` | string literal | humanized handler name | Short one-line summary shown in Swagger UI and docs generators. |
| `description` | string literal | first rustdoc `///` line | Longer prose description of the operation. |
| `tags` | array of string literals | `[]` | Tags for grouping operations in Swagger UI. |
| `deprecated` | boolean | `false` | Marks the operation as deprecated. |
| `group` | string literal | — | Path prefix joined at compile time (see [routing](/docs/core-concepts/routing/)). |

Any combination of parameters is valid. Unknown keys produce a compile error.

### Examples

```rust
// Explicit operation ID and summary:
#[get("/users", id = "users.list", summary = "List all users")]
async fn list_users() -> Json<Vec<User>> { /* ... */ }

// Full metadata on a POST endpoint:
#[post(
    "/users",
    id = "users.create",
    summary = "Create a user",
    description = "Creates a new user account. The email must be unique.",
    tags = ["users"],
)]
async fn create_user(body: Json<CreateUser>) -> Json<User> { /* ... */ }

// Tagging with multiple tags:
#[get("/admin/reports", tags = ["admin", "reports"])]
async fn get_reports() -> Json<Vec<Report>> { /* ... */ }

// Marking a legacy endpoint as deprecated:
#[get("/v1/users", deprecated = true, tags = ["v1"])]
async fn list_users_v1() -> Json<Vec<UserV1>> { /* ... */ }
```

### Operation IDs

When `id` is set it becomes the `operationId` in the spec. Without it, the handler function name is used. Rapina validates uniqueness at startup — two public routes with the same effective `operationId` (whether from `id` or the handler name) cause a panic with a clear message.

Using dotted notation (`"resource.action"`) is recommended when you have multiple API versions or nested resources, because it makes generated SDK clients easier to read.

### Summaries

When `summary` is omitted, Rapina generates one automatically by humanizing the handler name: underscores become spaces and the first letter is capitalized. For most handlers this is good enough; use `summary` when the function name is abbreviated or doesn't read naturally.

### Descriptions

`description` maps to the OpenAPI `Operation::description` field — a longer prose explanation rendered below the summary in Swagger UI. You can also write it as a rustdoc comment on the function and Rapina will pick up the first non-empty line:

```rust
/// Creates a new user account. The email address must be unique across
/// all tenants. Returns 409 if the email is already in use.
#[post("/users", tags = ["users"])]
async fn create_user(body: Json<CreateUser>) -> Json<User> { /* ... */ }
```

When both `description = "..."` on the macro and a `///` comment are present, the explicit attribute takes precedence.

### Tags

Tags are arrays of string literals. They group operations in Swagger UI's sidebar and in most API documentation generators. A single route can belong to multiple tags:

```rust
#[get("/admin/audit-log", tags = ["admin", "audit"])]
async fn get_audit_log() -> Json<Vec<AuditEntry>> { /* ... */ }
```

Tags don't affect routing or authentication — they're purely documentation metadata.

# This is the abstract of implementation of swagger-ui

---

## Swagger UI

Rapina can serve the interactive [Swagger UI](https://swagger.io/tools/swagger-ui/) explorer alongside your spec. Enable it with the `swagger-ui` Cargo feature.

### Setup

Add the feature to your dependency:

```toml
[dependencies]
rapina = { version = "0.13", features = ["swagger-ui"] }
```

Then call `.swagger_ui()` on the app builder. It must be combined with `.openapi()`:

```rust
#[tokio::main]
    Rapina::new()
        .openapi("My API", "1.0.0")
        .swagger_ui()               // serves at /__rapina/swagger/
        .discover()
        .listen("127.0.0.1:3000")
        .await
}
```

The UI is available at `http://localhost:3000/__rapina/swagger/` and automatically points at `/__rapina/openapi.json`.

### Custom path

Use `.swagger_ui_at()` to choose a different URL:

```rust
Rapina::new()
    .openapi("My API", "1.0.0")
    .swagger_ui_at("/docs/")
    .discover()
    .listen("127.0.0.1:3000")
    .await
```

Any path works as long as it doesn't conflict with your application routes. Trailing slashes are recommended but not required.

### CDN dependency

The Swagger UI bundle is loaded from the [unpkg](https://unpkg.com) CDN at runtime (`swagger-ui-dist@5`). The Rapina binary itself stays small — no static assets are embedded. In air-gapped environments you'll need to either proxy the CDN requests or serve the bundle yourself and point a custom HTML page at the spec URL.

### Development vs. production

Swagger UI is useful during development. For production deployments that don't need the explorer, simply omit `.swagger_ui()` — the feature flag has no runtime overhead when the method isn't called. If you want to expose it in staging but not production, wrap the call in a config check:

```rust
let app = Rapina::new()
    .openapi("My API", "1.0.0")
    .discover();

#[cfg(feature = "swagger-ui")]
let app = if config.enable_swagger_ui {
    app.swagger_ui()
} else {
    app
};

app.listen("127.0.0.1:3000").await
```

---

## Handler Names and Operation IDs

By default, handler function names are used as the `operationId` in the spec, and a humanized `summary` is derived automatically:

| Function | Default `operationId` | Default `summary` |
|----------|----------------------|-------------------|
| `list_users` | `list_users` | List users |
| `get_user` | `get_user` | Get user |
| `create_order` | `create_order` | Create order |

Override either with the `id` and `summary` macro parameters:

```rust
#[get("/users", id = "users.list", summary = "Browse users")]
async fn list_users() -> Json<Vec<User>> { /* ... */ }
```

Keep default handler names descriptive — `get_user` reads better than `user` in both the spec and any generated SDK clients. Use the `id` override when you need a stable identifier that won't change even if you rename the function, or when dotted namespacing helps organize a large API.

Path parameters are extracted automatically from `:param` segments in the route path and documented as required path parameters in the spec. `"/users/:id"` becomes `"/users/{id}"` with a required `id` parameter.

# Checklist

- [X] `cargo fmt` passes
- [X] `cargo clippy` has no warnings
- [X] Tests pass
- [X] Documentation updated


juv:
Hi @aparo , I will review this soonish. Thanks for your contribution and your patience.

rapina/src/openapi/spec.rs
juv, line 1: I think we are missing one of the most important features for production-grade applications here: the authentication. 

To get everyone on the same page for the discussion going forward, I have summarized it:

In the OpenAPI spec, these fields are used for it:
[`components.securitySchemes`](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.2.0.md#security-scheme-object): which authentication mechanisms are supported in general
[`security`](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.2.0.md#security-requirement-object): references one or more `securitySchemes` and is applied as the default for each operation, unless an operation has an overridden `security` field


Example: 
1) defining the `securitySchemes`
```
components:
  securitySchemes:
    bearerAuth:
      type: http
      scheme: bearer
      bearerFormat: JWT

    oauth2:
      type: oauth2
      flows:
        authorizationCode:
          authorizationUrl: https://example.com/oauth/authorize
          tokenUrl: https://example.com/oauth/token
          scopes:
            "reports:read": Read administrative reports
            "reports:write": Create or modify administrative reports

    apiKey:
      type: apiKey
      in: header
      name: X-API-Key
```

2) referencing them as default (or not): in this case, `bearerAuth` and `apiKey` are the defaults for all endpoints and take effect unless an "operation object" explicitly overrides `security` within an individual [operation object](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.2.0.md#operation-object)
```
security:
  - bearerAuth: []
  - apiKey: []
```

3) example of overriding for individual operation objects:
```
paths:
  /admin/reports:
    get:
     ...
      security:
        - oauth2:
            - reports:read
      ...
```
Here only oauth2 

4) **public** routes need to override the operation object's `security` field with value `[]`

---

Now there is another caveat in the OpenAPI spec here: For the `security` reference: multiple objects in the array mean `OR` logic. Either authentication type is fine:

```
security:
  - bearerAuth: []
  - apiKey: []
```

If an endpoint requires multiple authentication mechanisms (`AND` logic), they must be given in the same element to indicate that the server handles it this way.
```
    security:
      - bearerAuth: []
        apiKey: []
```

For the sake of the simplicity, I'd argue that the second option is a rare case and can be skipped for now.

juv, line 1: We are missing some important fields within the `info` field still:
`termsOfService`, `contact`, `license`.

I have found these fields to be very helpful in the past, especially when working with services from multiple teams within an organization. Some organizations even enforce the presence of these fields.

Reference: https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.2.0.md#info-object

juv, line 1: we're missing the [`servers` field](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.2.0.md#external-documentation-object), which means that when the openapi spec is served at `https://example.com/openapi.yaml`, the implied base url for all endpoints is `https://example.com/`. If If the API has a distinct host, port, or base path, one must define it explicitly or any example calls within the swagger ui will not work.

rapina-macros/src/lib.rs
juv, line 1: This "first rust doc line ends up as the description" feature, which I get was already in the code prior to this pr, is kind of interesting. Is the developer really aware of this? Do we potentially leak internal documentation?

Besides of that, the docs have this example:
```
/// Creates a new user account. The email address must be unique across
/// all tenants. Returns 409 if the email is already in use.
...
```

The comment was probably cut into two lines by auto-formatting.
But, well, the current implementation only takes the first line of the comment and the `description` will end up like this:
```
"description": "Creates a new user account. The email address must be unique across",
```

Are we okay with this? It kind of looks unintended.

rapina/src/app.rs
```diff
@@ -808,6 +814,56 @@ impl Rapina {
...
+    /// async fn main() -> std::io::Result<()> {
+    ///     Rapina::new()
+    ///         .openapi("My API", "1.0.0")
+    ///         .swagger_ui_at("/docs/")
+    ///         .listen("127.0.0.1:3000")
+    ///         .await
+    /// }
+    /// ```
+    #[cfg(feature = "swagger-ui")]
+    pub fn swagger_ui_at(mut self, path: impl Into<String>) -> Self {
```
juv, line 862: I think we should follow the `enable_...` and `with_...`  naming convention here instead of `swagger_ui(...)` and `swagger_ui_at(...)`.
In that case, you can have the default logic within `enable_swagger_ui`. and another `with_swagger_ui` function which takes a SwaggerUi config that can be extended in the future

rapina/src/app.rs
```diff
@@ -1088,6 +1144,25 @@ impl Rapina {
                     .get_named("/__rapina/openapi.json", "openapi_spec", openapi_spec);
         }
 
+        #[cfg(feature = "swagger-ui")]
```
juv, line 1147: Existing application routes may be shadowed here if they clash with the `swagger_ui_path`.
Example: application route `/whatever` is shadowed by the swagger ui route:

```
#[get("/whatever")]
async fn get_whatever() -> Json<String> {
    ...
}

```

```
.swagger_ui_at("/whatever")
```

Navigating to `/whatever` shows the swagger ui instead of what `/whatever` was doing previously without an error/warning. We should detect conflicts before registration

rapina/src/app.rs
```diff
@@ -808,6 +814,56 @@ impl Rapina {
...
+    ///     Rapina::new()
+    ///         .openapi("My API", "1.0.0")
+    ///         .swagger_ui()
+    ///         .listen("127.0.0.1:3000")
+    ///         .await
+    /// }
+    /// ```
+    #[cfg(feature = "swagger-ui")]
+    pub fn swagger_ui(self) -> Self {
+        self.swagger_ui_at("/__rapina/swagger/")
```
juv, line 839: Accessing `http://127.0.0.1:3000/__rapina/swagger/` works but `http://127.0.0.1:3000/__rapina/swagger` throws a 404 error. The same happens with regular endpoints in Rapina because the trailing slash is not normalized automatically. 

is there a specific reason why you added the trailing slash? I think for most other internal endpoints and the documentation in general we do not use those trailing slashes

rapina/src/app.rs
```diff
@@ -1088,6 +1144,25 @@ impl Rapina {
                     .get_named("/__rapina/openapi.json", "openapi_spec", openapi_spec);
         }
 
+        #[cfg(feature = "swagger-ui")]
+        if let Some(ref ui_path) = self.swagger_ui_path {
```
juv, line 1148: The swagger ui can be enabled when the user does not enable the OpenAPI feature.
You have already pointed that out with the warning in the code but let's add a discussion here whether that is okay. Maybe a compile error would be more meaningful? What do you think? For now, the swagger ui will show an "404"-error page in case the `.openapi` call is missing (as expected). 

Image

rapina/src/app.rs
juv, line 862: The `path` is not validated.
These examples compiles without an error but is broken: 
`.swagger_ui_at("broken")`
`.swagger_ui_at("__rapina/openapi.json")` (claiming reserved internal paths)
`.swagger_ui_at("")`

This is sus too but somehow works:
`.swagger_ui_at("/broken///")`

rapina/src/openapi/swagger_ui.rs
```diff
@@ -0,0 +1,112 @@
...
+  <link rel="stylesheet" href="https://unpkg.com/swagger-ui-dist@5/swagger-ui.css">
+  <style>
+    body {{ margin: 0; }}
+    #swagger-ui {{ max-width: 1460px; margin: 0 auto; }}
+  </style>
+</head>
+<body>
+  <div id="swagger-ui"></div>
+  <script src="https://unpkg.com/swagger-ui-dist@5/swagger-ui-bundle.js"></script>
+  <script src="https://unpkg.com/swagger-ui-dist@5/swagger-ui-standalone-preset.js"></script>
```
juv, line 66: Is using an unpinned version tag fine here?

rapina-macros/src/lib.rs
```diff
@@ -5,18 +5,32 @@ use syn::spanned::Spanned;
...
 struct RouteAttr {
     path: LitStr,
     group: Option<LitStr>,
     description: Option<LitStr>,
+    /// Stable OpenAPI operationId override (e.g. "users.create").
+    id: Option<LitStr>,
+    /// Short one-line summary for OpenAPI (overrides the auto-generated humanized name).
+    summary: Option<LitStr>,
+    /// OpenAPI tags for grouping operations (e.g. `["users", "admin"]`).
+    tags: Vec<LitStr>,
```
juv, line 20: I think there is no validation for empty / whitespaces for the new fields. Is this good to go regardless?
```
tags = ["", "  "]
id = ""
id = "  "
```

juv (changes requested):
Thank you @aparo for taking the time to improve Rapina’s OpenAPI ergonomics!

I added a few questions and discussion points before this is ready to merge.

One broader point: the current OpenAPI generation is still fairly incomplete in its existing state on `main`, independently of this PR. For example, it already has limitations around response status codes, typed path/query parameters, content types, security metadata, and reusable components. I do not necessarily think all of those need to be addressed in this PR, but it would be useful to clarify the intended scope of this pr. Where do we draw the lines?

Issue #735, opened 2026-08-01
Split extractor into submodules


arferreira:
### Suggestion

Let's split our extractor into submodules.  I'm splitting the work into four sub-issues/PRs because reviewing a 1,900 line move in one diff is how mistakes get through!

1. path machinery to path.rs
2. body extractors (Json, Form, Validated)
3. the small ones (Query, State, Context, Cookie, Headers)
4. Db and Jobs impls out of extract/ entirely

**If you want to help:** be aware the route macros generate code referencing these types by path, and a missing re-export won't fail our tests. It only breaks user code. That risk is concentrated in the first PR, so I'll take that one myself. The others should be safe to pick up once it lands.

### Context

Our extractor is turning into a mess, everything ended up in one file and it's at 1,911 lines, and about half of that is the test module. Everything lives there: the `FromRequest/FromRequestParts` traits, `Json`, `Path`, `Query`, `Form`, `Headers`, `Cookie`, `State`, `Context`, `Validated`, `PathParams`, and the custom deserializer for path params. Funny enough, header and multipart already sit in that directory as their own files. The rest never got the same treatment.

### Scope

Other

### Examples

```rust
My suggestion is to be structured in this way:


extract/
── mod.rs        traits, blanket FromRequestParts -> FromRequest impl, re-exports
── path.rs       Path<T>, PathParams, PathParamsDeserializer, StrDeserializer
── json.rs       Json<T>
── form.rs       Form<T>
── validated.rs  Validated<T> and its impls over Json/Form
── query.rs      Query<T>
── state.rs      State<T>
── cookie.rs     Cookie<T>
── context.rs    Context
── header.rs     already exists, absorbs Headers
── multipart.rs  already exists
```

This is a pure code move. Everything stays re-exported from `extract`, so `rapina::extract::Json` and the prelude don't change, and the test suite should pass untouched. Each extractor takes its rustdoc, `into_inner`, trait impls and tests with it.

PR #748, opened 2026-08-02
Move Path and path param machinery to extract/path

### LINKED ISSUE #736: Move Path, PathParams and the params deserializer to extract/path.rs
### Problem Statement

Part of #735. `Path<T>`, `PathParams`, `PathParamsDeserializer` and `StrDeserializer` move to `extract/path.rs`. It's the biggest self contained piece of mod.rs, plus the struct definitions and their tests) and nothing else in the file depends on it. 

Whatever shape this PR lands with is the shape the other slices will copy, so it's worth being picky here about what goes where.

### Proposed Solution

`Path` and `PathParams` get re-exported from `mod.rs` so public paths don't change. The two deserializers are private and can stay private.

### Alternatives Considered

**One thing to check:** `PathParams::push` is `pub(crate)` and the router calls it. Visibility survives the move fine, but grep for anything referencing it through the full module path.

Other / Not sure

### Contribution

arferreira:
Part of #735. Moves `Path<T>`, `PathParams`, `extract_path_params` and the private serde plumbing behind them into `extract/path.rs`, with their rustdoc and tests.
 
`path` is private and re-exported from `mod.rs`. That keeps `rapina::extract::Path` as the only path the docs list, and avoids adding eight new public module paths over the course of #735. It's different from `pub mod header` and `pub mod multipart`, on purpose: those two re-export everything they expose anyway, so being public only added surface we now owe compatibility on. The other slices should follow this one.

Follow-up: `extract_path_params` is `pub`, but its only caller is the fallback matcher in `router/mod.rs`. It probably belongs in router, or should be `pub(crate)`. Both are breaking changes, so neither fits in a move.

Closes #736.


ShiraiEd (approved):
LGTM

PR #750, opened 2026-08-02
Respect column attributes on composite primary key fields

### LINKED ISSUE #745: [Bug]: #[column] rename not respected in composite primary key
### Description

When creating an entity with composite primary key, if using `#[column]` to rename the column in db, the `#[column]` is completely ignored.

### Steps to Reproduce

Run the following schema:

```rust
use rapina::prelude::*;

schema! {
    #[table_name = "labels"]
    Label {
        #[unique]
        name: String,
        parent: Option<Label>,
    }

    #[table_name = "transactions"]
    Tx {
        value: f64,
    }

    #[table_name = "transactions_labels"]
    #[timestamps(none)]
    #[primary_key(tx_id, label_id)]
    TxLabel {
        #[column="transaction_id"]
        tx_id: i32,
        label_id: i32,
    }
}
```

Check generated table, it will have `tx_id` as column name, instead of `transaction_id`. 

### Expected Behavior

The macro would respect the `#[column]` rename when declaring the composite primary key.

### Code Example

```rust
use rapina::prelude::*;

schema! {
    #[table_name = "labels"]
    Label {
        #[unique]
        name: String,
        parent: Option<Label>,
    }

    #[table_name = "transactions"]
    Tx {
        value: f64,
    }

    #[table_name = "transactions_labels"]
    #[timestamps(none)]
    #[primary_key(tx_id, label_id)]
    TxLabel {
        #[column="transaction_id"]
        tx_id: i32,
        label_id: i32,
    }
}
```

### Environment

- Rapina version: 0.12.0
- Rust version: 1.96.0
- OS: MacOS 26.5.1

### Error Output / Logs

```shell

```

arferreira:
Closes #745.

We were silently ignoring #[column = "..."] on any field listed in #[primary_key(...)]. `generate_custom_pk_fields` built those fields from scratch and only ever emitted `primary_key`, `auto_increment = false`, never reading `field.attrs`, while `generate_model_fields` filtered the primary key columns out of the path that does read them. Two functions producing the same thing, one of them a stub.

The regression test asserts on `Column::TxId.as_str()`, the name that actually reaches SQL, rather than on the generated token stream, so it survives a change in how the attribute is spelled. `TestRegion` covers primary key, rename and column type together, which is the first case in the suite where the generated two-attribute output has to compile.

This should go out as a patch, ahead of 0.14.0.

### DIFF
--- a/rapina-macros/src/schema/generate.rs
+++ b/rapina-macros/src/schema/generate.rs
@@ -723,6 +723,23 @@ mod tests {
     }
 
     #[test]
+    fn test_generate_custom_pk_keeps_column_type() {
+        let input = quote! {
+            #[primary_key(code)]
+            #[timestamps(none)]
+            Region {
+                code: Text,
+            }
+        };
+
+        let parsed = parse_schema(input).unwrap();
+        let analyzed = analyze_schema(parsed).unwrap();
+        let output = generate_schema(analyzed).to_string();
+
+        assert!(output.contains("column_type = \"Text\""));
+    }
+
+    #[test]
     fn test_generate_default_pk_unchanged() {
         // Entities without #[primary_key] should still get auto id
         let input = quote! {
--- a/rapina-macros/src/schema/generate.rs
+++ b/rapina-macros/src/schema/generate.rs
@@ -118,21 +118,47 @@ fn generate_entity_module(entity: &AnalyzedEntity, schema: &AnalyzedSchema) -> T
     }
 }
 
+fn sea_orm_parts(field: &AnalyzedField) -> Vec<TokenStream> {
+    let mut parts = Vec::new();
+    if field.attrs.unique {
+        parts.push(quote! {unique});
+    }
+    if field.attrs.indexed {
+        parts.push(quote! {indexed});
+    }
+    if let Some(ref col_name) = field.attrs.column_name {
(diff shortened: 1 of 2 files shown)


ShiraiEd (approved):
LGTM

This pr solves the bug, but I have noticed that
if a user want a custom pk i32, or an auto incremented i64, in the current way is not possible
this could be a follow-up, what do you think?

arferreira:
Yeah, actually we're on the default (`i32`) the rest of the ecosystem walked away from, deserves a follow-up ASAP, the custom name one I'd have pushed back on if it were only about someone preferring `user_id` to `id`. Convention over configuration is our pitch and your key is called `id` is a fine thing to be opinionated about, however that import database feature changes it a bit, because it introspects a live schema, 100% sure legacy databases are full of these custom ids and importing one today probably gives you a schema the macro can't represent, it doesn't detect auto increment to begin with.
Want to take either of them?

ShiraiEd:
Just to be on the same page,
the follow-ups would be for the i64 and the auto-increment detection for the db import feature?

arferreira:
The `i64` one is exactly that, the second is two halves though, and import detection is only one of them. I mean, a field attribute the macro understands, and the detection in the import that fills it in. Doing only the detection leaves holding information it can't express.

PR #754, opened 2026-08-04
Release 0.13.1

arferreira:
Patch release for #750. `#[column]` was ignored on any field named in `#[primary_key(...)]`, so generated entities used the field name as the column.

Anyone who had `#[column]` on a primary key field has tables created with the field name and needs a rename migration after upgrading. The changelog entry says so explicitly, because the gh release notes are generated from commit titles and won't carry it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,11 @@ Routine dependency-only updates are intentionally omitted unless they change use
 
 ## [Unreleased]
 
+## [0.13.1] - 2026-08-04
+
+### Fixed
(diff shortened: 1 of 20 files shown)


ShiraiEd (dismissed):
Merge when checks are solved

ShiraiEd (commented):
Looks like the features version still to be updated, the snapshots too,
not sure about this one:

>(node:2189) [DEP0169] DeprecationWarning: `url.parse()` behavior is not standardized and prone to errors that >have security implications. Use the WHATWG URL API instead. CVEs are not issued for `url.parse()` vulnerabilities.
>Error: Resource not accessible by integration

arferreira:
> Looks like the features version still to be updated, the snapshots too,
> not sure about this one:

this is obvious lol

ShiraiEd (approved):
ship it

PR #755, opened 2026-08-04
Ignore rkyv advisory and let the audit job report findings

arferreira:
`rkyv` is an optional dependency of rust_decimal that no feature of ours activates, so it's never compiled, and Cargo Deny stayed green on the same commit because it resolves the feature graph instead of scanning lockfile entries.


ShiraiEd (approved):
ship it

PR #669, opened 2026-07-04
feat(cli): replace Ferris banner with Juca unicode art

CapThunder19:
## Summary

This PR replaces the Ferris (Rust mascot) startup banner with Juca, Rapina's mascot.

image

## Related Issues

Closes #633

PR #757, opened 2026-08-05
fix: juca.txt move inside rapina-cli

### LINKED ISSUE #756: Fix 0.13.1 release error
Error:
```bash
error: couldn't read `src/commands/../../../juca.txt`: No such file or directory (os error 2)
  --> src/commands/dev.rs:12:27
   |
12 | const JUCA_BANNER: &str = include_str!("../../../juca.txt");
   |                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
help: there is a file with the same name in a different directory
   |
12 | const JUCA_BANNER: &str = include_str!("../../../../../juca.txt");
   |                                                  ++++++

error: couldn't read `src/../../juca.txt`: No such file or directory (os error 2)
   --> src/main.rs:586:27
    |
586 | const JUCA_BANNER: &str = include_str!("../../juca.txt");
    |                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
help: there is a file with the same name in a different directory
    |
586 | const JUCA_BANNER: &str = include_str!("../../../../juca.txt");
    |                                               ++++++

error: could not compile `rapina-cli` (bin "rapina") due to 2 previous errors
error: failed to verify package tarball
Error: Process completed with exit code 101.
```

ShiraiEd:
## Summary

Moves `juca.txt` to inside `rapina-cli`, so it can be release together with the crate

## Related Issues

Closes #756


arferreira (changes requested):
Thanks for the quick fix, one thing I'd like in the same PR would be a `cargo publish --dry-run` step in CI covering all three crates. That's what would have caught this back in #669.

ShiraiEd:
like this?
        - name: Publish dry run
          run: cargo publish --workspace --dry-run --locked

right after "Run tests"? in the Check job

arferreira:
Good question, I'd give it its own job rather than using it in check. Also `cargo publish --workspace --dry-run --locked` so --workspace resolves the inter crate deps against the local packages.

arferreira:
Pushed the CI job onto your branch instead of waiting, since it's the middle of the night for you and `rapina-cli 0.13.1` still isn't on crates.io while the other two are. Didn't want to leave the workspace in that state overnight.

arferreira (approved):

ShiraiEd:
Sorry, I really didn't have the time yesterday

Issue #740, opened 2026-08-01
Split rapina-macros lib.rs into submodules


arferreira:
Same story as #735, now on the macro crate. `rapina-macros/src/lib.rs` is at 2,315 lines, about 850 of them tests. The `schema/` submodule already shows the shape we want (entry point in lib.rs, implementation in its own module), but everything else still lives at the root: the five route macros plus `#[public]`, `route_macro_core` with its helper cluster, `#[relay]`, `#[job]`, and `derive(Config)`.

Proc macro entry points have to stay at the crate root, so lib.rs keeps the thin `#[proc_macro_*]` functions and everything else moves:

```
rapina-macros/src/
── lib.rs           entry points only
── route/
  ── mod.rs       RouteAttr, join_paths, route_macro_core
  ── headers.rs   detect_header_type, collect_header_params, gen_header_extraction
     ── metadata.rs  doc/errors/cache/public attr extraction, request body meta
   ── relay.rs
   ── job.rs
   ── config.rs        derive(Config)
   ── schema/          already done
```

These are pure moves. The generated code uses absolute `::rapina::` paths, so token output can't change, and the only public surface of a proc macro crate is its entry points. Each macro takes its tests along.

Unlike the extract split there's no ordering constraint. The three standalone macros are safe to pick up right away:

1. move `#[relay]` to relay.rs
2. `#[job]` to job.rs
3. move `derive(Config)` to config.rs
4. move the route machinery to route/

PR #762, opened 2026-08-13
refactor: moves the job macro implementation

ShiraiEd:
## Summary

- Moves `job_macro_impl` and `JobAttr` and its impls and tests to `job.rs`
- Keeps `#[proc_macro_attribute]` in the root since it can't be moved form there
declared the its `job_macro_impl` pub(crate) instead of just pub or pub(super) for better readability
(the two of them makes the fn pub inside the crate, but pub(crate) tells it explicitly)

## Related Issues

Closes #742

## Checklist

- [x] `cargo fmt` passes
- [x] `cargo clippy` has no warnings
- [x] Tests pass
- [ ] Documentation updated (if needed)

PR #764, opened 2026-08-14
Move the relay macro to relay.rs

### LINKED ISSUE #741: Move the relay macro to relay.rs
Part of #740. `relay_macro_impl` moves to `relay.rs`; the `#[proc_macro_attribute] pub fn relay` entry stays in lib.rs and calls into it, same as schema does today. Its tests in the bottom test module move too. Nothing else in lib.rs touches this code.

arferreira:
Closes #741

Part of #740. `relay_macro_impl` moves to `relay.rs` along with its tests, and the `#[proc_macro_attribute] pub fn relay` entry point stays at the crate root and calls into it, the way `schema` already works. The function had no helpers and nothing else in the crate referenced it, so this is a lift with no rewiring. Output can't change either, since what the macro emits lives inside a `quote!` block.


ShiraiEd (dismissed):
LGTM

Making `relay_macro_impl` pub(crate) instead of just pub would improve readability by visibility explicitness or would be just nagging by my side?
In #762 I made it pub(crate) because of this

arferreira:
Not nagging, actually you're right. Changed it here!

ShiraiEd (approved):
Nice

PR #769, opened 2026-08-18
Move the Config derive into its own module

### LINKED ISSUE #743: Move derive(Config) to config.rs
Part of #740. `derive_config_impl` moves to `config.rs`, the `#[proc_macro_derive]` entry stays in lib.rs. Note the entry point and the impl currently sit ~70 lines apart with the `schema` entry between them, which is exactly the kind of thing this split fixes. Tests go along.

arferreira:
Part of #740, closes #743.

`derive_config_impl` lived well below its own entry point with the `schema` docs in between. It moves into its own module. The `#[proc_macro_derive]` stays at the crate root. The macro had no unit tests, so this adds some.

### DIFF
--- a/rapina-macros/src/config.rs
+++ b/rapina-macros/src/config.rs
@@ -147,4 +147,23 @@ mod tests {
         assert!(output_str.contains("\"API_KEY\""));
         assert!(output_str.contains("push"));
     }
+
+    #[test]
+    #[should_panic(expected = "Config derive only supports structs")]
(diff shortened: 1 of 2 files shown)


ShiraiEd (commented):
You could add a tests to check if rejects non-structs and one more to tests if rejects non named fields like tuple structs too

ShiraiEd (approved):
LGTM

PR #772, opened 2026-08-26
dependabot: groups the tungstenite crates into one single PR

### LINKED ISSUE #697: Group the tokio and hyper tungstenite dependabot bumps
## Why
When the tungstenite crates get updated, dependabot creates separete PRs to bump each one,
but the problem is that every time the bot tries to bump them, the checks always fails( like #688 and #692)
the `tungstenite` crates must be on the same version to compile

## Suggested solution
Use dependabot's grouping to group those updates together so all get updated to the same version at the same time

### example
```yml
groups:
    tungstenite:
        patterns:
            - "tokio-tungstenite"
            - "hyper-tungstenite"
            - "tungstenite"
```

## Limitations
It doesn't wait for all the crates in the group to open the pr, it runs according to the frequency specified on the .yml file, which is every week here
So if one of them gets release next week for example, it will open the PR just for that one and the checks will fail,
but as the other ones get released the bot will catch up to them , update the PR and the checks will eventually pass

But because `tungstenite` usually updates together this probably won't happen that often

ShiraiEd:
## Summary

The tungstenite crates depend on each other, every time they fail CI checks, only passing when bumped together.
This PR groups them in the `tungstenite` group

## Related Issues

Closes #697

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -7,6 +7,12 @@ updates:
     open-pull-requests-limit: 10
     commit-message:
       prefix: "Update"
+    groups:
+      tungstenite:
+        patterns:
+          - "tokio-tungstenite"
+          - "hyper-tungstenite"
+          - "tungstenite"
     ignore:
       - dependency-name: "sea-schema"
         update-types: ["version-update:semver-minor", "version-update:semver-major"]


arferreira (approved):
LGTM

